Issue #16, opened 2017-04-24
Cherry pick commits from Saul's omni


adampbeardsley:
make on heracal branch, cherry pick from capo.omni, issue pull request

SaulAryehKohn:
There is now a `polDev` branch on this repo, in which I've just pushed my changes to `heracal/omni.py` and `heracal/tests/test_omni.py`. The build passes on Travis CI and test coverage goes up by 2%. I'm waiting on Issue #11 to be resolved before I start editing scripts `omni_run` and `omni_apply` for polarization capabilities. Should I issue pull request ASAP or wait for #11 ?

bhazelton:
I think it depends on whether the edits you've already made are useful on their own or whether they really need the changes in `omni_run` and `omni_apply` to be useful. I don't have a strong opinion.

SaulAryehKohn:
This is accomplished by Pull Request #19

SaulAryehKohn:
These have been merged into the master branch.

Issue #14, opened 2017-04-24
Test omni_apply on cross-pol data


adampbeardsley:

bhazelton:
we need unit tests on this

SaulAryehKohn:
It's "tested" with a combination of PRs #19 and #21 . But we need unit tests now....

SaulAryehKohn:
We're working on this in PR #26

SaulAryehKohn:
Addressed in PR #41

Issue #15, opened 2017-04-24
integration test for cross-pol data


adampbeardsley:

SaulAryehKohn:
This is satisfied by Pull Request #21 (now merged).

PR #19, opened 2017-05-03
Polarization capabilities

SaulAryehKohn:
In response to Issue #16 , pseudo-Stokes V minimization capabilities added `heracal/omni.py` and tested in `heracal/tests/test_omni.py`. This should **not be merged** until at least it is able to retire Issue #14 . I need to do more testing to see how best to implement more than single-pol calibration in `heracal/scripts/omni_{run,apply}.py`. 

We need to figure out how (or if) we want to apply firstcal solutions to all polarizations of visibilities in `omni_run`. Since firstcal only really works in high S/N regimes, we can only use the xx and yy visibilities to get delay estimates. These are obviously OK to apply to xy and yx (since we get a solution per dipole arm), but it means that we require more than one firstcal file for these visibilities. This will require a [change in structure ](https://github.com/HERA-Team/heracal/blob/master/scripts/omni_run.py#L131) in `omni_run`. There are a number of ways to do this...


SaulAryehKohn:
The version of `omni_run.py` in the above commit adds functionality for cross-pol calibration. It has been shown to generate identical calibrations for a single linear polarization (xx) file as previous stable versions of `omni_run`. Actual cross-pol calibration still remains to be tested, so this pull request should still not be merged.

SaulAryehKohn:
I believe this is now ready for review. This pull request addresses Issue #16 directly, and the review should cover Issues #14 and #15 .

.gitignore
```diff
@@ -1 +1,2 @@
 *.pyc
+tmp*
```
pkgw: Does this change actually need to be committed to the main repository? You can add files in `.git/info/exclude` for ones that are specific to you personally.

heracal/omni.py
```diff
@@ -219,8 +219,40 @@ def compute_reds(nant, pols, *args, **kwargs):
...
+        
+    Return:
+        _reds: the adjusted array of redundant baseline sets.
+    '''
+    _reds = []
+    L = len(reds)
+    try:
+        assert(L%4)==0
+    except AssertionError:
+        raise Exception('Wrong number of polarizations supplied')
```
pkgw: Instead of this stanza, just write:
```
if L % 4 != 0:
   raise Exception(...)
```
Also, the exception text should be more on-point about the check that's being performed: "expected an array with a multiple of 4 elements" or something like that.

heracal/omni.py
```diff
@@ -219,8 +219,40 @@ def compute_reds(nant, pols, *args, **kwargs):
...
+    polarization xy and yx) _always_ in the middle.
+    
+    Args:
+        reds: the array of redundant baseline sets.
+        
+    Return:
+        _reds: the adjusted array of redundant baseline sets.
+    '''
+    _reds = []
+    L = len(reds)
```
pkgw: Avoid capitalized names for variables that aren't global constants. I suggest `n` (fewer potential mistakes than `l`, `1`, `|`, etc.).

heracal/omni.py
```diff
@@ -219,8 +219,40 @@ def compute_reds(nant, pols, *args, **kwargs):
...
+    try:
+        assert(L%4)==0
+    except AssertionError:
+        raise Exception('Wrong number of polarizations supplied')
+    _reds += reds[:L/4]
+    xpols = reds[L/4:3*L/4]
+    _xpols = []
+    for i in range(L/4): _xpols.append(xpols[i] + xpols[i+L/4])
```
pkgw: Please always use newlines for `if` and `for` statements of this kind:

```
for i in range(n/4):
    _xpols.append(xpols[i] + xpols[i+L/4])
```

heracal/omni.py
```diff
@@ -219,8 +219,40 @@ def compute_reds(nant, pols, *args, **kwargs):
             reds += [[(Antpol(i, pi, nant), Antpol(j, pj, nant)) for i, j in gp] for gp in _reds]
     return reds
 
-
-def aa_to_info(aa, pols=['x'], fcal=False, **kwargs):
+def minVreds(reds):
```
pkgw: I don't understand the name of this function. Minimal .. Stokes V ... redundancies? A longer, more informative name would be helpful.

heracal/omni.py
```diff
@@ -240,20 +273,20 @@ def aa_to_info(aa, pols=['x'], fcal=False, **kwargs):
...
         for z, pol in enumerate(pols):
-            z = 2**z
+            z = 2**z # exponential ensures diff xpols aren't redundant w/ each other
             i = Antpol(ant, pol, len(aa))
             antpos[int(i), 0], antpos[int(i), 1], antpos[int(i), 2] = x, y, z
     reds = compute_reds(nant, pols, antpos[:nant], tol=.1)
     ex_ants = [Antpol(i, nant).ant() for i in range(antpos.shape[0]) if antpos[i, 0] == -1]
     kwargs['ex_ants'] = kwargs.get('ex_ants', []) + ex_ants
     reds = filter_reds(reds, **kwargs)
+    if minV: reds = minVreds(reds)
```
pkgw: Again, please use a newline here.

heracal/omni.py
```diff
@@ -532,6 +565,102 @@ def from_fits(filename, pols=None, bls=None, ants=None, verbose=False):
...
 
+def write_uvdata_vis(filename, aa, m, v, xtalk=False, returnuv=True):
+    '''
+    Given meta information and visibilities, write out a uvfits file.
+    filename: filename to write uvfits object.(str)
+    aa      : aipy antenna array object (object)
+    m       : dictionary of information (dict)
+    v       : dictionary of visibilities with keys antenna pair and pol (dict)
+    xtalk   : visibilities given are xtalk visibilities. (bool)
+    returnuv: return uvdata object. If returned, file is not written. (bool)
```
pkgw: This is not the same argument/return documentation format as used in `minVreds`.

bhazelton:
It looks like this PR is significantly decreasing the test coverage. Can you add some tests to better cover the code?

heracal/omni.py
```diff
@@ -532,6 +565,102 @@ def from_fits(filename, pols=None, bls=None, ants=None, verbose=False):
...
+    uv.Nbls = len(v[pols[0]].keys())
+    uv.Nblts = uv.Nbls * uv.Ntimes
+    uv.Nfreqs = len(m['freqs'])
+    data = {}
+    for p in v.keys():
+        if p not in data.keys():
+            data[p] = []
+        for bl in v[p].keys():
+            data[p].append(v[p][bl])
+        data[p] = np.array(data[p]).reshape(uv.Nblts, uv.Nfreqs)
```
pkgw: I think the `for` loop above can be simplified significantly. Does the first `if` condition ever evaluate to False? — I don't think so, because otherwise the `data[p].append()` call won't work because `data[p]` ends up being a Numpy array, which doesn't support that.

I think you might just be able to write something like:

```
for p, bl_dict in v.items():
    data[p] = np.array(bl_dict.values()).reshape((uv.Nblts, uv.Nfreqs))
```

heracal/omni.py
```diff
@@ -532,6 +565,102 @@ def from_fits(filename, pols=None, bls=None, ants=None, verbose=False):
...
+    pols = v.keys()
+    bls = v[pols[0]].keys()
+
+    uv = UVData()
+    if xtalk:
+        uv.Ntimes = 1
+    else:
+        uv.Ntimes = len(m['times'])
+    uv.Npols = len(pols)
+    uv.Nbls = len(v[pols[0]].keys())
```
pkgw: You're not reusing the `bls` variable you created above.

heracal/omni.py
```diff
@@ -532,6 +565,102 @@ def from_fits(filename, pols=None, bls=None, ants=None, verbose=False):
...
+    filename: filename to write uvfits object.(str)
+    aa      : aipy antenna array object (object)
+    m       : dictionary of information (dict)
+    v       : dictionary of visibilities with keys antenna pair and pol (dict)
+    xtalk   : visibilities given are xtalk visibilities. (bool)
+    returnuv: return uvdata object. If returned, file is not written. (bool)
+    '''
+
+    pols = v.keys()
+    bls = v[pols[0]].keys()
```
pkgw: Is this safe? Is `v` an `OrderedDict` or anything? Python dictionaries are not guaranteed to iterate over their keys in any particular order, so you might get inconsistent orderings in the `pols` and `bls` variables, which seems like a bad thing if we're constructing arrays to write out in a FITS file.

heracal/omni.py
```diff
@@ -532,6 +565,102 @@ def from_fits(filename, pols=None, bls=None, ants=None, verbose=False):
...
+    uv.telescope_location = tobj.telescope_location
+    uv.history = m['history']
+
+    # phasing information
+    uv.phase_type = 'drift'
+    uv.zenith_ra = uv.lst_array
+    uv.zenith_dec = np.array([aa.lat] * uv.Nblts)
+
+    # antenna information
+    uv.Nants_telescope = 128
```
pkgw: Can we avoid hardcoding this number?

heracal/omni.py
```diff
@@ -532,6 +565,102 @@ def from_fits(filename, pols=None, bls=None, ants=None, verbose=False):
...
+    for k in aa:
+        antpos.append(k.pos)
+
+    uv.antenna_positions = np.array(antpos)
+
+    if returnuv:
+        return uv
+    else:
+        print('   Saving {0}'.format(filename))
+        uv.write_uvfits(filename, force_phase=True, spoof_nonessential=True)
```
pkgw: Please don't Swiss-Army-knife this function. This function should just return the `uv` object. Callers can add the single line of code to write out the data to a FITS file if that's what they want, or you can write a thin wrapper that does that if the pattern really occurs repeatedly.

heracal/tests/test_omni.py
```diff
@@ -84,6 +84,42 @@ def test_compute_reds(self):
...
+        for arr in reds: cr+=len(arr)
+        for arr in mVreds: cmv+=len(arr)
+        nt.assert_equal(cr,cmv)
+        # test that no crosspols are in linpol red arrays and vice versa
+        for arr in mVreds:
+            p0 = arr[0][0].pol()+arr[0][1].pol()
+            for ap in arr:
+                p = ap[0].pol()+ap[1].pol()
+                nt.assert_equal(len(set(p)),len(set(p0)))
+        # test that every xy has it's corresponding yx in the array
```
pkgw: it's => its

heracal/tests/test_omni.py
```diff
@@ -84,6 +84,42 @@ def test_compute_reds(self):
...
+            if len(set(p0))==1: continue #not interested in linpols
+            for ap in arr:
+                ai,aj = ap
+                bi,bj = omni.Antpol(ai.ant(),aj.pol(),self.info.nant),omni.Antpol(aj.ant(),ai.pol(),self.info.nant) 
+                if (bi,bj) in arr:
+                    continue
+                elif (bj,bi) in arr:
+                    continue
+                else:
+                    raise Exception('Something has gone wrong in polarized redundancy calculation (missing crosspols)')
```
pkgw: Can simplify this to:

```
if (bi, bj) not in arr and (bj, bi) not in arr:
   raise Exception('...')
```

heracal/tests/test_omni.py
```diff
@@ -84,6 +84,42 @@ def test_compute_reds(self):
...
+        # test AssertionError
+        _reds = reds[:-1]
+        try:
+            omni.minVreds(_reds)
+        except:
+            pass
```
pkgw: Never never never `except: pass`!!!

For one thing, this test can never fail. If no exception is raised nothing will happen.

There is probably a nice way to express "this operation should cause an exception" in Python tests, but at least:

```
try:
   omni.minVreds(_reds)
   assert False, 'should not have gotten here'
except Exception:
   pass
```

Also it would probably be better for `minVreds` to raise a `ValueError` rather than a generic exception since that's the kind of exception you're supposed to use when a function has been given invalid inputs.

heracal/tests/test_omni.py
```diff
@@ -169,7 +205,6 @@ def test_from_fits_delay(self):
 
         nt.assert_equal(xtalk, {})
 
-
```
pkgw: Please avoid unneeded whitespace churn

scripts/omni_apply.py
```diff
@@ -11,18 +11,30 @@
...
 o.add_option('--omnipath',dest='omnipath',default='%s.fits',type='string',
-            help='Format string (e.g. "path/%s.npz", where you actually type the "%s") which converts the input file name to the omnical npz path/file.')
+            help='Format string (e.g. "path/%s.fits", where you actually type the "%s") which converts the input file name to the omnical npz path/file.')
 o.add_option('--firstcal', action='store_true', 
             help='Applying firstcal solutions.')
 opts,args = o.parse_args(sys.argv[1:])
 args = np.sort(args)
+pols = opts.pol.split(',')
 
+def file2pol(filename): return filename.split('.')[3] #XXX assumes file naming format
```
pkgw: Ugh, the way that we are hardcoding the file naming format is very gross, but it does seem like the only method we have right now. We should embed metadata in the UV files so that we can actually do this correctly.

scripts/omni_run.py
```diff
@@ -2,220 +2,184 @@
...
+if len(args)==0: raise AssertionError('Please provide visibility files.')
+if opts.minV and len(list(set(''.join(pols))))==1: raise AssertionError('Stokes V minimization requires crosspols in the "-p" option.')
 
-    if returnuv:
-        return uv
-    else:
-        print('   Saving {0}'.format(filename))
-        uv.write_uvfits(filename, force_phase=True, spoof_nonessential=True)
+def getPol(fname): return fname.split('.')[3] #XXX assumes file naming format
+def linPol(polstr): return len(list(set(polstr)))==1
```
pkgw: Is `linPol` the right name for this function? It makes me think "linear polarization", which I don't think is what it's testing. If I understand the function, it's testing whether a polarization string is XX or YY rather than, e.g. XY. MIRIAD calls these "intensity-type" polarizations which I kind of like.

scripts/omni_run.py
```diff
@@ -2,220 +2,184 @@
...
-for f, filename in enumerate(args):
-    file_group = files[filename]  # dictionary with pol indexed files
-    print 'Reading:'
-    for key in file_group.keys():
-        print '   ' + file_group[key]
+### Collect all firstcal files ###
+firstcal_files = {}
+if not opts.firstcal:
+    raise ValueError('Please provide a firstcal file. Exiting...') #XXX this requires a firstcal file for any implementation
+else:
```
pkgw: No need to put this code inside an `else` clause — if the `if` test evaluates to true, the program's over.

scripts/omni_run.py
```diff
@@ -2,220 +2,184 @@
...
+    Nf=0
+    for pp in pols:
+        if linPol(pp):
+            # we cannot use cross-pols to firstcal
+            if '*' in opts.firstcal or '?' in opts.firstcal: flist = glob.glob(opts.firstcal)    
+            elif ',' in opts.firstcal:flist = opts.firstcal.split(',')             
+            else: flist = [str(opts.firstcal)]
+            firstcal_files[pp] = np.sort([s for s in flist if pp in s])
```
pkgw: `sorted(s for s in flist if pp in s)`.

scripts/omni_run.py
```diff
@@ -2,220 +2,184 @@
...
+        if ip==0:
+            for nbl, (i,j) in enumerate(map(uvdp.baseline_to_antnums, uvdp.baseline_array[:uvdp.Nbls])):
+                if not (i, j) in bls and not (j, i) in bls: continue
+                d[(i,j)] = {}
+                f[(i,j)] = {}
+        
+        #XXX I *really* don't like looping again, but I'm not sure how better to do it
+        for nbl, (i,j) in enumerate(map(uvdp.baseline_to_antnums, uvdp.baseline_array[:uvdp.Nbls])):
+            if not (i, j) in bls and not (j, i) in bls: continue
+            d[(i,j)][pp] = uvdp.data_array.reshape(uvdp.Ntimes, uvdp.Nbls, uvdp.Nspws, uvdp.Nfreqs, uvdp.Npols)[:, nbl, 0, :, 0]
```
pkgw: Can omit the parentheses: `d[i,j][pp] = ...`

scripts/omni_run.py
```diff
@@ -2,220 +2,184 @@
...
+        
+        if ip==0:
+            for nbl, (i,j) in enumerate(map(uvdp.baseline_to_antnums, uvdp.baseline_array[:uvdp.Nbls])):
+                if not (i, j) in bls and not (j, i) in bls: continue
+                d[(i,j)] = {}
+                f[(i,j)] = {}
+        
+        #XXX I *really* don't like looping again, but I'm not sure how better to do it
+        for nbl, (i,j) in enumerate(map(uvdp.baseline_to_antnums, uvdp.baseline_array[:uvdp.Nbls])):
+            if not (i, j) in bls and not (j, i) in bls: continue
```
pkgw: Can write `i, j not in bls` rather than `not (i, j) in bls`.

scripts/omni_run.py
```diff
@@ -2,220 +2,184 @@
...
+    
+    ## Append metadata parameters 
+    m2['history'] = 'OMNI_RUN: '+' '.join(sys.argv) + '\n'
     m2['times'] = t_jd
     m2['lsts'] = t_lst
     m2['freqs'] = freqs
-    m2['inttime'] = uvd.integration_time
-    optional = {'observer': 'Zaki Ali (zakiali@berkeley.edu)'}
+    m2['inttime'] = t_int
+    optional = {'observer': 'obs (example@institution.edu)'}
```
pkgw: Surely there's a better value to use here? Just 'heracal'?

SaulAryehKohn:
@pkgw Thanks for the review. I will work on those changes!
@bhazelton The test coverage decreases because I moved a block of @zakiali 's code from the `omni_run` script to the `omni` module. I think I will need his advice on how to best implement a test of that block (`write_uvdata_vis`).

bhazelton:
Ok. Moving code from scripts to modules when they become generally useful is good and it in fact makes them much easier to test. I think this is related to the existing issue # 13 about having test coverage for scripts.

heracal/omni.py
SaulAryehKohn: Do you think `reds_for_minimal_V` is OK?

scripts/omni_run.py
SaulAryehKohn: Sorry, I'm not sure what you're referring to. Do you mean to rename `obs` as `heracal`, or to leave our the email altogether? Or both?

heracal/tests/test_omni.py
SaulAryehKohn: I don't know what this means?

heracal/omni.py
pkgw: That certainly is more understandable to me.

heracal/tests/test_omni.py
pkgw: I.e., getting rid of an extra blank line is not what this PR is about, so it's better not to include the change here.

scripts/omni_run.py
pkgw: My inclination would be both. A fake email address serves no purpose, so why include it? Unless pyuvdata is picky enough that it requires one to be provided ...

scripts/omni_run.py
SaulAryehKohn: changed to `isLinPol`

scripts/omni_run.py
SaulAryehKohn: I think the with-parentheses version is easier to interpret as the baseline that it is. Otherwise I find it harder to keep track of the various indices for debugging etc.

scripts/omni_run.py
SaulAryehKohn: OK. Changed to `'observer': 'heracal'`

heracal/omni.py
```diff
@@ -219,8 +219,39 @@ def compute_reds(nant, pols, *args, **kwargs):
...
+
+    This works in the simple way that it does because
+    of the way the reds arrays are constructed in
+    aa_to_info when 4 polarizations are present: it
+    reprsents them as 4 co-located arrays in (NS,EW) and
+    displaced in z, with the cross-combinations (e.g. 
+    polarization xy and yx) _always_ in the middle.
+    
+    Args:
+        reds: the array of redundant baseline sets.
```
zakiali: reds are lists not arrays. This could get confusing.

heracal/omni.py
```diff
@@ -219,8 +219,39 @@ def compute_reds(nant, pols, *args, **kwargs):
...
+        
+    Return:
+        _reds: the adjusted array of redundant baseline sets.
+    '''
+    _reds = []
+    n = len(reds)
+    if n%4 != 0:
+        raise ValueError('Wrong number of polarizations supplied')
```
zakiali: is this always true? I could imagine that we could have 2 polarizations and still be divisible by 4. It depends on the number of redundancies and polarizations right?

heracal/omni.py
SaulAryehKohn: OK, changed

heracal/omni.py
SaulAryehKohn: If you only have `pols=['x']` in `omni.compute_reds`, `len(reds)=2`, so the ValueError would be raised appropriately, since you can't have cross-pols in that array.

If you have  `pols=['x','y']`,  `len(reds)=8`, so the method will proceed, since cross-pols can be created. 

Even numbers of polarizations will not raise a ValueError. So you could imagine calculating the redundancy array for a future configuration where we have every-other feed rotated 45 degress as in the polarization experiment array, and it would construct an array that minimized the difference of the cross products of those polarizations.... whether or not they should be permitted to do so. But let's cross that bridge when we come to it!

scripts/omni_apply.py
```diff
@@ -11,18 +11,30 @@
...
 
 filedict = {}
-solution_files = np.sort(glob.glob(opts.omnipath))
+solution_files = sorted(glob.glob(opts.omnipath))
```
zakiali: have you run this script? I think this will break if we follow the omnipath option. (I didn't change the help for the omnipath before and was putting in strings with normal wildcards).

SaulAryehKohn: I have, it worked for me. Could you try it also?

scripts/omni_run.py
```diff
@@ -2,220 +2,190 @@
...
+    for ip,pp in enumerate(pols):
+        uvdp = uvd_dict[pp]
+        
+        if ip==0:
+            for nbl, (i,j) in enumerate(map(uvdp.baseline_to_antnums, uvdp.baseline_array[:uvdp.Nbls])):
+                if not (i, j) in bls and not (j, i) in bls: continue
+                d[(i,j)] = {}
+                f[(i,j)] = {}
+        
+        #XXX I *really* don't like looping again, but I'm not sure how better to do it
```
zakiali: you could do something like... 
`if d.has_key((i,j)):
        d[(i,j)][pp] = uvdp.data_array.reshape(uvdp.Ntimes, uvdp.Nbls, uvdp.Nspws, uvdp.Nfreqs, uvdp.Npols)[:, nbl, 0, :, 0]
       d[(i,j)][pp] = uvdp.data_array.reshape(uvdp.Ntimes, uvdp.Nbls, uvdp.Nspws, uvdp.Nfreqs, uvdp.Npols)[:, nbl, 0, :, 0]
else:
    d[(i,j)] = {}`

or something like that...

zakiali: I don't know how to format this as code.. the `` are not working

SaulAryehKohn: That could work, but I don't think it's essential for this PR. People are welcome to disagree though.

PR #26, opened 2017-06-03
Apply autopep8 to files; move firstcal and omnical scripts into modules

plaplant:
Use autopep8 to format repo. Move bulk of firstcal and omnical scripts to module files to improve test coverage.


heracal/omni.py
```diff
@@ -611,7 +718,8 @@ def make_uvdata_vis(aa, m, v, xtalk=False):
     antnums = np.array(v[pols[0]].keys()).T
 
     uv = UVData()
-    bls = sorted(map(uv.antnums_to_baseline, antnums[0], antnums[1])) # XXX sort the baselines
+    # XXX sort the baselines
```
SaulAryehKohn: This "XXX" is now obselete

heracal/omni.py
```diff
@@ -877,15 +978,404 @@ def __init__(self, meta, gains, flags=None, DELAY=False, ex_ants=[], appendhist=
...
+                    break
+        if not fcalfile:  # 2 pol
+            fcalfile = [file2firstcal[file_group[pp]][0]
+                        for pp in linear_pol_keys]
+
+        _, g0, _, _ = from_fits(fcalfile)
+
+        #uvd = pyuvdata.UVData()
+        #uvd.read_miriad([file_group[pp] for pp in pols])
+        # XXX This will become much simpler when pyuvdata can read multiple MIRIAD
```
SaulAryehKohn: This is apparently possible now? Important for a future commit

heracal/omni.py
```diff
@@ -877,15 +978,404 @@ def __init__(self, meta, gains, flags=None, DELAY=False, ex_ants=[], appendhist=
...
+                for nsp, nspws in enumerate(mir.spw_array):
+                    if cal.cal_type == 'gain' and cal.gain_convention == 'multiply':
+                        mir.data_array[blmask, nsp, :, p] = \
+                            mir.data_array[blmask, nsp, :, p] * \
+                            cal.gain_array[antenna_index[ai], nsp, :, :, p1].T * \
+                            np.conj(cal.gain_array[
+                                    antenna_index[aj], nsp, :, :, p2].T)
+
+                    if cal.cal_type == 'gain' and cal.gain_convention == 'divide':
+                        mir.data_array[blmask, nsp, :, p] =  \
```
SaulAryehKohn: I don't like that we're forced to have four of these gain application brackets, but I don't think it's avoidable right now

heracal/__init__.py
```diff
@@ -1 +1,4 @@
-import omni, firstcal, xrfi, metrics
+import omni
+import firstcal
+import xrfi
+import metrics
```
SaulAryehKohn: We need to move `xrfi` and `metrics` out of heracal, but that's not the purpose of this commit

heracal/omni.py
jsdillon: Agreed. It's hideous. But not the purpose of this commit. We can go clean out the modules later now that they are well tested.

Issue #32, opened 2017-06-21
Enable 4pol+minV calibration


SaulAryehKohn:
Right now specifying the `--minV` option in `omni_run` causes the below error, since we overload certain redundancy arrays to tell omnical that xy and yx visibilities are redundant, causing pyuvdata to freak out.

```
Traceback (most recent call last):
  File "/users/plaplant/hera/heracal/scripts/omni_run.py", line 11, in <module>
    omni.omni_run(files, opts, history)
  File "/users/plaplant/anaconda2/envs/capo_sak/lib/python2.7/site-packages/heracal-0.0.0-py2.7.egg/heracal/omni.py", line 1242, in omni_run
    fsj, force_phase=True, spoof_nonessential=True)
  File "/users/plaplant/anaconda2/envs/capo_sak/lib/python2.7/site-packages/pyuvdata-1.1-py2.7.egg/pyuvdata/uvdata.py", line 950, in write_uvfits
    run_check_acceptability=run_check_acceptability)
  File "/users/plaplant/anaconda2/envs/capo_sak/lib/python2.7/site-packages/pyuvdata-1.1-py2.7.egg/pyuvdata/uvfits.py", line 349, in write_uvfits
    raise ValueError('The polarization values are not evenly spaced (probably '
ValueError: The polarization values are not evenly spaced (probably because of a select operation). The uvfits format does not support unevenly spaced polarizations.
```

(FYI the `capo_sak` isn't actually referring to the `capo` repository, it's the name of an anaconda environment.)

jsdillon:
Hold off on this until we figure out whether we actually want MinV.

SaulAryehKohn:
Ok, but FWIW see PR #48

PR #48, opened 2017-07-05
simple fix to enable minV mode.

SaulAryehKohn:
The problem was that, because in the V-minimization mode we treat xy=yx, there are only xy _or_ yx visibility models, not both. This creates an uneven spacing of polarizations (xx=-5, yy=-6 but then we're either missing -7 or -8 for the xy or yx respectively). 

I replicated the crosspol part of the visibility model array (`v3`) to account for the other polarization. It's a little inelegant. Right now I am not conjugating the other crosspol visibility model, but this would be a simple addition if we want it.

In any case, we should talk about whether we want to implement this in general, as noted in Issue #32 . But now we can!


jsdillon:
Let's talk about this on the datacon today.

Issue #51, opened 2017-07-06
Update hera_cal scripts to use new output from hera_qm


plaplant:
Currently hera_cal scripts for firstcal and omnical assume certain metrics, e.g. the bad antenna list, will be passed in as a comma-separated-list in a command-line argument. The scripts should be able to convert the JSON files output from hera_qm into this format. This should be altered in the scripts directory (not the module) so as not to create unnecessary dependencies between the repos.

plaplant:
This was addressed in PR #52.

PR #52, opened 2017-07-07
Add support for processing JSON files from hera_qm

plaplant:
This PR addresses Issue #51 by adding JSON processing of `hera_qm` metric files to firstcal and omnical option parsers. Right now, it's being used only for extracting excluded ants (`ex_ants`), but could be generalized further if desired. More test coverage is required, as well as a sample JSON file for testing purposes.

As a more general comment, it seems the way that we handle `ex_ants` is suboptimal. If multiple files are passed in to firstcal or omnical, the `ex_ants` option is applied to them all. If we're going to support processing multiple file arguments, it might make sense to allow for different lists of bad antennas for each file. (Or, failing that, we should document this behavior.) Further, for omnical operating on cross-pol data, if an antenna number is excluded for one polarization, it is excluded for all of them (e.g., saying antenna `(81, 'x')` is bad also implicitly flags `(81, 'y')`). This is noted in the code, but seems like an unnecessary restriction. Is this related to a fundamental limitation of omnical?


jsdillon:
It's not 100% clear to me what the advantage of doing firstcal on multiple files is unless we're planning to average the solutions. I don't believe we need to do that as we can get close enough solutions integration-by-intergration and we can always retroactively smooth out firstcal within the degenerate subspace of omnical after omnical (but maybe before LST binning or power spectra).

plaplant:
I agree that as a matter of practice, we primarily run first/omnical on an individual time slice. Maybe then the right approach is to just document this and make it the user's responsibility to implement correctly. (I think we already do this with polarization, in that there is no explicit check that, e.g., firstcal called with `-p xx` is operating on a `*.xx.uv` file.)

plaplant:
Further, as the code I've written stands, if multiple polarizations are requested for omnical, a flagged polarization for one antenna means all polarizations are excluded. Do we want to support separate flags for different polarizations for the same antenna?

jsdillon:
Ideally, yes. It's possible for one polarization to be dead but not both.

plaplant:
Okay, that's going to require a fair amount of reworking, and I'd want to allow for it in the command-line argument (something like `--ex_ants=81x,81y,102x`). It'll take me a little bit to put that together.

jsdillon:
This isn't super high priority...it's OK to flag an entire antenna if one of the two pols is idea. If you'd prefer not to fix this now, we can open an issue instead and fix it for a future version of the RTP.

plaplant:
Yeah, I'm trying to wade through how `aa_to_info` works, and not making much progress. I'll put a pin in it for now and just add it as an issue.

plaplant:
@jsdillon I've made `ex_ants` processing ignore polarization in the JSON files, so I think it's ready for another look.

PR #58, opened 2017-07-19
Remove deprecated function

plaplant:
The `concatenate_UVCal_on_pol` functionality is now built in to `UVCal` objects in `pyuvdata`, and so the function is unnecessary. The function and tests for it have been removed, and references to it in the code have been replaced.

### DIFF
--- a/hera_cal/omni.py
+++ b/hera_cal/omni.py
@@ -770,63 +770,6 @@ def make_uvdata_vis(aa, m, v, xtalk=False):
 
     return uv
 
-# XXX Eventually this may belong in pyuvdata
-
-
-def concatenate_UVCal_on_pol(calfitsList):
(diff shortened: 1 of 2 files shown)


SaulAryehKohn:
Can you point me to the equivalent functionality in `pyuvdata`? Also, it seems that the pol concatenation was locked-up with something else, since the coverage has decreased. I'll take a look at what that something is in a sec.

plaplant:
Sure, the function in `pyuvdata` is the `__add__` method in the `UVCal` object (defined in `uvcal.py`). Before I cut it out, I made sure the test was passing with the new function: I changed the function being called in the test from `concatenate_UVCal_on_pol` to the native `+` operation, and it passed.

PR #60, opened 2017-07-21
Fixed a bug in get_pos_reds

jsdillon:
A bug in get_pos_reds caused baselines to appear non-redundant due to
rounding errors. Now I check for all possible rounding errors. I added
a unittest to expose this error. This does not affect hera_cal.omni.


jsdillon:
@AaronParsons approved in person today.

Issue #59, opened 2017-07-21
Make linsolve an optional package


plaplant:
Currently `linsolve` is imported in `redcal.py`, which is needed only for a very small number of tasks for on-site deployment. We should make `linsolve` optional for tests and specific funcions (similar to `casacore` in `pyuvdata`), so that essential tasks and functions can work without having it installed.

jsdillon:
Done in #62

PR #62, opened 2017-07-21
Move linsolve and DataContainer imports to only where they are strictly necessary

jsdillon:


jsdillon:
This addresses #59

hera_cal/redcal.py
```diff
@@ -1,5 +1,3 @@
-from hera_qm.datacontainer import DataContainer
```
plaplant: I think since we've officially decided that `hera_qm` is a dependency of `hera_cal`, this import statement can stay

hera_cal/redcal.py
```diff
@@ -286,6 +286,7 @@ def logcal(self, data, sol0={}, wgts={}, sparse=False):
             sol: dictionary of gain and visibility solutions in the {(index,antpol): np.array}
                 and {(ind1,ind2,pol): np.array} formats respectively
         """
+        import linsolve
```
plaplant: To truly make this optional without breaking tests, we need something with the following syntax (borrowed from pyuvdata):
```
try:
    import linsolve
except(ImportError):
    from nose.plugins.skip import SkipTest
    raise SkipTest('linsolve not detected. linsolve'
                   ' must be installed for this functionality')
```
This should go in each function that wants to import linsolve

PR #64, opened 2017-07-26
Cm integration

dannyjacobs:
Integrates hera_cal with HERA M&C configuration management.  Introduces a custom "get_aa" function which imports a built-in cal file. This cal file reads in a csv file of positions exported by hera_mc/scripts/write_antenna_location_file.py.

Note: that the test which covers this integration has to explicitly exclude antenna 6 which M&C thinks was active when the test data were recorded.


hera_cal/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
+import nose.tools as nt
+from hera_cal.utils import get_HERA_aa
+import numpy as np,sys
+from hera_cal.calibrations import CAL_PATH
+freqs = np.array([0.15])
+
+class Test_utils(object):
```
jpober: As you mentioned, a lot of the things in the calfile aren't covered by tests.  I'm OK with that if you are.  Some things are officially covered, but not tested all that gracefully though. The first test here only works by defaulting to the hardcoded date/locations file.  Maybe anything more advanced is essentially testing M&C (i.e. retrieving other array configurations from other dates)?  But you could probably test more things than just the number of antennas being returned correctly, e.g., are you getting the lat/lon right (you're not), or whether reading the default cal file with different dates returns the same number of antennas (it shouldn't).

hera_cal/hera_cm.py
```diff
@@ -0,0 +1,198 @@
+import numpy as np
+import pandas as pd
+import os
+import aipy
+import pyuvdata.utils as uv_utils
+from astropy.time import Time
+from dateutil.parser import parse
+
+cm_p_m = 100
```
dannyjacobs: For consistency should probably be in aipy.const.  For now a comment has been added.

hera_cal/hera_cm.py
```diff
@@ -0,0 +1,198 @@
...
+#locations_file = os.path.join(os.path.dirname(__file__), 'data/hera_ant_locs_05_16_2017.csv')
+
+# this sets when to find the antenna positions for (they changed as HERA was built)
+# this should not be later than the date that the file was pulled from M&C
+array_epoch_jd = 2457458
```
dannyjacobs: A default, hard coded date is included in the cal file to be somewhat consistent with  historical use of cal files. The way we usually used cal files was that a cal file corresponded to a specific date or date range. 

Here we have this as the default so that the cal file can still be used in the normal way.  I've changed the get_aa function call so that all non-standard inputs revert to defaults within the cal file so that the custom get_HERA_aa function is not required to use this cal file.

hera_cal/hera_cm.py
```diff
@@ -0,0 +1,198 @@
...
+                    if val == 'tau_ns':
+                        prms['aa']['tau_ns'] = self.tau_ns
+                    elif val == 'tau_ew':
+                        prms['aa']['tau_ew'] = self.tau_ew
+                    elif val == 'gain':
+                        prms['aa']['gain'] = self.gain
+            else:
+                try:
+                    top_pos = np.dot(self._eq2zen, self[int(k)].pos)
+                # XXX should multiply this by len_ns to match set_params.
```
dannyjacobs: Ok, I think I've sorted that out?

hera_cal/hera_cm.py
```diff
@@ -0,0 +1,198 @@
...
+def cofa_latlonalt(locations_file):
+    stn_locs_df = pd.read_csv(locations_file)
+
+    cofa = stn_locs_df[stn_locs_df['station_type'] == 'cofa'].to_dict('records')
+    cofa_lat = cofa[0]['latitude']
+    cofa_lon = cofa[0]['longitude']
+    cofa_alt = cofa[0]['elevation']
+
+    return cofa_lat, cofa_lon, cofa_alt
+
```
dannyjacobs: It was a radians vs degrees problem.  Now covered by tests.

hera_cal/tests/test_firstcal.py
```diff
@@ -298,3 +298,17 @@ def test_single_file_execution(self):
         nt.assert_true(os.path.exists(objective_file))
         os.remove(objective_file)
         return
+    def test_mc_calfile(self):
```
dannyjacobs: firstcal calls get_HERA_aa with no calfile specified.

hera_cal/utils.py
```diff
@@ -0,0 +1,33 @@
+import os
+def get_HERA_aa(freqs,calfile='hera_cm',
+                locations_file=None,
+                array_epoch_jd=None):
+    # create an aipy AntennaArray object with the HERA configuration as it
+    # was on array_epoch_jd.
+    #
+    # A default file is supplied but will soon be out of date.
+    #   - create a new one with hera_mc/scripts/XXX.py
```
dannyjacobs: This code has been updated and tested. Comments have been improved here substantially.

hera_cal/utils.py
```diff
@@ -0,0 +1,33 @@
...
+    # NOTE, if you didn't specify a array_epoch_jd, one will be chosen for you
+    # by the cal file.
+    #
+    # Also supports loading of old style cal files, just set calfile away from default
+    #
+
+    #sometimes the default option is None, but we really mean no input
+    if calfile is None:
+        calfile='hera_cm'
+        #set defaults that hera_cm needs.
```
dannyjacobs: All defaults have been moved to the cal file.

hera_cal/utils.py
```diff
@@ -0,0 +1,33 @@
...
+    if locations_file is None:
+        locations_file = os.path.join(os.path.dirname(__file__), 'data/hera_ant_locs_05_16_2017.csv')
+    if array_epoch_jd is None:
+        array_epoch_jd = 2457458  #need a date
```
dannyjacobs: Moved to cal file.

hera_cal/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+    exec('from {calfile} import get_aa'.format(calfile=calfile))
+    if calfile!='hera_cm': #load an alternate cal file
+        #WARNING, array position is being set by whatever happens to be in your cal file
```
dannyjacobs: revised to be more of a comment

hera_cal/tests/test_utils.py
dannyjacobs: Added a test to check the two "epochs" we currently have, H19 and ~H34.

Note: antpos_ideal is currently constructed with "filler" antennas at position -1,-1,-1.  omnical, firstcal all automatically ignore these fillers but other things probably wont.

hera_cal/utils.py
```diff
@@ -0,0 +1,33 @@
...
+                locations_file=None,
+                array_epoch_jd=None):
+    # create an aipy AntennaArray object with the HERA configuration as it
+    # was on array_epoch_jd.
+    #
+    # A default file is supplied but will soon be out of date.
+    #   - create a new one with hera_mc/scripts/XXX.py
+    #
+    # NOTE, if you didn't specify a array_epoch_jd, one will be chosen for you
+    # by the cal file.
```
dannyjacobs: Yeah, having defaults in two places is confusing. Changed so that all defaults are in the cal file, this function is now just limited to handing the case of defaulting to a base cal file which uses extra options.  

Hopefully the docstring is now the most helpful get_aa docstring ever written.

dannyjacobs:
Trimmed down cal file and improved its code coverage.  @jpober Should address your comments.

jpober:
Looks good, review is approved.  I'll leave you the joy of rebasing and merging.

PR #65, opened 2017-08-01
Refactor tests to eliminate warnings

plaplant:
This PR changes many of the tests to prevent warnings from occurring during `nosetests`. This is done by wrapping many of the file reading calls (which trigger warnings about the antenna diameter not being set) with `pyuvdata.tests.checkWarnings` function.


plaplant:
Looks like the merge conflict was just a call to `omni.omni_apply` that was wrapped in `checkWarnings` vs not

PR #67, opened 2017-08-04
update travis aipy install

SaulAryehKohn:
This will fail until it's also fixed for pyuvdata and hera_qm


adampbeardsley:
With pyuvdata and hera_qm now updated, this branch is failing on a uvdata equality check.

plaplant:
Looks like it might be the antenna positions part that's not equal. Is this due to the recent change in pyuvdata? (https://github.com/HERA-Team/pyuvdata/pull/222)

adampbeardsley:
Could be. I was going to check by restarting the build on master... but that will probably fail because of the aipy upgrade on pyuvdata side. We'll have to debug locally I think.

plaplant:
@adampbeardsley after some offline discussion with @bhazelton, I rewrote the data file that was used for the test. It seems like it was indeed related to the change in how `pyuvdata` handles antenna positions. Making a plot on my local machine of old positions vs. new ones, they seems to be rotated with respect to each other. However, the positions themselves are internally self-consistent, so calculating redundancies (which I think is the only reason why we include antenna positions for HERA) should not be affected. I'm happy to share these plots if you like. Other than that, I think this is now ready to merge.

bhazelton:
@plaplant the rotation is expected -- it turns out that both uvfits and miriad use a rotated frame for their antenna locations, where the x-axis goes through the local meridian rather than the prime meridian. Of course, for South Africa, that's only a small rotation.

bhazelton:
The change to pyuvdata was to account for that rotation, which we had not been properly accounting for before. The locations on the UVData object are stored in the more standard reference frame where the x-axis goes through the prime meridian. This is the same reference frame that we use for the telescope location, rather than having the antenna positions and the telescope location in different coordinate systems (as uvfits and miriad do).

PR #74, opened 2017-08-24
DO NOT MERGE: Add call to M&C database for array configuration

plaplant:
This PR adds functionality for making a live call to the CM database in the M&C system to get array information. If it is unavailable for any reason (`hera_mc` not installed, M&C system is down, etc.), it falls back on using the provided CSV file.

However, there are some loose ends that need tying off. This code is untested, since we need to install `hera_mc`, and spin up a sample CM database to test the rest of the function. I think we should set this up, but wanted to get other people's thoughts on doing this.

Also, this only queries the database "as-is" at the time the function is run, and does not support getting a past configuration state. I was envisioning running "post-processing" in that fashion by using the CSV file, so that we don't need to be plugged in to the live M&C database.

### DIFF
--- a/hera_cal/hera_cm.py
+++ b/hera_cal/hera_cm.py
@@ -8,10 +8,10 @@ from dateutil.parser import parse
 
 cm_p_m = 100 #centimeters per meter, used in AntennaArray
 
-#default locations file
+# default locations file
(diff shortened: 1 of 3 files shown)


plaplant:
Now antenna position information is being correctly written into the files by the correlator (via a call to M&C), and so there is no need to add it in later in the pipeline. With the merge of PR #80, this call is no longer necessary.

PR #75, opened 2017-08-26
Fix ex_ants bug

plaplant:
This PR changes a small bug to make the ex_ants function work properly in omni_run.


SaulAryehKohn:
For some reason the wifi I'm using has firewalled Slack, but I'm happy to approve this PR if it needs approving.

PR #76, opened 2017-08-28
add "--outpath" option to omni_apply.py

nkern:
This PR adds the "--outpath" option to the omni_apply.py executable so that the user can write out omnical-ibrated visibility data to a directory **other than** the directory where the uncalibrated data is loaded. Follows the same convention as found in the firstcal_run.py executable.

### DIFF
--- a/hera_cal/omni.py
+++ b/hera_cal/omni.py
@@ -1452,7 +1452,7 @@ def omni_apply(files, opts):
             outpath = '/'.join(f.split('/')[:-1])
         else:
             outpath = opts.outpath
-        out_filename = "%s/%s" % (outpath, inp_filename)
+        out_filename = "{0}/{1}".format(outpath, inp_filename)
 
         # Write to file
         if opts.firstcal:
--- a/hera_cal/omni.py
+++ b/hera_cal/omni.py
@@ -1081,6 +1081,8 @@ def get_optionParser(methodName):
                      help='Applying firstcal solutions.')
         o.add_option('--extension', dest='extension', default='O', type='string',
                      help='Filename extension to be appended to the input filename')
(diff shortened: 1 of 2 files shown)


hera_cal/omni.py
```diff
@@ -1444,11 +1446,20 @@ def omni_apply(files, opts):
                         np.logical_or(cal.flag_array[antenna_index[ai], nsp, :, :, p1].T,
                                       cal.flag_array[antenna_index[aj], nsp, :, :, p2].T))
 
+        # Define output path and filename
+        inp_filename = f.split('/')[-1]
+        if opts.outpath is None:
+            outpath = '/'.join(f.split('/')[:-1])
+        else:
+            outpath = opts.outpath
+        out_filename = "%s/%s" % (outpath, inp_filename)
```
plaplant: This is more style than substance, but I think we should shy away from this form of string formatting for the more robust `"".format()` statement. So, this could alternatively be written as:
```
out_filename = "{0}/{1}".format(outpath, inp_filename)
```
Not a necessary change, but something to think about.

nkern: sounds good

hera_cal/omni.py
```diff
@@ -1444,11 +1446,20 @@ def omni_apply(files, opts):
                         np.logical_or(cal.flag_array[antenna_index[ai], nsp, :, :, p1].T,
                                       cal.flag_array[antenna_index[aj], nsp, :, :, p2].T))
 
+        # Define output path and filename
+        inp_filename = f.split('/')[-1]
+        if opts.outpath is None:
+            outpath = '/'.join(f.split('/')[:-1])
+        else:
+            outpath = opts.outpath
```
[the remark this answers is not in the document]
nkern: cool, done

PR #85, opened 2017-09-18
adds --overwrite option to firstcal_run.py to clobber file if it exists

nkern:
modified:   hera_cal/firstcal.py
	modified:   hera_cal/tests/test_firstcal.py

### DIFF
--- a/hera_cal/tests/test_firstcal.py
+++ b/hera_cal/tests/test_firstcal.py
@@ -313,3 +313,25 @@ class Test_firstcal_run(object):
         nt.assert_true(os.path.exists(objective_file))
         os.remove(objective_file)
         return
+
+    def test_overwrite(self):
+        objective_file = os.path.join(
+            DATA_PATH, 'zen.2457698.40355.xx.HH.uvcAA.first.calfits')
+        xx_vis4real = os.path.join(DATA_PATH, xx_vis)
+        if os.path.exists(objective_file):
+            os.remove(objective_file)
+        _ = open(objective_file, 'a').close()
+        o = firstcal.firstcal_option_parser()
+        cmd = "-C {0} -p xx --overwrite {1}".format(calfile, xx_vis4real)
+        opts, files = o.parse_args(cmd.split())
+        history = 'history'
+        firstcal.firstcal_run(files, opts, history)
+        # check its a calfits file
+        uvc = UVCal()
+        uvc.read_calfits(objective_file)
+        # check a metadata column for accuracy
+        nt.assert_equal(uvc.Nants_data, 19)
+        # remove file
+        os.remove(objective_file)
+        return
+
--- a/hera_cal/tests/test_omni.py
+++ b/hera_cal/tests/test_omni.py
@@ -855,26 +855,26 @@ class Test_omni_run(object):
         xx_fcal4real = os.path.join(DATA_PATH, 'test_input', xx_fcal)
         xx_vis4real = os.path.join(DATA_PATH, xx_vis)
         omnipath = os.path.join(DATA_PATH, 'test_output')
-	# create blank files
-	_ = open(objective_file1, 'a').close()
+        # create blank files
+        _ = open(objective_file1, 'a').close()
         _ = open(objective_file2, 'a').close()
         _ = open(objective_file3, 'a').close()
-	# run firstcal
+        # run firstcal
         cmd = "-C %s -p xx --firstcal=%s --ex_ants=81 --omnipath=%s --overwrite %s" % (
             calfile, xx_fcal4real, omnipath, xx_vis4real)
         opts, files = o.parse_args(cmd.split())
         history = 'history'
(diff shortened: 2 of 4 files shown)


hera_cal/tests/test_firstcal.py
```diff
@@ -313,3 +313,20 @@ def test_single_file_execution_nocalfile(self):
         nt.assert_true(os.path.exists(objective_file))
         os.remove(objective_file)
         return
+
+    def test_overwrite(self):
+        objective_file = os.path.join(
```
plaplant: In my understanding, this test will remove the objective file if it exists, make a new empty file, and runs firstcal with the `--overwrite` flag to make an actual calfits file. However, the test just checks that the file exists, which would be true if the test silently failed (for whatever reason). I think a more precise test is to open and read the newly written file to ensure it's a calfits file, and verify one or two of the file properties are what we expect it to be. I know this is a higher bar than many of the firstcal tests typically have (though incidentally we should probably be more stringently testing the output of other tests too). It's just that in this case, we're purposely putting a file where we're eventually going to go looking for one, so assuming nothing throws an error, this test will pass.

hera_cal/tests/test_firstcal.py
```diff
@@ -313,3 +313,20 @@ def test_single_file_execution_nocalfile(self):
...
+            DATA_PATH, 'zen.2457698.40355.xx.HH.uvcAA.first.calfits')
+        xx_vis4real = os.path.join(DATA_PATH, xx_vis)
+        if os.path.exists(objective_file):
+            os.remove(objective_file)
+        os.system("touch {0}".format(objective_file))
```
plaplant: There is probably a more OS-independent way to achieve this. Maybe something like:
```python
open(objective_file, 'a').close()
```

hera_cal/tests/test_firstcal.py
nkern: good point!

hera_cal/tests/test_firstcal.py
nkern: good idea

hera_cal/tests/test_firstcal.py
```diff
@@ -313,3 +313,21 @@ def test_single_file_execution_nocalfile(self):
...
+        if os.path.exists(objective_file):
+            os.remove(objective_file)
+        _ = open(objective_file, 'a').close()
+        o = firstcal.firstcal_option_parser()
+        cmd = "-C {0} -p xx --ex_ants=81 --overwrite {1}".format(calfile, xx_vis4real)
+        opts, files = o.parse_args(cmd.split())
+        history = 'history'
+        firstcal.firstcal_run(files, opts, history)
+        uvc = UVCal()
+        uvc.read_calfits(objective_file)
```
plaplant: Great fix! Could you just add a couple of simple assert statements here? I'm thinking along the lines of `nt.assert_equal(uvc.Nants_data, 18)` (or whatever it's supposed to be for this file). That way we can make sure the metadata is being set properly.

nkern: Hey, yeah I added a metadata test for the test_overwrite() function. It seems however that those kinds of tests really belong in the test_singlefile_execution() test, b/c that is directly looking to test that sort of thing. I think the test_overwrite() function should look to specifically target the additional functionality brought by the --overwrite option, which is to simply write out the file even if the file already exists.

nkern:
Also added an --overwrite option for omni_apply.py

nkern:
@plaplant let me know if we need anything else before we merge this

hera_cal/tests/test_firstcal.py
plaplant: I totally agree, and I think we should have more robust tests in the main test as well. My concern was more with just adding an assert statement for nose to find, since I've seen instances of nose skipping tests if an assert statement is missing. I appreciate the addition of the assert statement.

hera_cal/tests/test_omni.py
```diff
@@ -962,6 +962,27 @@ def test_single_file_execution_omni_apply_overwrite(self):
...
+        nt.assert_true(os.path.exists(objective_file))
+	# make sure its a miriad file
+	uvd = UVData()
+	uvd.read_miriad(objective_file)
+	# make sure a metadata column is set correctly
+        nt.assert_equal(uvd.Nants_data, 19)
+        # clean up when we're done
+        shutil.rmtree(objective_file)
+
+    def test_single_file_execution_omni_apply_overwrite(self):
```
plaplant: It looks like the name of this test is the same as the one above it. Are they supposed to be different? Also, it looks to me like some of the columns are out of alignment here, but not in my text editor--maybe it's a tabs vs spaces thing?

nkern: ah, yeah that was a mistake...

nkern:
@plaplant okay fixed that, and added --overwrite option to omni_run.py as well. thanks for reviewing!

PR #88, opened 2017-09-18
deleted a bunch of repeated/extraneous code from the HERACal object

SaulAryehKohn:


hera_cal/omni.py
```diff
@@ -936,21 +934,23 @@ def __init__(self, meta, gains, flags=None, DELAY=False, ex_ants=[], appendhist=
             self.history = appendhist
         self.Nants_data = len(ants)  # only ants with data
         self.Nants_telescope = len(allants)  # all antennas in telescope
-        self.antenna_names = namarray[:self.Nants_telescope]
-        self.antenna_numbers = numarray[:self.Nants_telescope]
```
adampbeardsley: Is there a reason this indexing is no longer needed? Or was it never actually needed?

SaulAryehKohn: Yeah, those indexing operations always select the entire array. Not sure why they're there. But maybe @zakiali had a reason? I just didn't get it?

zakiali: I don't remember either... but probably a relic from a pre calfits era when namarray/numarray etc had size nants*ntimes.

adampbeardsley: Good enough for me.

PR #92, opened 2017-09-23
Feature/find rotated antennas

zakiali:
This PR address looking for rotated feeds (180 degree rotated) and writes out the rotated antennas for a json file. When calculating the firstcal solutions, the input data to firstcal corrects for the pi shift, so the firstcal solutions relfect the correct rotation states of the antennas. This does not correct the data downstream.


hera_cal/firstcal.py
```diff
@@ -383,7 +387,7 @@ def flatten_reds(reds):
         freds += r
     return freds
 
-def UVData_to_dict(uvdata_list, filetype='miriad'):
+def UVData_to_dict(uvdata_list, filetype='miriad', ex_ants=[]):
```
jsdillon: This shouldn't be necessary for firstcal. The ex_ants are already in the info object.

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
     Returns:
         "file".first.calfits: delay calibrations for each antenna (up to some overall delay). (pyuvdata.calfits file)
     '''
+
+    def _apply_first_cal(data_dict, sols, fqs):
+        """Apply delay calibration solutions.
+        
+        Args:
+            data_dict (dict): Dictionary of visibilities indexed by bl and pol.
+            sols (dict): Sictionary of delay solutions (output of Firstcal.run)
```
jsdillon: Typo

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
...
+    def _apply_pi_shift(data_dict, invert_these):
+        """Apply a pi shift to visibilities.
+
+        Args:
+            data_dict (dict): Dictionary of visibilities indexed by bl and pol.
+            invert_these (list): list of antennas to add a pi phase shift too
+        """
+        for ai, aj in data_dict.keys():
+            for pol in data_dict[ai, aj].keys():
+                if (ai, pol[0]) in invert_these or (aj, pol[1]) in invert_these:
```
jsdillon: not or but xor

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
...
+            option_parser (OptionParser): option parser object
+
+        Return:
+            dict : firstcal solutions
+            list : list of rotated antennas
+        '''
+        switched_history = []  # keep track of all rotated antennas
+        switched = []  # keep track of rotated antennas in each iteration
+        niters = 0
+        try:
```
jsdillon: This is already in the info object.

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
...
+            ex_ants = map(int, option_parser.ex_ants.split(','))
+        except ValueError:
+            ex_ants = []
+
+        while niters == 0 or len(switched) > 0:
+            datapack, wgtpack = UVData_to_dict([uv], ex_ants=ex_ants)
```
jsdillon: rename to flagpack

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
...
+                          verbose=option_parser.verbose,
+                          average=option_parser.average, 
+                          window='none')
+
+            # Now we need to check if antennas are flipped
+            medians = {}
+            cal_data = copy.deepcopy(datapack)
+            cal_data = _apply_first_cal(cal_data, sols, fqs*1e9)
+
+            ratio_bls = [(bl1, bl2) for bls in info.get_reds() for b1, bl1 in enumerate(bls) for bl2 in bls[b1 + 1:] if bl1 != bl2]
```
jsdillon: See if len(bls) == 1 breaks this.

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
...
+            for i, ((a1, a2), (a3, a4)) in enumerate(ratio_bls):
+                # need to get polarizations if run in dual pol. This works because we are looking at reds
+                # so that if one antenna is x, all should be x.
+                if a1 < info.nant:
+                    pol = 'x'
+                else:
+                    pol = 'y'  
+                if (a1, a2) not in cal_data.keys():
```
jsdillon: Need to check data conjugation when flipping a1, a2. Alternatively, look into using datacontainer.

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
...
+                       medians[antpol] = np.append(medians[antpol], median)
+                    else:
+                        medians[antpol] = median
+            medians = {k: np.median(np.abs(m)) for k, m in medians.items()}
+            # if the ratio is greater than np.pi/2, it is a switched antenna.
+            switched = [k for k, m in medians.items() if m > np.pi / 2] 
+            switched_history += switched
+            niters += 1 
+        return sols, switched_history
+
```
jsdillon: Check whether your flipped antennas agree x and y pols

hera_cal/firstcal.py
```diff
@@ -462,6 +469,106 @@ def firstcal_run(files, opts, history):
...
+                if (a3, a4) not in cal_data.keys():
+                    a3, a4 = a4, a3
+                median = np.median(np.angle(cal_data[(a1,a2)][pol*2]*np.conj(cal_data[(a3,a4)][pol*2])), axis=1)
+                for ai in [a1, a2, a3, a4]:
+                    antpol = (ai,pol)
+                    if antpol in medians:
+                       medians[antpol] = np.append(medians[antpol], median)
+                    else:
+                        medians[antpol] = median
+            medians = {k: np.median(np.abs(m)) for k, m in medians.items()}
```
jsdillon: Make variables names more explicit. Median over what???

hera_cal/firstcal.py
zakiali: It does not... will just return an empty list, which is a no op in this situation.

jsdillon:
Also, as mentioned in today's commissioning telecon, we need to show that omnical works properly using gains-formatted first.calfits files that could out of here...basically an integration test of the pipeline on the command line.

zakiali:
This plot shows omnical gains with the new modified firstcal algorithm to detect rotated antennas and the previous algorithm. Gains look pretty stable. There are differences because the input firstcal solutions are slightly different. 
screen shot 2017-09-27 at 12 33 01 pm

hera_cal/firstcal.py
```diff
@@ -621,7 +612,14 @@ def _search_and_iterate_firstcal(uv, info, option_parser):
             uv_in.unphase_to_drift()
         
         sols, write_to_json = _search_and_iterate_firstcal(uv_in, info, opts)
-        rotated_antennas = {'rotated_antennas': str(write_to_json)}
+        rotated_antennas = {'rotated_antennas': str(write_to_json),
```
jsdillon: rename

jsdillon:
This is currently failing unittests on travis because of issues installing pyephem (as part of pyuvdata), but it works fine locally so I'm going to merge despite that.

PR #93, opened 2017-09-25
Add AbsCal calibration type

plaplant:
This PR adds the `AbsCal` calibration type for loading gains determined via a calibration type external of omnical (e.g., absolutely calibrated gains via CASA) into a pyuvdata `UVCal` object. It assumes that gains are saved as an `.npz` object, where the dictionary keys are antenna numbers, and the values are an array `Nfreqs` long. The user must also specify whether the gains are to be multiplied or divided when applied to the data. Currently the object only supports gains-type (i.e., omnical-like) solutions, and not delay-type (firstcal-like) solutions, though adding this functionality is desirable.

Also, this PR slightly reorganizes the repo, so that the new `AbsCal` object and the `HERACal` object (previously defined in `omni.py`) are both in a module called `cal_types.py`. Functions and methods internal to `hera_cal` have been changed to reflect this reorganization, but note that any scripts external to the repo using the `HERACal` object will have to be changed to reflect this reorganization. (In the author's estimation, such scripts likely do not exist, and is simply noted here for posterity.)


plaplant:
Thanks for reviewing! Just to clarify your comment, am I not capturing the flags the way I should be for cross pol data? I save the two sets of flags along different indices in the Jones axis (I think), is there something I should be doing differently?

zakiali:
No, you're doing it that way. I think I was only looking at the first commit you had which was only saving flags for one pol. I wrote the review, before realizing that the change was made. All good!

PR #107, opened 2017-10-12
Add flag_missing option to omni_apply

plaplant:
This PR adds a `flag_missing` option to `omni_apply`. If specified, then all visibilities corresponding to antennas missing from the calibration solution are flagged, as a way of indicating that those visibilities are uncalibrated.

### DIFF
--- a/hera_cal/omni.py
+++ b/hera_cal/omni.py
@@ -962,6 +962,9 @@ def get_optionParser(methodName):
         o.add_option('--outpath', dest='outpath', default=None, type='string',
                      help='Directory to write-out omnical-ibrated visibility data. Will use input file path by default.')
         o.add_option('--overwrite', action='store_true', default=False, help='Overwrite output file if it exists.')
+        o.add_option('--flag_missing', action='store_true', default=False,
+                     help='Flag visibilities of antennas that exist in the data, but do not have a '
+                     'corresponding gain in the calibration solution. Default is False')
 
     return o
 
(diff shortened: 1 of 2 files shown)


hera_cal/omni.py
```diff
@@ -1297,7 +1300,10 @@ def omni_apply(files, opts):
             for bl, k in zip(*np.unique(mir.baseline_array, return_index=True)):
                 blmask = np.where(mir.baseline_array == bl)[0]
                 ai, aj = mir.baseline_to_antnums(bl)
-                if not ai in cal.ant_array or not aj in cal.ant_array:
+                if ai not in cal.ant_array or aj not in cal.ant_array:
```
zakiali: I think you'll need to move this block of code into the next for loop because nsp doesn't exist yet.

hera_cal/omni.py
```diff
@@ -1297,7 +1300,10 @@ def omni_apply(files, opts):
...
+                    if opts.flag_missing:
+                        # flag the visibilities if either antenna is missing from calibration solutions
+                        mir.flag_array[blmask, nsp, :, p] = True
```
zakiali: Des the flag array not need to be an array? Does the True transfer to all freqs and times?

hera_cal/omni.py
plaplant: Ah, good catch. I'll move it

hera_cal/omni.py
```diff
@@ -1297,7 +1300,10 @@ def omni_apply(files, opts):
             for bl, k in zip(*np.unique(mir.baseline_array, return_index=True)):
                 blmask = np.where(mir.baseline_array == bl)[0]
                 ai, aj = mir.baseline_to_antnums(bl)
-                if not ai in cal.ant_array or not aj in cal.ant_array:
+                if ai not in cal.ant_array or aj not in cal.ant_array:
+                    if opts.flag_missing:
```
zakiali: Should it not be an option to flag missing antennas, but a necessity? I'm more in favor of having this be the default behavior of applying cal solutions. At the very least have the options default be True.

hera_cal/omni.py
plaplant: Yes, the `True` value broadcasts to all of the specified array indices

hera_cal/omni.py
plaplant: That's a good point. I'll make the option default to True.

PR #109, opened 2017-10-13
Adding ex ants to history of extracted hh files

adampbeardsley:
I made a first attempt at putting the ex_ants into extract_hh files. Not sure if it's the exact format we want though - I assumed the ex_ants file will simply have the excluded antennas, where each antenna is on its own line. Should I have instead used a json so we can use omni.process_ex_ants?


plaplant:
Code looks good! I think having the list with one antenna on each line is perfect, and we can reorganize it into a comma-separated list as part of RTP and feed it in with the `ex_ants` command line argument. 

I think as a part of this PR, we should also make the initial list of ex_ants, since it's all part of the same functionality. (I also think that having it as a part of hera_cal makes sense, since we're going to be putting this repo's version into the history along with the excluded antennas.) I propose putting it in `hera_cal/calibrations/herahex_ex_ants.txt`. What do other people think?

adampbeardsley:
Good idea, just added. Note that I included all antennas up to 350, so we have to actively remove them from the list once they get commissioned.
What do we think about 82? ARP noticed bad xtalk, but it doesn't seem to be killing calibration solutions.

plaplant:
Thanks! I'm for leaving it in, since calibration is doing okay.

Issue #114, opened 2017-10-24
Fix amplitude normalization to include all visibilities, not just unique ones.


jsdillon:
https://github.com/HERA-Team/hera_cal/blob/8de90f0f2afe01adec5a5e317716af5a8e7811b4/hera_cal/redcal.py#L461

jsdillon:
Addressed in PR #115

PR #115, opened 2017-10-25
Fix Amplitude Renormalization in Remove_Degen

jsdillon:
At one point, we renormalized the amplitudes of the gains (and correspondingly, the visibilities) after omnical by forcing the average abs(gains) to be 1.0 (or whatever was put into the degen_sol).

Later, we decided that we wanted to normalize gains so that the mean gain product over the baseline pairs in the visibilities was taken to be 1 (or whatever was put into the degen_sol). This was originally accomplished with statements like

`meanSqAmplitude = np.mean([np.abs(g[(ant1,pol[0])] * g[(ant2,pol[1])]) for (ant1,ant2,pol) in v.keys() if pol == 2*antpol], axis=0)`

However, this is incorrect. `v.keys()` only gives the unique baselines, but we want to loop over all baselines.

We've implemented this fix, by introducing
`bls_for_average = [bl for bls in self.reds for bl in bls if (bl[2] == 2*antpol) and (bls[0] in bl_pairs)]` and looping over that. However, this means that remove_degen now needs to know about the `reds`. Redcal uses a polarization-aware format for reds than omni.py, so we had to edit omni.py to produce redcal's style of reds from `info.get_reds()`. Tests have also been updated.

This addresses #114. This may be of interest to you, @miguelfmorales

### DIFF
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -460,10 +460,11 @@ class RedundantCalibrator:
 
         #Amplitude renormalization: fixes the mean abs product of gains (as they appear in visibilities)
         for antpol in antpols:
+            bls_for_average = [bl for bls in self.reds for bl in bls if (bl[2] == 2*antpol) and (bls[0] in bl_pairs)]
             meanSqAmplitude = np.mean([np.abs(g[(ant1,pol[0])] * g[(ant2,pol[1])])
-                for (ant1,ant2,pol) in bl_pairs if pol == 2*antpol], axis=0)
+                for (ant1,ant2,pol) in bls_for_average], axis=0)
             degenMeanSqAmplitude = np.mean([np.abs(degen_sol[(ant1,pol[0])] * degen_sol[(ant2,pol[1])])
-                for (ant1,ant2,pol) in bl_pairs if pol == 2*antpol], axis=0)
+                for (ant1,ant2,pol) in bls_for_average], axis=0)
             gainSols[gainPols == antpol] *= (degenMeanSqAmplitude / meanSqAmplitude)**.5
             visSols[visPols[:,0] == antpol] *= (meanSqAmplitude / degenMeanSqAmplitude)**.5
             visSols[visPols[:,1] == antpol] *= (meanSqAmplitude / degenMeanSqAmplitude)**.5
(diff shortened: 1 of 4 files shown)


hera_cal/redcal.py
```diff
@@ -460,10 +460,11 @@ def remove_degen(self, antpos, sol, degen_sol=None):
...
         for antpol in antpols:
+            bls_for_average = [bl for bls in self.reds for bl in bls if (bl[2] == 2*antpol) and (bls[0] in bl_pairs)]
             meanSqAmplitude = np.mean([np.abs(g[(ant1,pol[0])] * g[(ant2,pol[1])])
-                for (ant1,ant2,pol) in bl_pairs if pol == 2*antpol], axis=0)
+                for (ant1,ant2,pol) in bls_for_average], axis=0)
             degenMeanSqAmplitude = np.mean([np.abs(degen_sol[(ant1,pol[0])] * degen_sol[(ant2,pol[1])])
-                for (ant1,ant2,pol) in bl_pairs if pol == 2*antpol], axis=0)
+                for (ant1,ant2,pol) in bls_for_average], axis=0)
             gainSols[gainPols == antpol] *= (degenMeanSqAmplitude / meanSqAmplitude)**.5
             visSols[visPols[:,0] == antpol] *= (meanSqAmplitude / degenMeanSqAmplitude)**.5
```
zakiali: Should the visSols be multiplied by the ratio^1 instead of ratio^.5

jsdillon: It's ratio**.5, but it's done twice...once first pol and once for the second. So `visSols[visPols[:,0] == 'xx'` for example will select all visibilities with 'xx' in the first position.  This allows us to remove degen from xy and yx visibilities too.

zakiali: Got it

miguelfmorales:
Have we run this on data (say the IDR)? There are clear amplitude errors in the original, and I’d love to see if this fixes it. 

Also, as a general rule, I think we are at a point that all PRs that affect the behavior of the RTP need to be tested on data with a plot showing what has changed. We really don’t want to accidentally introduce errors and data jackknifes are the only way I know to do that.

M

dannyjacobs:
Yes, I think that was the conclusion from our discussion yesterday.  New changes to RTP need to be run on a day or two of data and notebooks or other relevant analysis redone for comparison.

PR #118, opened 2017-11-14
added "get_antpos" function to utils.py w/ tests

nkern:
I wrote this script for myself and i'm finding I'm using it a lot, so I figured other people would benefit from this. should be a quick & easy review

### DIFF
--- a/hera_cal/tests/test_utils.py
+++ b/hera_cal/tests/test_utils.py
@@ -60,3 +60,27 @@ class TestAA(object):
         new_top = [new_params['0'][key] for key in antpos.keys()]
         old_top = [antpos[key] for key in antpos.keys()]
         nt.assert_true(np.allclose(old_top, new_top))
+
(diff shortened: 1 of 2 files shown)


jsdillon:
Shouldn't this be in pyuvdata?

nkern:
yeah could be.. there are a few functions like this one in `hera_cal` (`get_aa_from_uv` and `UVData_to_dict`), so its not clear to me exactly what belongs in `pyuvdata` and what is just being swept into `hera_cal`...

jsdillon:
Yeah...I'm not 100% sure either. Curious what @bhazelton thinks.

bhazelton:
I'm not at all opposed to adding it to pyuvdata as a method on the uvdata instance. I think other people are more likely to find it and use it there. But I think you should come up with a slightly more clear name for the function.

nkern:
ok, i'll make a PR on pyuvdata then!

PR #119, opened 2017-11-17
bug fix to firstcal.UVData_to_dict: now works w/ sorted baseline arrays

nkern:
`firstcal.UVData_to_dict` assumed the `uvd.baseline_array` was not sorted. for some files (like omnical model visibility files) this sometimes isn't the case. It now works with any form of sorting for `uvd.baseline_array`. tests to ensure this works are included.

### DIFF
--- a/hera_cal/firstcal.py
+++ b/hera_cal/firstcal.py
@@ -397,26 +397,29 @@ def UVData_to_dict(uvdata_list, filetype='miriad'):
             uv_in = UVData()
             # read in file without multiple if statements
             getattr(uv_in, 'read_' + filetype)(fname)
-        # reshape data and flag arrays to make slicing time and baselines easy
-        data = uv_in.data_array.reshape(
-            uv_in.Ntimes, uv_in.Nbls, uv_in.Nspws, uv_in.Nfreqs, uv_in.Npols)
-        flags = uv_in.flag_array.reshape(
-            uv_in.Ntimes, uv_in.Nbls, uv_in.Nspws, uv_in.Nfreqs, uv_in.Npols)
 
+        # iterate over unique baselines
         for nbl, (i, j) in enumerate(map(uv_in.baseline_to_antnums, np.unique(uv_in.baseline_array))):
             if (i, j) not in d:
                 d[i, j] = {}
                 f[i, j] = {}
             for ip, pol in enumerate(uv_in.polarization_array):
                 pol = pol2str[pol]
+                try:
+                    new_data = copy.copy(uv_in.get_data((i, j))[:, :, ip])
+                    new_flags = copy.copy(uv_in.get_flags((i, j))[:, :, ip])
+                except:
+                    new_data = copy.copy(uv_in.get_data((i, j)))
+                    new_flags = copy.copy(uv_in.get_flags((i, j)))
+
                 if pol not in d[(i, j)]:
-                    d[(i, j)][pol] = data[:, nbl, 0, :, ip]
-                    f[(i, j)][pol] = flags[:, nbl, 0, :, ip]
+                    d[(i, j)][pol] = new_data
+                    f[(i, j)][pol] = new_flags
                 else:
                     d[(i, j)][pol] = np.concatenate(
-                        [d[(i, j)][pol], data[:, nbl, 0, :, ip]])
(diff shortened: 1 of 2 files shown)


nkern:
@zakiali this was what we saw when the `model` dictionary only had two baselines in it.. this fixes that. could you review this when you get the chance? it should be relatively painless....

hera_cal/firstcal.py
```diff
@@ -403,7 +403,7 @@ def UVData_to_dict(uvdata_list, filetype='miriad'):
         flags = uv_in.flag_array.reshape(
             uv_in.Ntimes, uv_in.Nbls, uv_in.Nspws, uv_in.Nfreqs, uv_in.Npols)
 
-        for nbl, (i, j) in enumerate(map(uv_in.baseline_to_antnums, uv_in.baseline_array[:uv_in.Nbls])):
+        for nbl, (i, j) in enumerate(map(uv_in.baseline_to_antnums, np.unique(uv_in.baseline_array))):
```
zakiali: np.unique returns a sorted list of unique elements. Are the first Nbls elements necessarily sorted in the baseline array?  I can see this being a problem in the loop since data is being indexed by nbl. If the bl order in the data doesn't correspond to the sorted baseline array of the loop, the dictionaries would have the wrong mapping.

hera_cal/tests/test_firstcal.py
```diff
@@ -195,6 +195,14 @@ def test_UVData_to_dict(self):
                 np.testing.assert_equal(f[i, j][pol], np.resize(
                     uvd.flag_array[uvmask][:, 0, :, uvpol], f[i, j][pol].shape))
 
+    def test_UVData_to_dict_keys(self):
+        uvd = UVData()
+        uvd.read_miriad(os.path.join(DATA_PATH, 'zen.2457698.40355.xx.HH.uvcAA'))
+        nt.assert_equal(len(firstcal.UVData_to_dict([uvd])[0]), uvd.Nbls)
+        # reorder baseline array
+        uvd.baseline_array = uvd.baseline_array[np.argsort(uvd.baseline_array)]
```
zakiali: Check for equality instead of assigning to uvd.baseline_array?

nkern:
@jsdillon or josh you could review this too (its less than one line of code in `firstcal`), since you have merge authorization...

hera_cal/tests/test_firstcal.py
nkern: sorry, what do you mean by that? the baseline_array is purposely being reassigned b/c that is what causes the bug: when a baseline array is sorted in ascending order.

hera_cal/firstcal.py
nkern: yeah good point, let me look into that..

hera_cal/tests/test_firstcal.py
zakiali: This was basically getting at a test for my previous comment. But I think since the error comes about because the baseline array is sorted in this way, I'm ok with this test.

hera_cal/firstcal.py
zakiali: We can also get around this by just using the get_data methods for a uvdata object.

nkern: ok, good point, ill try that

nkern: i think that worked well!

hera_cal/firstcal.py
```diff
@@ -397,26 +397,29 @@ def UVData_to_dict(uvdata_list, filetype='miriad'):
...
 
-        for nbl, (i, j) in enumerate(map(uv_in.baseline_to_antnums, uv_in.baseline_array[:uv_in.Nbls])):
+        # iterate over unique baselines
+        for nbl, (i, j) in enumerate(map(uv_in.baseline_to_antnums, np.unique(uv_in.baseline_array))):
             if (i, j) not in d:
                 d[i, j] = {}
                 f[i, j] = {}
             for ip, pol in enumerate(uv_in.polarization_array):
                 pol = pol2str[pol]
+                try:
```
zakiali: we can get rid of the try except by using uv_in.get_data(i, j, pol).

nkern: ah, cool, didn't even know you could do that..

PR #123, opened 2017-11-22
Fix uvfits coords

plaplant:
This PR fixes how antenna coordinates are saved in the model visibility files generated by omnical. Proper coordinate handling requires a coordinate transformation, a change in units, and making antenna positions relative to the telescope location. Note that for instances where a calfile is used, rather than the information inside of the file itself, the antenna positions are not guaranteed to be correct. However, antenna positions were not correct before, and so there is no net change.


hera_cal/omni.py
```diff
@@ -823,16 +825,59 @@ def make_uvdata_vis(aa, m, v, xtalk=False):
...
+
+    # Save antenna information.
+    if len(ants_telescope) < uv.Nants_data:
+        # Not enough valid antenna positions; default to antennas in data
+        # and give them positions of 0.
+        uv.Nants_telescope = uv.Nants_data
+        uv.antenna_numbers = np.asarray(sorted(ants_data))
+        uv.antenna_positions = np.zeros((uv.Nants_data, 3))
+    else:
+        # Save antenna positions we computed.
```
zakiali: So for the current way we run things we'll have 143 (I think that's the largest antenna number so far?) antenna positions, antenna numbers, and Nants_telescope = 143?

hera_cal/omni.py
```diff
@@ -823,16 +825,59 @@ def make_uvdata_vis(aa, m, v, xtalk=False):
...
+            antpos[idx, :] = rel_pos
+
+            # also save antenna number to list of antennas
+            ants_telescope.append(iant)
+
+            # increment counter
+            idx += 1
+
+    # Save antenna information.
+    if len(ants_telescope) < uv.Nants_data:
```
zakiali: Would we ever run into this situation with the way things are currently run?

hera_cal/omni.py
plaplant: Not quite. pyuvdata doesn't require you to have 1..ant_max antennas in the array. Nants_telescope for current data is 47, and there are 47 antenna numbers/positions. The mapping between antenna number and position is given by enumerating uv.antenna_numbers.

hera_cal/omni.py
plaplant: We shouldn't. I left this here primarily for backwards compatibility with previous tests, but it also ensures we produce sensible UVData objects where Nants_telescope >= Nants_data (which should always be true).

hera_cal/omni.py
zakiali: But you're looping over the antenna array which has length max(antenna_number). Or do antennas not in the array have antenna positions far from the size of the earth?

plaplant: Yes, antennas not in the array are not "far" from the center of the Earth in absolute coordinates (typically a few meters, based on the default value of (-1, -1, -1) that then gets rotated to a different frame). So we loop over all antennas in the AntennaArray, but only real antennas will get added to the `ants_telescope` array, with a corresponding entry in `antpos`. The length of `ants_telescope` defines `Nants_telescope`, so right now we get 47.

Issue #134, opened 2018-01-16
update redcal get_pos_reds to put the precision factor in physical units


jsdillon:

jsdillon:
Also add in xants that takes in [(1,'x'),(2,'y)] etc. and removes them from the reds (even if they are in antpos)

jsdillon:
Addressed in #136.

PR #136, opened 2018-01-17
Update get_reds to include clearer position tolerance and ex_ants functionality.

jsdillon:
This PR addresses #134.


nkern:
this looks great, and its fast! one thing is that the `get_pos_reds` returns antenna pairs with (hi, low) ordering sometimes and (low, hi) ordering. It seems standard to stick to (low, hi) ordering.

jsdillon:
By design, all baseline pairs in a group should have the same orientation. I think this is really useful, personally. The documentation says how the orientation is selected, though there was a typo which I've fixed.

jsdillon:
I could add a default=False keyword argument to get_pos_reds to do it differently if you'd find that useful.

nkern:
I think all baselines can have the same orientation, and still have all keys (low, hi), no? its just that in our UVData objects, I think all the baselines are (low, hi) format...

nkern:
this looks great. thanks for adding in low_hi capability. i'm ready to merge this whenever possible

PR #138, opened 2018-01-18
added __setitem__ to datacontainer and added tests

nkern:
to make datacontainer more flexible

### DIFF
--- a/hera_cal/datacontainer.py
+++ b/hera_cal/datacontainer.py
@@ -68,7 +68,7 @@ class DataContainer:
                 if key in self.keys():
                     self._data[key] = value
                 else:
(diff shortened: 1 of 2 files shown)


hera_cal/datacontainer.py
```diff
@@ -10,7 +10,7 @@ def __init__(self, data):
             data (dict): dictionary of visibilities with keywords of pol/ant pair
                 in any order.
         """
-        self._data = {}
+        self._data = odict()
```
jsdillon: This is fine, but if you input a non-ordered dictionary with length == 3 keys, line 24 will turn it back into a standard dict.

PR #140, opened 2018-01-19
Abscal edits

nkern:
adds the following capabilities to the `abscal` module:
1. use with `omnical` model visibilities as data vector
2. creates the command-line drive script for RTP-like usage


nkern:
finally, it passes...

hera_cal/abscal.py
```diff
@@ -72,9 +72,8 @@ def __init__(self, model, data, wgts=None, antpos=None, freqs=None, pol_select=N
         model : visibility data of refence model, type=dictionary
                 keys are antenna-pair + polarization tuples, Ex. (1, 2, 'xx').
                 values are complex ndarray visibilities.
-                these must be at least 2D arrays, with [0] axis indexing time
-                and [1] axis indexing frequency. If the arrays are 3D arrays, the [2] axis
-                should index polarization, in which case the key loses its pol entry, Ex. (1, 2).
+                these must be 2D arrays, with [0] axis indexing time
+                and [1] axis indexing frequency.
  
                 Optionally, model can be a path to a miriad file, or a pyuvdata.UVData object
```
jsdillon: If it's a pyuvdata object, need the input be miriad?

hera_cal/abscal_funcs.py
```diff
@@ -643,10 +757,11 @@ def array_axis_to_data_key(data, array_index, array_keys, key_index=-1, copy_dic
         return new_data
 
 
-def UVData2AbsCalDict(filenames, pol_select=None, pop_autos=True, return_meta=False):
+def UVData2AbsCalDict(filenames, pol_select=None, pop_autos=True, return_meta=False, filetype='miriad',
+                        pick_data_ants=True):
     """
     turn pyuvdata.UVData objects or miriad filenames 
-    into the dictionary form that AbsCal requires. This format is
```
jsdillon: `filenames` is a confusing command line argument, since you take uvdata objects and also (lists of) miriad or uvfits files.

hera_cal/abscal_funcs.py
```diff
@@ -661,27 +776,46 @@ def UVData2AbsCalDict(filenames, pol_select=None, pop_autos=True, return_meta=Fa
 
     return_meta : boolean, if True: also return antenna and unique frequency and LST arrays
 
+    filetype : string, filetype of data if filenames is a string
```
jsdillon: Explicitly mention the valid options. Also say that this is option is ignored if the data is not a string or list of strings.

hera_cal/abscal.py
jsdillon: Looks like the docstring is out of date. Seems like you can handle uvfits too.

hera_cal/abscal.py
```diff
@@ -118,17 +117,22 @@ def __init__(self, model, data, wgts=None, antpos=None, freqs=None, pol_select=N
...
+        # set pols to None
         pols = None
+
         # check format of model
         if type(model) == list or type(model) == np.ndarray or type(model) == str or type(model) == UVData:
-            (model, model_flags, model_antpos, model_ants, model_freqs,
+            (model, model_flags, model_antpos, model_ants, model_freqs, model_lsts,
              model_times, model_pols) = UVData2AbsCalDict(model, pop_autos=True, return_meta=True, pol_select=pol_select)
 
         # check format of data
```
jsdillon: Clarify in comments that this is for when the model is not a dictionary already.

hera_cal/abscal.py
```diff
@@ -118,17 +117,22 @@ def __init__(self, model, data, wgts=None, antpos=None, freqs=None, pol_select=N
         pol_select : list of polarizations you want to keep in data
                      type=list, dtype=str, Ex. ['xx', 'yy']
         """
+        # set pols to None
         pols = None
+
         # check format of model
```
jsdillon: Clarify in comments that this is for when the model is not a dictionary already.

hera_cal/abscal.py
```diff
@@ -315,13 +318,80 @@ def delay_lincal(self, medfilt=True, kernel=(1, 11), time_ax=0, freq_ax=1, verbo
...
 
         if solve_offsets:
-            self._ant_dly_phi = odict(map(lambda k: (k, copy.copy(fit["phi_{}_{}".format(k[0],k[1])])), self._flatten(self._gain_keys)))
+            self._ant_dly_phi = odict(map(lambda k: (k, copy.copy(fit["phi_{}_{}".format(k[0],k[1])])), flatten(self._gain_keys)))
             self._ant_dly_phi_arr = np.moveaxis(map(lambda pk: map(lambda k: self._ant_dly_phi[k], pk), self._gain_keys), 0, -1)
 
+    def delay_slope_lincal(self, medfilt=True, kernel=(1, 11), time_ax=0, freq_ax=1, verbose=True, time_avg=False,
+                            four_pol=False):
+        """
+        Solve for per-antenna delay according to the equation
```
jsdillon: Update documentation. It's really unclear how this is different from the other delay lincal.

hera_cal/abscal.py
```diff
@@ -532,9 +602,63 @@ def ant_dly_phi_gain_arr(self):
...
+            return copy.deepcopy(self._dly_slope)
+        else:
+            return None
+
+    @property
+    def dly_slope_gain(self):
+        """ form complex gain from _dly_slope dict """
+        if hasattr(self, '_dly_slope'):
+            dly_slope = self.dly_slope
+            return odict(map(lambda k: (k, np.exp(2j*np.pi*self.freqs.reshape(1, -1)*np.einsum("i...,i->...", dly_slope[k], self.antpos[k[0]][:2]))), flatten(self._gain_keys)))
```
jsdillon: This is line is too long. Also, can you add some comments to explain what's going on here...it's really hard to read.

hera_cal/abscal.py
```diff
@@ -532,9 +602,63 @@ def ant_dly_phi_gain_arr(self):
...
+        else:
+            return None
+
+    def custom_dly_slope_gain(self, gain_keys, antpos):
```
jsdillon: This is line is too long. Also, can you add some comments to explain what's going on here...it's really hard to read.

hera_cal/abscal.py
```diff
@@ -532,9 +602,63 @@ def ant_dly_phi_gain_arr(self):
...
+        """
+        write dly_slope_gain with custom gain keys
+
+        gain_keys : type=list, list of unique (ant, pol). Ex. [(0, 'x'), (1, 'x'), (0, 'y'), (1, 'y')]
+        antpos : type=dictionary, contains antenna position vectors. keys are ant integer, values are ant positions
+        """
+        if hasattr(self, '_dly_slope'):
+            dly_slope = self.dly_slope[self._gain_keys[0][0]]
+            return odict(map(lambda k: (k, np.exp(2j*np.pi*self.freqs.reshape(1, -1)*np.einsum("i...,i->...", dly_slope, antpos[k[0]][:2]))), gain_keys))
```
jsdillon: This is line is too long. Also, can you add some comments to explain what's going on here...it's really hard to read.

hera_cal/abscal.py
```diff
@@ -532,9 +602,63 @@ def ant_dly_phi_gain_arr(self):
...
+        if hasattr(self, '_dly_slope_arr'):
+            return copy.copy(self._dly_slope_arr)
+        else:
+            return None
+
+    @property
+    def dly_slope_gain_arr(self):
+        """ form complex gain from _dly_slope_arr array """
+        if hasattr(self, '_dly_slope_arr'):
+            return np.exp(2j*np.pi*self.freqs.reshape(-1, 1)*np.einsum("hi...,hi->h...", self._dly_slope_arr, self.antpos_arr[:, :2]))
```
jsdillon: This is line is too long. Also, can you add some comments to explain what's going on here...it's really hard to read.

hera_cal/abscal.py
```diff
@@ -315,13 +318,80 @@ def delay_lincal(self, medfilt=True, kernel=(1, 11), time_ax=0, freq_ax=1, verbo
...
+
+        Parameters:
+        -----------
+        medfilt : boolean, if True median filter data before fft
+
+        kernel : size of median filter across (time, freq) axes, type=(int, int)
+
+        time_avg : boolean, if True, average resultant antenna delays across time 
+
+        four_pol : boolean, if True, form a join polarization solution
```
jsdillon: typo: "joint"

hera_cal/abscal.py
```diff
@@ -632,60 +793,216 @@ def TT_Phi_gain_arr(self):
             return None
 
 
-''' TO DO
-    def smooth_data(self, data, flags=None, kind='linear'):
-        """
+def abscal_arg_parser():
+    a = argparse.ArgumentParser(description="")
```
jsdillon: Add a description.

hera_cal/abscal.py
```diff
@@ -632,60 +793,216 @@ def TT_Phi_gain_arr(self):
...
+
+
+def abscal_run(data_files, model_files, pol_select=None, verbose=True, overwrite=False,
+               write_calfits=True, calfits_fname=None, return_gains=False, outdir=None,
+               match_red_bls=False, tol=1.0, reweight=False, interp_model=True, alt_gains=False,
+               delay_slope_cal=True, abs_amp_cal=True, TT_phs_cal=True,
+               delay_cal=False, avg_phs_cal=False, gen_amp_cal=False, gen_phs_cal=False):
+    """
+    run AbsCal on a set of time-contiguous data miriad files, using 
+    time-contiguos model miriad files that cover the data_files across LST
```
jsdillon: typo: "contiguos"

hera_cal/abscal.py
nkern: I agree these lines are long and a little cumbersome. It isn't very necessary for them to be "readable" per-se because they are not really exposed to the user, but better documentation will certainly help readability. The reason why I have it compressed to a single line is so that I don't have to perform unnecessary FOR loops.

hera_cal/abscal.py
```diff
@@ -632,60 +793,216 @@ def TT_Phi_gain_arr(self):
...
 
-        # interpolate flagged data
+    match_red_bls : type=boolean, match unique data baselines to model baselines based on redundancy
 
+    reweight : type=boolean, reweight unique baseline data based on redundancy
 
-        # sphere training data x-values
+    interp_model : type=boolean, interpolate model freq and time onto data freq and time
 
+    alt_gains : type=boolean, if True, use full antenna list in data file to make gains,
```
jsdillon: I find this description confusing. What exactly is the use case here?

hera_cal/abscal.py
```diff
@@ -632,60 +793,216 @@ def TT_Phi_gain_arr(self):
...
+                rather than just antennas present in the raw data. It is not possible
+                to run delay_cal, avg_phs_cal, gen_phs_cal and gen_amp_cal when alt_gains is True.
 
-        # ravel training data
+    delay_slope_cal : type=boolean, if True, perform delay slope calibration
```
jsdillon: These booleans (which kinds of cal to run) should be in a separate section of the docstring from the other parameters. It should be clear that these calibrations are *both* run and applied. Also, they should appear exactly in the same order as they are run, to make it easier to understand what exactly is happening.

hera_cal/abscal.py
```diff
@@ -632,60 +793,216 @@ def TT_Phi_gain_arr(self):
...
+
+        # write to file
+        if write_calfits:
+            gains2calfits(calfits_fname, gain_dict, data_freqs, data_lsts, AC.gain_pols, overwrite=overwrite)
+
+        # append gain dict to gains
+        gains.append(gain_dict)
+
+    # return gains if desired
+    if return_gains:
```
jsdillon: I'd also include the option to return the abscal object.

hera_cal/abscal.py
```diff
@@ -632,60 +793,216 @@ def TT_Phi_gain_arr(self):
...
+            if alt_gains:
+                raise ValueError("can't run delay_cal when alt_gains is True")
+            AC.delay_lincal(verbose=verbose, time_avg=True)
+            AC.data = apply_gains(AC.data, (AC.ant_dly_gain, AC.ant_dly_phi_gain))
+            gain_list.append(AC.ant_dly_gain)
+            gain_list.append(AC.ant_dly_phi_gain)
+
+        if avg_phs_cal:
+            if delay_cal == False:
+                echo("it is recommended to run a delay_cal before avg_phs_cal", verbose=verbose)
```
jsdillon: These might rise to the level of warnings rather than echos. Your call.

hera_cal/abscal.py
jsdillon: An acceptable solution is to just add a linebreak (you don't have to use for loops) and to add a comment explaining exactly what's going on. You could also separate the exponentiation from the einsum.

hera_cal/abscal.py
```diff
@@ -578,7 +714,19 @@ def abs_psi_gain(self):
...
+
+    def custom_abs_psi_gain(self, gain_keys):
+        """
+        write abs_psi_gain with custom gain keys
+
+        gain_keys : type=list, list of unique (ant, pol). Ex. [(0, 'x'), (1, 'x'), (0, 'y'), (1, 'y')]
+        """
+        if hasattr(self, '_abs_psi'):
+            abs_psi = self.abs_psi[self._gain_keys[0][0]]
+            return odict(map(lambda k: (k, np.exp(abs_psi).astype(np.complex)), gain_keys))
```
jsdillon: Where's the 1.0j in the exponent?

hera_cal/abscal.py
nkern: for a `omnical` model visibility data. The antennas present only in the omnical model data are only those that form unique baselines, but the file still contains information on all of the antennas actually present in the array. When we want to actually create a gain solution to apply to the data, we want to form a gain solution using all of these antennas, not just the ones present in the data.

hera_cal/abscal.py
nkern: I don't want to manually restrict the kinds of analysis the user can do, even if to us it seems odd. that seems unneccessarily restrictive.

hera_cal/abscal.py
nkern: good call, thanks!

hera_cal/abscal.py
nkern: ah sorry, you said warnings--i read errors for some reason.. yeah warnings and echoes seem the same to me, unless the user insists on turning verbose off (which is set to on by default)

hera_cal/abscal.py
jsdillon: Ok..."alt gains" is a strange name for this...perhaps we can brainstorm a better one.

nkern: changed to `full_antenna_gains`, which is more inline with what its doing

jsdillon:
General question, why are the calfits and uvdata files in the repo changing? It's not a huge deal, but we should change them as little as possible to avoid blowing up the repo.

hera_cal/abscal_funcs.py
```diff
@@ -661,29 +776,49 @@ def UVData2AbsCalDict(filenames, pol_select=None, pop_autos=True, return_meta=Fa
...
+            suffix = os.path.splitext(datanames)[1]
+            if filetype == 'uvfits' or suffix == '.uvfits':
+                uvd.read_uvfits(datanames)
+                uvd.unphase_to_drift()
+            elif filetype == 'miriad':
+                uvd.read_miriad(datanames)
+
         else:
-            uvd = filenames
+            uvd = datanames
```
jsdillon: You might want to put some assert statements in here (like assert type is pyuvdata) so that it's easier for people to debug if they use this function incorrectly.

nkern:
@jsdillon the files are changing because I accidentally made subset arrays of bad antennas, so I had to change them. They are small files, so its not too big of an issue, but I agree we should limit this if we can.

hera_cal/abscal_funcs.py
nkern: added comments to make this clearer...

hera_cal/abscal_funcs.py
```diff
@@ -901,10 +1077,9 @@ def gains2calfits(calfits_fname, abscal_gains, freq_array, time_array, pol_array
 
     Parameters:
     -----------
-    calfits_fname : string, path and filename to output calfits file
+    calfits_fname : type=string, path and filename to output calfits file
 
-    abscal_gains : dictionary, antenna integer as key, ndarray complex gain
-        as values with shape (Ntimes, Nfreqs, Npols)
+    abscal_gains : type=dictionary, (ant, pol) as key, (Ntimes, Nfreqs) complex ndarray as value
```
jsdillon: gain_convention and overwrite should be in the list of parameters

hera_cal/abscal_funcs.py
```diff
@@ -944,6 +1120,7 @@ def gains2calfits(calfits_fname, abscal_gains, freq_array, time_array, pol_array
         echo("saving {}".format(calfits_fname))
         uvc.write_calfits(calfits_fname, clobber=overwrite)
 
+
```
jsdillon: fill_dict_nans seems unnecessarily long. Is it worthwhile to try to accomplish it more succinctly without so many if statements?

hera_cal/abscal_funcs.py
```diff
@@ -1059,201 +1213,124 @@ def __eq__(self, B2):
...
-        # construct GP kernel
-        ls = np.array([1e-1 for i in range(ndim)])
-        kernel = 1.0**2 * gaussian_process.kernels.RBF(ls, np.array([1e-2, 1e1])) + \
-                 gaussian_process.kernels.WhiteKernel(1e-4, (1e-8, 1e1))
-        model = gaussian_process.GaussianProcessRegressor(kernel=kernel, n_restarts_optimizer=n_restart)
-        # fit GP
-        model.fit(Xtrain, ytrain)
-
-    ypred = model.Predict(X)
+    data : type=dictionary, data dictionary holding complex visibilities.
```
jsdillon: Parameters are out of order (relative to the function definition)

hera_cal/abscal_funcs.py
```diff
@@ -1059,201 +1213,124 @@ def __eq__(self, B2):
...
     if flags is None:
-        flags = copy.deepcopy(data)
-        for k in flags.keys(): flags[k] = np.zeros_like(flags[k]).astype(np.bool)
+        flags = DataContainer(odict(map(lambda k: (k, np.zeros_like(data[k]).astype(np.bool)), data.keys())))
 
     # get redundant baselines
-    reds = compute_reds(bls, antpos, tol=tol)
+    _reds = redcal.get_reds(antpos, bl_error_tol=tol, pols=pols, low_hi=True)
+
+    # strip reds of keys not in data
```
jsdillon: Low priority, but I think this can be accomplished in a line or two with list comprehension.

hera_cal/abscal_funcs.py
```diff
@@ -1059,201 +1213,124 @@ def __eq__(self, B2):
...
+            elif np.isclose(np.arccos(dot), np.pi, atol=tol/self.len) or (dot < -1.0):
                 return 'conjugated'
             # else return False
             else:
                 return False
         else:
             return False
 
-def smooth_data(Xtrain, ytrain, Xpred, kind='gp', n_restart=3, degree=1, ls=10.0, return_model=False):
+def match_red_baselines(model, model_antpos, data, data_antpos, tol=1.0, verbose=True):
```
jsdillon: This function (and the baseline class) seem too general and complicated. More importantly, we need to figure out how to properly propagate weights.

hera_cal/abscal_funcs.py
```diff
@@ -1267,28 +1344,53 @@ def avg_data_across_red_bls(data, bls, antpos, flags=None, broadcast_flags=True,
...
                             write_miriad=True, output_data=False, overwrite=False, 
                             verbose=True, **kwargs):
     """
+    Open a file and run avg_data_across_red_bls on data, then write to file
+
+    Parameters:
+    -----------
+    data_fname : type=str, path to miriad file
+
+    outdir : type=str, output directory
```
jsdillon: If None, does it just not write?

hera_cal/tests/test_utils.py
```diff
@@ -69,4 +71,25 @@ def test_LST2JD(self):
         nt.assert_almost_equal(utils.LST2JD(12.0, 2458042, 21.), 2458042.8720297855)
 
 
+class Test_combine_calfits:
```
jsdillon: I'd like to see some more verification tests just asserting that the file exists and the number of antennas is right. I'd like to see that the number of times is correct and that the gain from a particular channel and integration from test_file1 and test_file2 actually ends up in the right place.

hera_cal/abscal_funcs.py
nkern: its fast right now, so no need from a computation perspective, and from an aesthetic perspective, I don't think its an issue

hera_cal/abscal_funcs.py
nkern: fixed

hera_cal/abscal_funcs.py
nkern: agreed

hera_cal/abscal_funcs.py
nkern: if none, it uses the current working directory

hera_cal/tests/test_abscal.py
```diff
@@ -131,10 +145,78 @@ def test_gains2calfits(self):
...
+        nt.assert_equal(d[(24,37,'xx')].shape, (60, 64))
+        # test w/o medfilt
+        d = hc.abscal.wiener(self.data, window=(5, 15), medfilt=False)
+        nt.assert_equal(d[(24,37,'xx')].shape, (60, 64))
+        # test as array
+        d = hc.abscal.wiener(self.data[(24,37, 'xx')], window=(5, 15), medfilt=False, array=True)
+        nt.assert_equal(d.shape, (60, 64))
+        nt.assert_equal(d.dtype, np.complex)
+
+    def test_Baseline(self):
```
jsdillon: This test seems incomplete. Can you make it show that it works on baselines created from different pairs of antennas? Also, can you show that it works on baselines that are similar to just within the tolerance and just outside of it?

hera_cal/tests/test_utils.py
nkern: done

hera_cal/tests/test_abscal.py
```diff
@@ -131,10 +145,78 @@ def test_gains2calfits(self):
...
+        nt.assert_true((25, 38, 'xx') in data)
+
+    def test_mirror_data_to_red_bls(self):
+        # make fake data
+        reds = hc.redcal.get_reds(self.antpos, pols=['xx'])
+        data = DataContainer(odict(map(lambda k: (k[0], self.data[k[0]]), reds[:5])))
+        # test execuation
+        d = hc.abscal.mirror_data_to_red_bls(data, self.antpos)
+        nt.assert_equal(len(d.keys()), 16)
+        nt.assert_true((24, 25, 'xx') in d)
```
jsdillon: Also test that the correct value gets propagated.

hera_cal/tests/test_abscal.py
```diff
@@ -131,10 +145,78 @@ def test_gains2calfits(self):
...
+        nt.assert_equal(bls[0], bls[0])
+        nt.assert_false(bls[0] == bls[1])
+        nt.assert_equal(bls[0] == bls_conj[0], 'conjugated')
+
+    def test_match_red_baselines(self):
+        model = copy.deepcopy(self.data)
+        model = odict([((k[0]+1, k[1]+1, k[2]), model[k]) for i,k in enumerate(model.keys())])
+        model_antpos = odict([(k+1, self.antpos[k]) for i,k in enumerate(self.antpos.keys())])
+        data = hc.abscal.match_red_baselines(self.data, self.antpos, model, model_antpos, tol=2.0, verbose=False)
+        nt.assert_equal(len(data.keys()), 9)
```
jsdillon: I'd like to see some more tests here showing that specific visibilities line up when they should and other that shouldn't do not line up

hera_cal/abscal_funcs.py
```diff
@@ -1059,201 +1216,125 @@ def __eq__(self, B2):
...
 
-        # ensure data keys match model keys
-        for i, k in enumerate(data):
-            if k not in model:
-                data.pop(k)
+def avg_data_across_red_bls(data, antpos, flags=None, broadcast_flags=True, median=False, tol=1.0,
+                            mirror_red_data=False):
+    """
+    Given complex visibility data spanning one or more redundant
+    baseline groups, average redundant visibilities and write to file
```
jsdillon: I think this docstring is out of date. You don't write to file here.

hera_cal/tests/test_abscal.py
```diff
@@ -131,10 +145,78 @@ def test_gains2calfits(self):
...
+    def test_flatten(self):
+        l = hc.abscal.flatten([['hi']])
+        nt.assert_equal(np.array(l).ndim, 1)
+
+    def test_avg_data_across_red_bls(self):
+        data, flags, antpos, ants, freqs, times, lsts, pols = hc.abscal.UVData2AbsCalDict(self.data_file, return_meta=True)
+        rd, rf, rk = hc.abscal.avg_data_across_red_bls(data, antpos, flags=self.wgts, tol=2.0)
+        rd, rf, rk = hc.abscal.avg_data_across_red_bls(data, antpos, tol=2.0, median=True)
+        rd, rf, rk = hc.abscal.avg_data_across_red_bls(data, antpos, tol=2.0, broadcast_flags=True)
+        nt.assert_equal(len(rd.keys()), 9)
```
jsdillon: Again, I think you you need to test a few visibilities to see if this worked properly.

jsdillon:
Ok, that's everything. Nice work! LMK when you're done and I'll take one final look before merging in.

hera_cal/tests/test_abscal.py
nkern: good call. done

hera_cal/abscal_funcs.py
nkern: yup

hera_cal/tests/test_abscal.py
nkern: yup

hera_cal/abscal_funcs.py
nkern: yup

hera_cal/tests/test_abscal.py
nkern: yup

hera_cal/tests/test_abscal.py
nkern: yup

nkern:
yeah, i couldn't find a way to easily account for the 2nd argument that standard pop takes. if you want to give it a shot feel free ;) 

I think the _bls and _pols calls aren't actually that slow (in the hundreds of microseconds regime), and also one needs to look at every key because you could for example pop (1,2,'xx') but (1,2,'yy') still exists and therefore (1,2) needs to still be in _bls. Likewise, if you pop (1,2,'xx') but say (3,4,'xx') still exists, than 'xx' should still be in _pols.

yeah i know generally pop() returns something, but I actually dislike that feature of the standard pop function :)

jsdillon:
I guess it's not worth pre-optimizing speed for now. However, as written, you're really should make this an overload of __delitem__, not pop. Pop is explicitly for returning something (it comes from pushing and popping in stack data structures, which you really don't see much of in python).

https://docs.python.org/2/library/operator.html#mapping-operators-to-functions

nkern:
yeah point well taken.

PR #152, opened 2018-01-27
adds lstbin module to hera_cal

nkern:
This PR adds the lstbin module to hera_cal. Currently, this module introduces two capabilities:

lst_align, or interpolating visibility data to align with a universal LST grid
lst_bin, or binning data into a universal LST grid
Also includes some edits to utils and datacontainer to accomodate new capabilities. Also includes a scripts/notebooks/running_lstbin.ipynb for basic tutorial on LST binning, as well as a drive script lstalign_run.py and lstbin_run.py.


jsdillon:
@zakiali Shall we meet tomorrow to go over the PR? I think it was really helpful to do the last one together.

hera_cal/abscal_funcs.py
```diff
@@ -758,7 +758,7 @@ def array_axis_to_data_key(data, array_index, array_keys, key_index=-1, copy_dic
 
 
 def UVData2AbsCalDict(datanames, pol_select=None, pop_autos=True, return_meta=False, filetype='miriad',
-                        pick_data_ants=True):
+                      pick_data_ants=True, return_wgts=False):
```
jsdillon: Return wgts needs documentation.

hera_cal/abscal_funcs.py
```diff
@@ -793,9 +793,9 @@ def UVData2AbsCalDict(datanames, pol_select=None, pop_autos=True, return_meta=Fa
     antpos : dictionary containing antennas numbers as keys and position vectors
     ants : ndarray containing unique antennas
     freqs : ndarray containing frequency channels (Hz)
-    times : ndarray containing julian date bins of data
-    lsts : ndarray containing LST bins of data (radians)
-    pols : ndarray containing list of polarization index integers
+    times : ndarray containing time stamps data in julian date
+    lst : ndarray containing time stamps in local sidereal time
```
zakiali: radians? hours? does it matter?

hera_cal/datacontainer.py
```diff
@@ -84,12 +85,24 @@ def __delitem__(self, key):
         else:
             raise ValueError('only supports setting (ant1, ant2, pol) keys')
 
-  
+    def __add__(self, D):
```
zakiali: So this just grows visibilities in time? Do we want to support any other add method? Also we should we check that all the keys between D and self match otherwise this will fail.

zakiali: Also add a doc string

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+    data_avg : dictionary of data having averaged the LST bins
+
+    wgts_avg : dictionary of data weights averaged in LST bins
+    
+    data_std : dictionary of data with real component holding std along real axis
+        and imag component holding std along imag axis
+
+    lst_bins : ndarray containing final lst grid of data
+
+    data_num : dictionary containing the number of data points averaged in each LST bin.
```
zakiali: can we call this data_count instead? data_num doesn't convey count values to me. 
I would also rearrange outputs lst_bins comes out before data_std.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+        data_bin : dictionary with (ant1,ant2,pol) as keys and ndarrays holding
+            un-averaged complex visibilities in each LST bin as values. 
+    """
+    # construct lst_grid if not provided
+    if lst_grid is None:
+        # get dlst if not provided
+        if dlst is None:
+            dlst = np.median(np.diff(lst_list[0]))
+        # construct lst_grid
+        lst_grid = np.arange(lst_init, lst_init + wrap_point - dlst/2, dlst) + dlst/2
```
zakiali: Why does the default go from pi -> 3pi? we don't seem to be wrapping at the wrap point in this scheme. 0 to 2pi makes more sense.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+
+    # restrict lst_grid based on lst_low and lst_high
+    if lst_low is not None:
+        lst_grid = lst_grid[lst_grid > lst_low - atol]
+    if lst_hi is not None:
+        lst_grid = lst_grid[lst_grid < lst_hi + atol]
+
+    # move lst_grid centers to the left
+    lst_grid_left = lst_grid - dlst / 2
+    lst_grid_left = unwrap(lst_grid_left)
```
jsdillon: Should this pass the wrap_point through?

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+    wgts = odict()
+    all_lst_indices = set()
+
+    # iterate over data_list
+    for i, d in enumerate(data_list):
+        # get lst array
+        l = lst_list[i]
+
+        # unwrap LST array
+        l = unwrap(l)
```
jsdillon: Should this pass the wrap_point through?

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+import gc as garbage_collector
+import datetime
+
+
+def lst_bin(data_list, lst_list, lst_grid=None, wgts_list=None, lst_init=np.pi, dlst=None,
+            lst_low=None, lst_hi=None, wrap_point=2*np.pi, atol=1e-8, median=False,
+            sigma_clip=False, sigma=2.0, return_no_avg=False, verbose=True):
+    """
+    Bin data in Local Sidereal Time (LST) onto an LST grid. An LST grid
+    is defined as an array of points increasing in Local Sidereal Time, with each point marking
```
zakiali: Good to make a comment here that this function is unit agnostic.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+    std = np.sqrt(astats.biweight_midvariance(array, axis=axis))
+
+    # set cut data to nans
+    cut = np.where(np.abs(array-mean)/std > sigma)
+    array[cut] *= np.nan
+
+    return array
+
+
+def unwrap(array, wrap_point=2*np.pi):
```
zakiali: Why don't we just use numpy.unwrap for this. You can add a wrap point as a keyword in the function.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+
+    Output:
+    -------
+    new_array : type=ndarray, 1D array with unwrapped LST values
+    """
+    array = copy.copy(array)
+    new_array = np.empty_like(array)
+    for i, v in enumerate(array):
+        if i == 0:
+            while v < 0:
```
jsdillon: This function seems to be able to add wrap_point to array[0] many times but can only add it once to array[1:]. Perhaps you need a check to make sure all data are within a wrap_point of one another.

Also, it's a bit strange to have an if i == 0 in the loop...why not do it outside the loop?

Also, why not use np.unwrap?

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+        # construct lst_grid
+        lst_grid = np.arange(lst_init, lst_init + wrap_point - dlst/2, dlst) + dlst/2
+    else:
+        dlst = np.median(np.diff(lst_grid))
+        lst_init = lst_grid[0] - dlst/2
+        if lst_init < 0:
+            lst_init += wrap_point
+
+    # restrict lst_grid based on lst_low and lst_high
+    if lst_low is not None:
```
jsdillon: I don't think this works well with a wrap point. What if you want dataset that includes 6 radians through 2pi and then 0 radians to 1 radian? Low would be 6 and the high would be 1, but that would throw out all the data, no?

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+
+        # add wrap_point to l where less than lst_init
+        l[np.where(l < lst_init)] += wrap_point
+
+        # digitize data lst array "l", moving bin centers to the left
+        grid_indices = np.digitize(l, lst_grid_left[1:], right=True)
+
+        # make data_in_bin boolean array, and set to False data that don't fall in any bin
+        data_in_bin = np.ones_like(l, np.bool)
+        data_in_bin[(l<lst_grid_left.min()-atol)|(l>lst_grid_left.max()+dlst+atol)] = False
```
jsdillon: For readability, this should probably be two lines.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+        if v >= wrap_point:
+            wrap_diff = (v // wrap_point) * wrap_point
+            v -= wrap_diff
+            array[i:] -= wrap_diff
+        new_array[i] = v
+
+    return new_array
+
+
+def switch_bl(key):
```
jsdillon: Do you still need this if you're using DataContainers?

hera_cal/lstbin.py
```diff
@@ -0,0 +1,893 @@
...
+        data_in_bin = np.ones_like(l, np.bool)
+        data_in_bin[(l<lst_grid_left.min()-atol)|(l>lst_grid_left.max()+dlst+atol)] = False
+
+        # update all_lst_indices
+        all_lst_indices.update(set(grid_indices[data_in_bin]))
+
+        # iterate over keys in d
+        for j, key in enumerate(d.keys()):
+            # conjugate d[key] if necessary
+            if key in data:
```
zakiali: All the conjugation stuff seems unnecessary since we are using DataContainers as the input data type.

nkern: can't use DataContainer here, so this is necessary

hera_cal/abscal_funcs.py
nkern: done

hera_cal/datacontainer.py
```diff
@@ -84,12 +92,81 @@ def __delitem__(self, key):
         else:
             raise ValueError('only supports setting (ant1, ant2, pol) keys')
 
+    def __xor__(self, D):
```
jsdillon: Generally, operator overloading only happens when there's some notion of "addition" or whatever on the object. Concatenation across time is not obviously xor (if anything it's +). This notation is highly non-intutive and makes code very difficult to read.  You should make it `append(axis=0)` i.e. default axis is 0 (time), but it can also append along frequency.

hera_cal/datacontainer.py
```diff
@@ -84,12 +92,81 @@ def __delitem__(self, key):
...
+        if D[D.keys()[0]].shape[0] != self.__getitem__(self.keys()[0]).shape[0]:
+            raise ValueError("[0] axis of dictionary values don't match")
+        if D[D.keys()[0]].shape[1] != self.__getitem__(self.keys()[0]).shape[1]:
+            raise ValueError("[1] axis of dictionary values don't match")
+
+        # start new object
+        newD = odict()
+
+        # iterate over D keys
+        for i, k in enumerate(D.keys()):
```
zakiali: Maybe raise a warning if keys in D and self are not the same. It would hard/annoying to keep track of which elements have been added to. The warning brings the users attention to this.

hera_cal/datacontainer.py
```diff
@@ -84,12 +92,81 @@ def __delitem__(self, key):
...
+            if self.__contains__(k):
+                newD[k] = self.__getitem__(k) + D[k]
+
+        return DataContainer(newD)
+
+
+    def __pow__(self, D):
```
jsdillon: Same comment as above. Should be `.average_with(D2)` or something similar. Also, it's strange that you can only average two datacontainers and not N datacontainers.

hera_cal/datacontainer.py
```diff
@@ -84,12 +92,81 @@ def __delitem__(self, key):
...
+
+        # iterate over D keys
+        for i, k in enumerate(D.keys()):
+            if self.__contains__(k):
+                newD[k] = (self.__getitem__(k) + D[k]) / 2.0
+
+        return DataContainer(newD)
+
+
+    def __mul__(self, D):
```
zakiali: maybe add a conjugate somewhere? either as an attribute to datacontainer itself or as an option. This is not necessary, but just suggestion.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,943 @@
...
+    std = np.nanmedian(np.abs(array - mean), axis=axis) * 1.4
+
+    # set cut data to nans
+    cut = np.where(np.abs(array-mean)/std > sigma)
+    array[cut] *= np.nan
+
+    return array
+
+
+def switch_bl(key):
```
jsdillon: Also, you should switch the order of the polarization (and add that to the unit test)

hera_cal/lstbin.py
jsdillon: Bump...we haven't figured this out yet.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,943 @@
...
+            u = np.array([u])
+
+        # get phasor
+        phs = np.exp(-2j*np.pi*freqs[None, :]*u[:, None]/aipy.const.c*100)
+
+        # multiply into data
+        data[k] *= phs
+
+
+def data_to_miriad(fname, data, lst_array, freq_array, antpos, time_array=None, flags=None,
```
jsdillon: Move to utils

Issue #153, opened 2018-01-31
Update calfiles produced by firstcal/omnical to be explicitly "redundancy" type


plaplant:
A recent PR in pyuvdata (https://github.com/HERA-Team/pyuvdata/pull/290) added sky-based calibration as a type to the calfits implementation. The old format is "redundancy" type, and is applicable to firstcal and omnical calibration solutions. We should update the functions for producing these calfiles to explicitly set its type, as well as mark abscal as a "sky-based" type (as well as add the corresponding metadata). It may also mean that tests break in the meantime, but remains to be seen.

plaplant:
This was addressed in PR #155.

jsdillon:
It looks like the unittests are producing warnings because of this issue. I'm reopening it.

nkern:
the warnings are b/c the `calfits` files we have on disk are not written out w/ a cal_style. `pyuvdata` has a backwards compatibility mechanism that assigns it a default if its not written, so its an "expected" warning in that sense. to get rid of the warning, you just need to read it in and write it back out and commit it to the repo.

plaplant:
As @nkern said, we can read, write, and commit the new files. Another possibility is to wrap the calls to read them with the pyuvdata test function that silences (expected) warnings without raising them.

I agree in principle that it's a little sloppy for there to be warnings in tests, but there are quite a few `divide by zero encountered in divide` and `invalid value encountered in divide` errors that were introduced in past PRs and have so far been unaddressed. I'm closing this issue here because the root issue I raised was addressed in PR #155, but you can open a new one about getting rid of errors in tests if you want.

nkern:
I re-wrote them with an appropriate `cal_style` in #160, so this should be completely taken care of.

Issue #144, opened 2018-01-25
Correct Abscal writer to calfits to pass optional kw instead of meta for certain attributes


zakiali:
For example, gain_convention should be passed into the optional keyword args instead of the meta, which ignores it.

PR #155, opened 2018-01-31
accounts for `pyuvdata.UVCal.cal_style` parameter in `cal_formats`

nkern:
adapted cal_formats to account for new UVCal cal_style parameter
also updated abscal_funcs.gains2calfits to account for this
and added tests to cal_formats

### DIFF
--- a/hera_cal/cal_formats.py
+++ b/hera_cal/cal_formats.py
@@ -9,7 +9,7 @@ class HERACal(UVCal):
        This can then be saved to a file, plotted, etc.
     '''
 
-    def __init__(self, meta, gains, flags=None, DELAY=False, ex_ants=[], appendhist='', optional={}):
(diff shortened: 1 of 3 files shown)


hera_cal/cal_formats.py
```diff
@@ -112,10 +110,31 @@ def __init__(self, meta, gains, flags=None, DELAY=False, ex_ants=[], appendhist=
         self.time_array = time
         self.integration_time = meta['inttime']
         self.gain_convention = 'divide'
+        self.set_redundant()
```
plaplant: I am fine with having this be the default value, but we should mention in the documentation that this is the case.

nkern: ok, added it in

nkern:
adresses #153

nkern:
also addresses #144

nkern:
@plaplant whenever you ok w/ this feel free to merge it in

PR #159, opened 2018-02-03
updated doc strings and functions in abscal_funcs

nkern:
PR for edits to `abscal`, part 1 of the LST bin PR effort.

### DIFF
--- a/hera_cal/abscal_funcs.py
+++ b/hera_cal/abscal_funcs.py
@@ -1041,8 +1041,8 @@ def wiener(data, window=(5, 11), noise=None, medfilt=True, medfilt_kernel=(3,9),
 
 
 def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=None,
-                 kx=3, ky=3, s=0, flag_extrapolate=True, medfilt_flagged=True, medfilt_window=(3, 7),
-                 aggressive_flagging=False):
+                 kind='cubic', flag_extrapolate=True, medfilt_flagged=True, medfilt_window=(3, 7),
+                 fill_value=None):
     """
     Interpolate complex visibility model onto the time & frequency basis of
     a data visibility. See below for notes on flag propagation if flags is provided.
@@ -1061,8 +1061,10 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
 
     data_freqs : 1D array of the data freq axis, dtype=float, shape=(Nfreqs,)
 
-    flags : type=DataContainer, dictionary containing model flags. Can also contain model wgts and
-                                will convert appropriately.
+    flags : type=DataContainer, dictionary containing model flags. Can also contain model wgts
+            as floats and will convert to booleans appropriately.
+
+    kind : type=str, kind of interpolation, options=['linear', 'cubic', 'quintic']
 
     medfilt_flagged : type=bool, if True, before interpolation, replace flagged pixels with output from 
                       a median filter centered on each flagged pixel.
@@ -1070,13 +1072,10 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
     medfilt_window : type=tuple, extent of window for median filter across the (time, freq) axes.
                      Even numbers are rounded down to odd number.
 
-    kx : type=int, type of spline to fit to freq axis of data. See scipy.interpolate.RectBivariateSpline
-
-    ky : type=int, type of spline to fit to time axis of data. See scipy.interpolate.RectBivariateSpline
-
-    s : type=int, smoothing factor for interpolation. Default is no smoothing. See scipy.interpolate.RectBivariateSpline
-
     flag_extrapolate : type=bool, flag extrapolated data_lsts if True.
+
+    fill_value : type=float, if fill_value is None, extrapolated points are extrapolated
+                 else they are filled with fill_value.
 
     Output: (new_model, new_flags)
     -------
@@ -1086,9 +1085,12 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
     Notes:
     ------
     If the data has flagged pixels, it is recommended to turn medfilt_flagged to True. This runs a median
-    filter on the flagged pixels and replaces their values with the results, however, they remain as flagged.
-    This happens *before* interpolation. This means that interpolation near these points on unflagged pixels
-    aren't significantly biased by the presence of flagged pixels.
+    filter on the flagged pixels and replaces their values with the results, but they remain flagged.
+    This happens *before* interpolation. This means that interpolation near flagged pixels
+    aren't significantly biased by their presence.
+
+    In general, if flags are fed, flags are propagated if a flagged pixel is a nearest neighbor
+    of an interpolated pixel.
     """
     # make flags
     new_model = odict()
@@ -1140,6 +1142,8 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
                 flow, fhi = find-f_ext, find+f_ext+1
                 ll = 0
                 while True:
+                    # iterate until window has non-flagged data in it
+                    # with a max of 10 iterations
                     if tlow < 0: tlow = 0
                     if flow < 0: flow = 0
                     r_med = np.nanmedian(real[tlow:thi, flow:fhi])
@@ -1177,11 +1181,9 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
         else:
             f = np.zeros_like(real, bool)
         
-        # interpolate
-        real_spl = interpolate.RectBivariateSpline(model_lsts, model_freqs, real, kx=kx, ky=ky, s=s)
-        imag_spl = interpolate.RectBivariateSpline(model_lsts, model_freqs, imag, kx=kx, ky=ky, s=s)
-        interp_real = real_spl(data_lsts, data_freqs)
-        interp_imag = imag_spl(data_lsts, data_freqs)
+        # interpolate 
+        interp_real = interpolate.interp2d(model_freqs, model_lsts, real, kind=kind, copy=False, bounds_error=False, fill_value=fill_value)(data_freqs, data_lsts)
+        interp_imag = interpolate.interp2d(model_freqs, model_lsts, imag, kind=kind, copy=False, bounds_error=False, fill_value=fill_value)(data_freqs, data_lsts)
 
         # flag extrapolation if desired
         if flag_extrapolate:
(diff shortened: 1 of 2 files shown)


hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
...
+                                will convert appropriately.
+
+    medfilt_flagged : type=bool, if True, run a median filter on flagged model pixels
 
-    fill_value : values to put for interpolation points outside training set
-        if None, values are extrapolated
+    medfilt_window : type=tuple, window for median filter across the (time, freq) axes.
 
-    zero_tol : for amplitudes lower than this tolerance, set real and imag components to zero
+    kx : type=int, type of spline to fit to freq axis of data. See scipy.interpolate.RectBivariateSpline
```
jsdillon: Why are we fitting over frequency? Shouldn't the frequencies line up?

Also, is it slower to fit over both time and freq instead of just time (looping over freq)?

hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
 
     data_freqs : 1D array of the data freq axis, dtype=float, shape=(Nfreqs,)
 
-    kind : kind of interpolation method, type=str, options=['linear', 'cubic', ...]
-        see scipy.interpolate.interp2d for details
+    flags : type=DataContainer, dictionary containing model flags. Can also contain model wgts and
```
jsdillon: Explain what happens if you just use the default (None)

hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
...
-    zero_tol : for amplitudes lower than this tolerance, set real and imag components to zero
+    kx : type=int, type of spline to fit to freq axis of data. See scipy.interpolate.RectBivariateSpline
 
-    bounds_error : type=boolean, if True, raise ValueError when extrapolating. If False, extrapolate.
+    ky : type=int, type of spline to fit to time axis of data. See scipy.interpolate.RectBivariateSpline
 
-    flag_extrapolate : flag extrapolated data if True
+    s : type=int, smoothing factor for interpolation. Default is no smoothing. See scipy.interpolate.RectBivariateSpline
+
+    flag_extrapolate : type=bool, flag extrapolated data if True
```
jsdillon: What happens if you have both medfilt_flagged and flag_extrapolate as true?

hera_cal/abscal_funcs.py
nkern: yeah its a "general purpose" tool assuming you perhaps want to interpolate at a different frequency grid than what's provided. in practice for us it has no effect because the freq grid is the same.

jsdillon: Right...but does it slow things down? I don't know whether splines are smart enough to not perform the trival interpolation...

hera_cal/abscal_funcs.py
nkern: `medfilt_flagged` will replace flagged data with a median filter centered at the flagged data point. That happens for all flagged data regardless of where it is in LST and frequency. 

`flag_extrapolate` will flag the boundary of the lst grid if it isn't nested between two data lst points

hera_cal/abscal_funcs.py
nkern: i think be definition it is exact. if it fits a line it uses two nearest neighbor points, a quadratic is three and a cubic is four. so my understanding is that by definition its exact.

not sure how much it may be sped up by doing only 1d interpolation. it is already quite quick (relative to other computations) so its not a practical necessity currently...

hera_cal/abscal_funcs.py
jsdillon: Ok, I see. I was confused.

hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
...
+            f = flags[k][time_nn, freq_nn]
+            # check f is boolean type
+            if np.issubdtype(f.dtype, np.float):
+                f = ~(f.astype(np.bool))
+        else:
+            f = np.zeros_like(real, bool)
+         
+        # median filter flagged data if desired
+        if medfilt_flagged and flags is not None:
+            f_ext = int(medfilt_window[1]/2.)
```
jsdillon: This has different behavior for even and odd medfilt_window sizes. Perhaps you want to clarify that in the documentation.

hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
...
+            t_ext = int(medfilt_window[0]/2.)
+            for find, tind in zip(freq_nn[f], time_nn[f]):
+                tlow, thi = tind-t_ext, tind+t_ext
+                flow, fhi = find-f_ext, find+f_ext
+                if tlow < 0: tlow = 0
+                if flow < 0: flow = 0
```
jsdillon: What about the case where fhi or thi is bigger than the size of the model?

hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
...
+                real[tind, find] = np.nanmedian(real[tlow:thi, flow:fhi])
```
jsdillon: A few questions:

1) Why nanmedian? Do you expect there to be nans in the data?
2) Your medians include flagged pixels. Do you want to set them to nan so they are excluded from the median?
3) What happens if the window is so small that it includes only flagged visibilities?

hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
...
+            if np.issubdtype(f.dtype, np.float):
+                f = ~(f.astype(np.bool))
+        else:
+            f = np.zeros_like(real, bool)
+         
+        # median filter flagged data if desired
+        if medfilt_flagged and flags is not None:
+            f_ext = int(medfilt_window[1]/2.)
+            t_ext = int(medfilt_window[0]/2.)
+            for find, tind in zip(freq_nn[f], time_nn[f]):
```
jsdillon: This is a tricky bit of indexing, so I'd add a comment explaining how this iteration works. I'm pretty sure it's right, but it's definitely hard to read.

hera_cal/abscal_funcs.py
```diff
@@ -1055,56 +1063,83 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs,
...
-        interp_imag = interpolate.interp2d(model_freqs, model_lsts, imag,
-                                           kind=kind, fill_value=np.nan, bounds_error=bounds_error)(data_freqs, data_lsts)
         # set flags
-        f = np.zeros_like(interp_real, dtype=float)
-        if flag_extrapolate:
-            f[np.isnan(interp_real) + np.isnan(interp_imag)] = 1.0
-        interp_real[np.isnan(interp_real)] = fill_value
-        interp_imag[np.isnan(interp_imag)] = fill_value
+        if flags is not None:
+            f = flags[k][time_nn, freq_nn]
```
jsdillon: I do not understand this line at all. 

First off, how are you indexing a 2D flag array with two 2D arrays of indices?

Second, it seems to me like your flags are not being propagated correctly. The propagated flags only come from the nearest neighbor, but each datapoint is an interpolation between multiple neighbors. In theory, it should inherit an OR of the flags from both. 

I think Aaron's suggestion was just to define abs model visibilities such that they had no flags, either by combining days (to get rid of time-dependent) or via delay filtering (to get rid of freq-dependent flags).

hera_cal/abscal_funcs.py
jsdillon: Also clarify that by "can also contain model wgts" you mean if it's floats rather than bools.

hera_cal/abscal_funcs.py
nkern: yeah so `f` is an array containing the nearest neighbor (NN) flag for each interpolated point. it should be a 2d ndarray. 

yeah so the flag propagation is a little tricky. one thing i don't think we want is an OR on all NN flags. this would mean that, for example, for bicubic interpolation that has an RFI flag on a single time and freq pixel, that rfi flags would spread from a 1x1 region to a 4x4 region.

i think its more reasonable, but not as aggressive, to only let the RFI flag propagate if its is a nearest neighbor for one of the interpolation points. the effect this has can be mitigated by combining this with the `medfilt_flagged` option, which will downweight the ability of the RFI flag to severely bias the interpolation.

nkern:
update doc string for `interp2d_vis` to explain flag propagation

nkern:
make medfilt_flagged True by default

nkern:
raise warning if `medfilt_flagged` if False and flags are detected in data

nkern:
allow user to propagate only NN flags (by default) or also an OR on all flags in interpolator set

hera_cal/abscal_funcs.py
```diff
@@ -874,6 +876,10 @@ def UVData2AbsCalDict(datanames, pol_select=None, pop_autos=True, return_meta=Fa
     # turn into datacontainer
     data, flags = DataContainer(d), DataContainer(f)
 
+    # turn into weights
+    if return_wgts:
+       flags = DataContainer(odict(map(lambda k: (k, (~flags[k]).astype(np.float)), flags.keys())))
```
zakiali: Maybe it would be nice to override the __invert__ operator on the DataContainer object so it's easy to go between weights and flags? This can be done at some later time if it seems worthwhile to you and others.

hera_cal/abscal_funcs.py
nkern: its slice indexing, so it will just take the extent of the array

hera_cal/abscal_funcs.py
nkern: thats an interesting idea.. seems like there is some ambiguity there, because you might want to invert a flag DC to get the inverse flags, but still want them as booleans, whereas w/ weights you want to invert and change `dtype`...

nkern: we could add a `convert_to_wgts` and a `convert_to_flgs` method, but I'm happy with this one-liner for now...

zakiali: That's a good point. Sounds good to me!

nkern:
great, thanks. going to merge this..

PR #160, opened 2018-02-03
Utils and datacontainer update

nkern:
PR for edits to `utils` and `datacontainer`, part 2 of the LST bin PR effort.


hera_cal/datacontainer.py
```diff
@@ -84,12 +92,80 @@ def __delitem__(self, key):
         else:
             raise ValueError('only supports setting (ant1, ant2, pol) keys')
 
+    def concat(self, D, axis=0):
+        """ concatenates DataContainers across an axis """
```
jsdillon: Clarify that this D is a DataContainer or a list of Datacontainers.

hera_cal/datacontainer.py
```diff
@@ -1,8 +1,16 @@
...
 
+
 class DataContainer:
-    """Object that abstracts away the pol/ant pair ordering of data dict's."""
+    """
+    Object that abstracts away the pol/ant pair ordering of data dict's.
+
+    adding two DataContainer adds their values: DC1 + DC2
+    multiplying two DataContainers multiplies their values: DC1 * DC2
+    xor two DataContainer concatenates their values along time axis: DC1 ^ DC2
```
jsdillon: Update docstring.

hera_cal/datacontainer.py
```diff
@@ -84,12 +92,80 @@ def __delitem__(self, key):
         else:
             raise ValueError('only supports setting (ant1, ant2, pol) keys')
 
+    def concat(self, D, axis=0):
```
jsdillon: To match numpy, I'd call this `concatenate`

hera_cal/datacontainer.py
```diff
@@ -84,12 +92,80 @@ def __delitem__(self, key):
...
+            if self.__contains__(k):
+                if axis == 0:
+                    newD[k] = np.concatenate([self.__getitem__(k)] + map(lambda d: d[k], D), axis=0)
+                elif axis == 1:
+                    newD[k] = np.concatenate([self.__getitem__(k)] + map(lambda d: d[k], D), axis=1)
+
+        return DataContainer(newD)
   
-  
+    def __add__(self, D):
```
jsdillon: Since you're already supporting lists for `.concat`, do you want to support them for `__add__` and `__mul__`? Not necessary, but it's your call.

hera_cal/datacontainer.py
```diff
@@ -84,12 +92,80 @@ def __delitem__(self, key):
...
+        """ adds values of two DataContainers together """
+        # check time and frequency structure matches
+        if D[D.keys()[0]].shape[0] != self.__getitem__(self.keys()[0]).shape[0]:
```
jsdillon: This shape verification code is used a lot. Perhaps you want to break it out in its own function?

hera_cal/utils.py
```diff
@@ -3,10 +3,12 @@
 import astropy.constants as const
 from astropy.time import Time
 import pyuvdata.utils as uvutils
-from pyuvdata import UVCal
+from pyuvdata import UVCal, UVData
 import os
 import hera_cal
 import copy
+import ephem
```
jsdillon: Wasn't there some reason we were looking to remove ephem as a dependency and use astropy instead? I vaguely remember hearing that. Maybe @plaplant knows?

hera_cal/utils.py
```diff
@@ -263,7 +326,233 @@ def combine_calfits(files, fname, outdir=None, overwrite=False, broadcast_flags=
...
+def get_miriad_times(filepaths, add_int_buffer=False):
+    """
+    Use aipy to get filetimes for Miriad file paths.
+    All times mark the center of an integration, which is not the standard Miriad format.
+
+    Parameters:
+    ------------
+    filepaths : type=list, list of filepaths
+
+    add_int_buffer : type=bool, if True, extend start and stop times by a single integration duration
```
jsdillon: I don't understand this parameter. What's the use case? What does the result represent in this case?

hera_cal/utils.py
```diff
@@ -263,7 +326,233 @@ def combine_calfits(files, fname, outdir=None, overwrite=False, broadcast_flags=
...
+
+    if _array is False:
+        file_starts = file_starts[0]
+        file_stops = file_stops[0]
+        int_times = int_times[0]
+
+    return file_starts, file_stops, int_times
+
+
+def data_to_miriad(fname, data, lst_array, freq_array, antpos, time_array=None, flags=None,
```
jsdillon: Does this do everything we need for omnical's vis.uvfits and for omni_apply's .uvcO and xrfi_apply's .uvcOR files? If so, perhaps you want to make an issue that we should be updating those functions to use this. Otherwise we're duplicating code and it makes it harder to keep up with changes to pyuvdata.

hera_cal/utils.py
```diff
@@ -263,7 +326,233 @@ def combine_calfits(files, fname, outdir=None, overwrite=False, broadcast_flags=
...
+              'uvw_array', 'vis_units', 'antenna_positions', 'zenith_dec', 'zenith_ra']              
+    for p in params:
+        uvd.__setattr__(p, locals()[p])
+
+    # write uvdata
+    if write_miriad:
+        # check output
+        fname = os.path.join(outdir, fname)
+        if os.path.exists(fname) and overwrite is False:
+            hera_cal.abscal.echo("{} exists, not overwriting".format(fname), verbose=verbose)
```
jsdillon: Perhaps you don't want to do it now (if not, open an issue), but echo should probably live in utils if you want to call it in utils. Ideally, other hera_cal functions would rely on utils but not vice versa. We should probably have a good reason for making dependences circular.

hera_cal/datacontainer.py
nkern: sure

hera_cal/datacontainer.py
nkern: sure

hera_cal/datacontainer.py
nkern: sure

hera_cal/datacontainer.py
nkern: that's not standard usage of operator overload

hera_cal/datacontainer.py
nkern: meh

hera_cal/utils.py
nkern: its used for the binning. in the case where an LST bin falls between file boundaries, this is needed to let the `f_select` know to load in the boundary files. not completely necessary for the user to understand what this does, its all under the hood

hera_cal/utils.py
nkern: ya

hera_cal/utils.py
nkern: I think they are doing fundamentally different things. For `omni_apply`, there is a strong case to be made that we *should* be reading-in a miriad file and using that as a UVData template which we then edit and write out. In this case, we are creating a UVData object from scratch. So I'm not sure if `data_to_miriad` can replace what's in `omni.py`, but maybe...

hera_cal/utils.py
nkern: @plaplant says it checks out: once a Julian date is converted to UTC time via `astropy`, we can use `ephem` to get the zenith RA.

hera_cal/datacontainer.py
jsdillon: Good point. I withdraw my comment.

hera_cal/utils.py
jsdillon: Ok, cool.

hera_cal/utils.py
jsdillon: Fair enough. Perhaps just add a comment to the docstring explaining the primary use case?

hera_cal/utils.py
jsdillon: I feel like our interfaces to pyuvdata are a complete mess right now. We've got lots of duplicated functionality and a lot of hard-coded code is all over the place. I'm going to make a new issue.

nkern:
ok, i think you should feel free to merge this in whenever you get the chance josh, unless @zakiali has any major comments...

nkern:
ok i'm going to merge this in then. thanks!

jsdillon:
Great! 1 down, 2 to go.

PR #162, opened 2018-02-04
adds `lstbin` module to `hera_cal`

nkern:
part 3 / 3 of the `lstbin` PR effort. this has the `utils_update` and `abscal_updates` branches rebased in, so ignore changes to those files in this PR


hera_cal/lstbin.py
```diff
@@ -0,0 +1,883 @@
...
+
+    # form new dictionaries
+    # data is a dictionary that will hold other dictionaries as values, which will
+    # themselves hold lists of ndarrays
+    data = odict()
+    flags = odict()
+    all_lst_indices = set()
+
+    # iterate over data_list
+    for i, d in enumerate(data_list):
```
jsdillon: See if this for loop can be broken out into its own function for readability.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,883 @@
...
+        lst_bins = lst_grid[sorted(all_lst_indices)]
+    else:
+        # keep all lst_grid bins and fill empty ones with unity data and mark as flagged
+        for index in range(len(lst_grid)):
+            if index in all_lst_indices:
+                # skip if index already in data
+                continue
+            for key in data.keys():
+                # fill data with blank data
+                data[key][index] = [np.ones(Nfreqs, np.complex)]
```
zakiali: Make place holder data zeros instead on ones.

jsdillon:
Move data_num into nsample_array.

hera_cal/lstbin.py
```diff
@@ -0,0 +1,883 @@
...
+
+    # shift grid by lst_start
+    if lst_start is not None:
+        lst_start = lst_grid[np.argmin(np.abs(lst_grid - lst_start))] - dlst/2
+        lst_grid += lst_start
+
+    return lst_grid
+
+
+def lst_rephase(data, bls, freqs, dlst, lat=-30.72152):
```
jsdillon: If you want to put this in the PR, but you're not 100% confident of it's functionality, at least make an issue to check that it's working properly.

hera_cal/lstbin.py
nkern: turns out its not as easy as we thought, because this for loop is connected to the rest of the function via the parameter `d`, which this FOR loop (can) overwrite if a bl in conjugated, meaning if we separated this into its own function it would be reaching through namespaces to overwrite this variable, which seems like it could easily lead to more confusion...

nkern:
ok, thanks for the review everyone. going to merge this now

Issue #164, opened 2018-02-05
Use `astropy` for RA calculation instead of `ephem`


plaplant:
Using `ephem` to perform calculations of RA yields incorrect results. We should use `astropy` to do it instead.

jsdillon:
I asked about the same thing and @nkern said you said ephem was fine for what he was doing. Are we miscommunicating here?

nkern:
naw, paul and I thought it was fine, but @bhazelton informed us after that `ephem` is apparently still slightly incorrect

nkern:
still trying to see exactly how to do the calculation using only `astropy`. its currently not clear to me how this works...

jaguirre:
where in the code is this?  I've spend a fair amount of time wrestling with astropy coordinates ...

plaplant:
@jaguirre here is where the calculation is being done right now: https://github.com/HERA-Team/pyuvdata/pull/287/files#diff-7d62f05bd3d6a58723cde3facfa3af53R297. Any insight is appreciated!

nkern:
the conversion from JD -> LST using `astropy` is understood i think. The code in `pyuvdata` and the code in `hera_cal` seem to be using the same functionalities with the `astropy.time.Time` class to do this. the question we are concerned with is to convert a Local Sidereal Time --> Equatorial right ascension (preferably in J2000 epoch) at zenith for a time integration. you can do this with `pyephem`, but we are concerned that `pyephem` is inaccurate.

jsdillon:
This is done, right @nkern ?

nkern:
ya

PR #165, opened 2018-02-06
updated LST2JD and JD2RA funcs

nkern:
the time conversion functions in `hera_cal` now only use `astropy` (instead of `astropy` + `ephem`). This goes for `LST2JD()`, `JD2LST()` and `JD2RA()`. 

The `JD2RA()` function can only calculate RA in the "current" epoch. The ability to convert an LST to an RA in J2000 epoch is still needed, but not necessary for its current role in `hera_cal`.

This addresses #164

### DIFF
--- a/hera_cal/utils.py
+++ b/hera_cal/utils.py
@@ -2,12 +2,13 @@ import numpy as np
 import aipy
 import astropy.constants as const
 from astropy.time import Time
+from astropy import units
+from astropy import coordinates as crd
 import pyuvdata.utils as uvutils
 from pyuvdata import UVCal, UVData
 import os
 import hera_cal
 import copy
-import ephem
 
 
 class AntennaArray(aipy.pol.AntennaArray):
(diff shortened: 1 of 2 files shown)


hera_cal/utils.py
```diff
@@ -157,94 +158,74 @@ def JD2LST(JD, longitude=21.42830):
...
-        obs.date = Time(JD, format='jd', scale='utc').datetime
-        ra = obs.radec_of(0, np.pi/2)[0] * 180 / np.pi
-        RA.append(ra)
+    # iterate over JD
+    LST = []
+    for jd in JD:
+        # construct astropy Time object
+        t = Time(jd, format='jd', scale='utc')
+        # get LST in radians at epoch of jd
+        LST.append(t.sidereal_time('apparent', longitude=longitude).radian)
```
plaplant: While it's true that the longitude is treated as degrees in this function, I think it's better to make it an explicit `Quantity` object, e.g., define `lon = longitude * u.deg` (assuming you `import astropy.units as u`), and pass the new `lon` in as the keyword parameter. The docs only specify that a `Quantity`, `str`, or `None` is allowed for the `longitude` keyword, so support for it as a `float` may go away in the future. See http://docs.astropy.org/en/stable/api/astropy.time.Time.html#astropy.time.Time.sidereal_time.

nkern: good call

nkern:
thanks @plaplant if you are all set w/ this i'll merge it

Issue #166, opened 2018-02-07
DataContainers should support `.values()` and `.items()`, just like regular dictionaries. 


jsdillon:

jsdillon:
Addressed by #167

PR #167, opened 2018-02-07
Added .items(), .values(), and __len__ to DataContainers

jsdillon:
This PR addresses #166 (and also implements __len__ for good measure).


jsdillon:
Thank you @nkern

PR #170, opened 2018-02-08
fixed cal_formts test w/ pyuvdata upgrade

nkern:
updated test in `test_cal_formats` to do a slightly different comparison when comparing `antenna_names` and `antenna_numbers`, which have apparently been allowed to float between lists and ndarrays in `pyuvdata` without an error. in the test, instead of comparing `_antenna_names` we compare `antenna_names` which passes when comparing a list to an ndarray so long as the elements are equal.

### DIFF
--- a/hera_cal/tests/test_cal_formats.py
+++ b/hera_cal/tests/test_cal_formats.py
@@ -51,8 +51,8 @@ class Test_HERACal(UVCal):
         uv.read_calfits(os.path.join(
             DATA_PATH, 'test_input', 'zen.2457698.40355.xx.HH.uvc.first.calfits'))
         for param in hc:
-           # print param
-           # print getattr(hc, param).value, getattr(uv, param).value
+            print param
+            print getattr(hc, param).value, getattr(uv, param).value
             if param == '_history':
                 continue
             elif param == '_git_hash_cal':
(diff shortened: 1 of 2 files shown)


hera_cal/tests/test_cal_formats.py
```diff
@@ -25,6 +25,8 @@ def test_gainHC(self):
             elif param == '_extra_keywords':
                 continue
             else:
+                if "_antenna_" in param:
+                    param = param[1:]
```
plaplant: What is the purpose of this statement? It's not clear.

hera_cal/tests/test_cal_formats.py
```diff
@@ -49,8 +51,8 @@ def test_delayHC(self):
         uv.read_calfits(os.path.join(
             DATA_PATH, 'test_input', 'zen.2457698.40355.xx.HH.uvc.first.calfits'))
         for param in hc:
-            print param
-            print getattr(hc, param).value, getattr(uv, param).value
+           # print param
```
plaplant: Instead of commenting these out, we should just get rid of them.

hera_cal/tests/test_cal_formats.py
nkern: that is because for whatever reason, a `getattr(hc, '_antenna_numbers') == `getattr(uv, '_antenna_numbers')` fails but a similar comparison of `getattr(hc, 'antenna_numbers') == `getattr(uv, 'antenna_numbers')` does not fail.. i'll add a comment

hera_cal/tests/test_cal_formats.py
nkern: ya, that was a holdout from debugging...

PR #174, opened 2018-02-10
added J2000 epoch to utils.JD2RA

nkern:
modified:   hera_cal/tests/test_utils.py
	modified:   hera_cal/utils.py

### DIFF
--- a/hera_cal/utils.py
+++ b/hera_cal/utils.py
@@ -181,7 +181,7 @@ def JD2LST(JD, longitude=21.42830):
     LST = []
     for jd in JD:
         # construct astropy Time object
-        t = Time(jd, format='jd', scale='utc')
+        t = Time(jd, format='jd')
         # get LST in radians at epoch of jd
         LST.append(t.sidereal_time('apparent', longitude=longitude*unt.deg).radian)
     LST = np.array(LST)
@@ -250,7 +250,7 @@ def LST2JD(LST, start_jd, longitude=21.42830):
         return jd_array[0]
 
 
-def JD2RA(JD, longitude=21.42830, epoch='current'):
+def JD2RA(JD, longitude=21.42830, latitude=-30.72152, epoch='current'):
     """
     Convert from Julian date to Equatorial Right Ascension at zenith
     during a specified epoch.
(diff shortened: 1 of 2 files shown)


hera_cal/utils.py
```diff
@@ -181,7 +181,7 @@ def JD2LST(JD, longitude=21.42830):
     LST = []
     for jd in JD:
         # construct astropy Time object
-        t = Time(jd, format='jd', scale='utc')
+        t = Time(jd, format='jd')
```
plaplant: What is the motivation for removing `scale='utc'`? Most examples I've seen online specify this, but admittedly I'm not familiar enough with the other options to know if this is correct or not.

hera_cal/utils.py
```diff
@@ -259,17 +259,18 @@ def JD2RA(JD, longitude=21.42830, epoch='current'):
...
+    latitude : type=float, latitude of observer in degrees north, default=HERA latitutde
+               This only matters when using epoch="J2000"
     
-    epoch : type=str, epoch for RA calculation. options=['current'].
+    epoch : type=str, epoch for RA calculation. options=['current', 'J2000'].
             The 'current' epoch is the epoch at JD. Note that
-            LST is *defined* as the zenith RA in the current epoch. 
-
-    ## TODO : add a "J2000" option to return RA in J2000 epoch.
+            LST is defined as the zenith RA in the current epoch.
```
plaplant: Though it's pedantic, I think that it's worth mentioning that it's not _technically_ J2000, but rather ICRS, that we're phasing up to. The difference of right ascension is evidently only ~78 mas (per wikipedia: https://en.wikipedia.org/wiki/International_Celestial_Reference_System), but worth noting. I _do_ think phasing to ICRS is the better choice, since it's an externally-defined system, so just adding a comment here is sufficient.

hera_cal/utils.py
nkern: So 1) the default is `utc` I believe, however that doesn't mean we shouldn't have it there, but 2) I was just going to follow the `pyuvdata` convention which calls it without a `scale` parameter. We can leave it in if you want. I'm happy either way

hera_cal/utils.py
nkern: ok good point

nkern:
@plaplant thanks for the review. let me know if there is anything else you think needs addressed! if not i'll merge this in

Issue #157, opened 2018-02-02
Add writer functionality and unit tests to delay_filter module


jsdillon:
See add_dspec branch.

jsdillon:
Addressed with #181

PR #181, opened 2018-02-14
Add delay_filter module

jsdillon:
This gets us up and running with a basic interface to `uvtools.dspec.delay_filter` that should enable simple interactive delay filtering of data and thin scripts for delay filtering files as part of the pipeline.

Also addresses #157.


jsdillon:
Also, @zakiali if you have any idea how to eliminate my doubled commits, let me know.

jsdillon:
Nevermind, I figured out how to squash commits with a rebase.

hera_cal/delay_filter.py
```diff
@@ -0,0 +1,113 @@
...
+                or list of UVData instances to concatenate into a single internal DataContainer
+            format: file format of data. Default 'miriad.' Ignored if input_data is UVData object(s).
+        '''
+        if isinstance(input_data, (str,UVData)):
+            self.writable = True
+            self.input_data, self.format = input_data, format
+        self.data, self.flags, self.antpos, _, self.freqs, self.times, _, _ = io.load_vis(input_data, return_meta=True, format=format)
+
+
+    def load_dicts(self, data, flags, freqs, antpos):
```
zakiali: Was there a reason you split the load_data and load_dicts function? Can we combine them?

hera_cal/delay_filter.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __init__(self):
+        '''Class for loading data, performing uvtools.dspec.delay_filter, and writing out data using pyuvdata.
+        To use, run either self.load_data() or self.load_dicts(), then self.run_filter(). If data is loaded with a single 
+        string path or a single UVData object, it can be written to a new file using self.write_filtered_data().
+        '''
+        self.writable = False
+
+
+    def load_data(self, input_data, format='miriad'):
```
zakiali: Can we change "format" to "file_type". format is a python builtin and we use file_type in other places.

hera_cal/delay_filter.py
zakiali: rename function to load_data_dicts. Makes it a bit clear.

hera_cal/delay_filter.py
jsdillon: I'm using filetype (no underscore, which matches abscal and firstcal).

hera_cal/delay_filter.py
jsdillon: I'm renaming load_dicts to load_data_as_dicts to be more clear.

Issue #183, opened 2018-02-16
Fix Abscal Gain Conventions


jsdillon:
We should be consistent in our gain conventions. The default should be 

V_observered = gi gj* V_true.

This is what omnical and firstcal use, but not abscal right now.

jsdillon:
Addressed in #185

PR #185, opened 2018-02-18
changed gain convention of abscal to divide

nkern:
updated gain convention of abscal to V_data = g_i * conj(g_j) * V_model

addresses #183


hera_cal/abscal.py
```diff
@@ -1044,7 +1044,7 @@ def abscal_run(data_files, model_files, pol_select=None, verbose=True, overwrite
             if all_antenna_gains:
                 raise ValueError("can't run gen_phs_cal when all_antenna_gains is True")
             AC.phs_logcal(verbose=verbose)
-            AC.data = apply_gains(AC.data, AC.ant_phi_gain)
+            AC.data = apply_gains(AC.data, AC.ant_phi_gain, gain_convention='divide')
```
jsdillon: Isn't this already the default? Why specify it here and not elsewhere in run_abscal?

hera_cal/abscal_funcs.py
```diff
@@ -627,12 +627,15 @@ def apply_gains(data, gains, gain_convention='multiply'):
             values are complex ndarrays
             with shape matching data's visibility ndarrays
 
-        Optionally, can be a tuple holding multiple gains dictionaries
-        that will all be multiplied together.
+            Optionally, can be a tuple holding multiple gains dictionaries
+            that will all be multiplied together.
+
+    gain_convention : type=str, options=['multiply', 'divide']
+                      option to multiply in or divide in gain solutions to data.
```
jsdillon: I'd be more explicit here:
```
'divide' means that the gains need to be divided out of the observation to get the model/truth, i.e. that V_obs = gi gj* V_true.
'multiply' means that the gains need to be multiplied into the observation to get the model/truth, i.e. that V_true = gi gj& V_obs.
In Abscal (as on omnical and redcal), the standard gain convention is 'divide'.
```

hera_cal/abscal.py
nkern: yup, i added it to all `apply_gains` calls

hera_cal/abscal_funcs.py
nkern: sure

Issue #178, opened 2018-02-13
Standardize writing to calfits files created from scratch


jsdillon:
It should support input from dictionaries with keys of the form (1,'x'). This can be built off hera_cal.cal_formats.HERACal.

This is a followup from #163 that didn't make it into #175.

PR #190, opened 2018-02-27
adds write_cal to hera_cal.io

nkern:
`hera_cal.io.write_cal` is a method to format gain dictionaries into `pyuvdata.UVCal` structures, and to then write them out as `calfits` files.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -100,9 +100,14 @@ def load_vis(input_data, return_meta=False, filetype='miriad', pop_autos=False, 
         return data, flags
 
 
-def write_vis(outfilename, data, flags, filetype='miriad', history='', clobber=False, **kwargs):
-    '''TODO: migrate in hera_cal.utils.data_to_miriad and generalize to also write uvfits.'''
-    raise NotImplementedError('This function has not been implemented yet.')
+def write_vis(fname, data, flags, filetype='miriad', history=' ', overwrite=False, **kwargs):
+    """
+
+    """
+    
+
+    
+
 
 
 def update_uvdata(uvd, data=None, flags=None, add_to_history='', **kwargs):
@@ -368,7 +373,7 @@ def write_cal(fname, gains, freqs, times, pols, flags=None, quality=None, write_
         return uvc
 
 
-def update_cal(infilename, outfilename, gains=None, flags=None, quals=None, add_to_history='', overwrite=False, **kwargs):
+def update_uvcal(infilename, outfilename, gains=None, flags=None, quals=None, add_to_history='', overwrite=False, **kwargs):
     '''Loads an existing calfits file with pyuvdata, modifies some subset of of its parameters, 
     and then writes a new calfits file to disk. Cannot modify the shape of gain arrays. 
     More than one spectral window is not supported.
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -245,9 +245,124 @@ def load_cal(input_cal, return_meta=False):
         return gains, flags
 
 
-def write_cal():
-    '''TODO: copy over code from hera_cal.cal_formats.HERACal'''
-    raise NotImplementedError('This function has not been implemented yet.')
+def write_cal(fname, gains, freqs, times, pols, flags=None, quality=None, write_file=True,
+              return_uvc=True, outdir='./', overwrite=False, gain_convention='divide', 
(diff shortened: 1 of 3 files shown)


hera_cal/io.py
```diff
@@ -243,15 +247,131 @@ def load_cal(input_cal, return_meta=False):
...
+    '''Format gain solution dictionary into pyuvdata.UVCal and write to file
+
+    Arguments:
+        fname : type=str, output file basename
+        gains : type=dictionary, holds complex gain solutions. keys are antenna + pol
+                tuple pairs, e.g. (2, 'x'), and keys are 2D complex ndarrays with time
+                along [0] axis and freq along [1] axis.
+        freqs : type=ndarray, holds unique frequencies channels in Hz
+        times : type=ndarray, holds unique times of integration centers in Julian Date
+        pols : type=ndarray, holds unique polarization of gains in 'x' and/or 'y' form
```
zakiali: "...gains in string form. Can be either 'x' or 'y'".

hera_cal/io.py
```diff
@@ -243,15 +247,131 @@ def load_cal(input_cal, return_meta=False):
...
+    freq_array = freq_array[None, :]
+    spw_array = np.arange(Nspws)
+    channel_width = np.median(np.diff(freq_array))
+
+    # form gain, flags and qualities
+    gain_array = np.empty((Nants_data, Nspws, Nfreqs, Ntimes, Njones), np.complex)
+    flag_array = np.empty((Nants_data, Nspws, Nfreqs, Ntimes, Njones), np.bool)
+    quality_array = np.empty((Nants_data, Nspws, Nfreqs, Ntimes, Njones), np.float)
+    for i, p in enumerate(pol_array):
+        for j, a in enumerate(antenna_numbers):
```
zakiali: This should be a loop over ant_array. From pyuvdata docs, ant array is the index of the antenna in data arrays.

nkern: but its an enumerate, so `j` is the index, which is being used to index into `gain_array`. this must be `antenna_numbers` because `a` is used to index into the `gains` dictionary. right?

hera_cal/tests/test_io.py
```diff
@@ -226,12 +226,39 @@ def test_load_cal(self):
...
+            for j, a in enumerate(ants):
+                gains[(a, p)] = np.ones((Ntimes, Nfreqs), np.complex)
+                quality[(a, p)] = np.ones((Ntimes, Nfreqs), np.float) * 2
+                flags[(a, p)] = np.zeros((Ntimes, Nfreqs), np.bool)
+
+        # test basic execution
+        uvc = io.write_cal("ex.calfits", gains, freqs, times, pols, flags=flags, quality=quality,
+                           overwrite=True, return_uvc=True, write_file=True)
+        self.assertTrue(os.path.exists("ex.calfits"))
+        self.assertAlmostEqual(uvc.gain_array[0,0,0,0,0], (1+0j))
```
zakiali: Do you want to loop over all antennas and pols and do these assertions?
I understand it's redundant, but there may be some weird edge cases not found by this test.

hera_cal/io.py
zakiali: In this case it can be either antenna_numbers or ant_array, but from the docs, ant_array is the parameter you want to use to do this, not antenna_numbers. This is more for consistency than anything. My understanding is that antenna_numbers can be in any order and is just a way to label/name the antennas where ant_array is the array you look at for indexing into the data arrays. It's the same loop you have just replacing antenna_numbers with ant_array.

zakiali: oh wait, just realized that ant_array is just an array from 0, nants-1. Why?

zakiali: Oh wait, I misunderstood the docs. I would be happy if it was something like for j, a in zip(ant_array, antenna_numbers)

nkern: but I think that is exactly the same as what I have here: `enumerate(antenna_numbers)`

hera_cal/io.py
nkern: got rid of pol argument actually, can get it from the gains dictionary

hera_cal/tests/test_io.py
nkern: meh, not a fan of the idea of looping through the whole array... instead of looping over I just checked the sum of the entire array is what I would expect

nkern:
thanks for the comments!

nkern:
addresses #178

Issue #177, opened 2018-02-13
Standardize writing to miriad or uvfits files created from scratch


jsdillon:
It should support input from DataContainers. This should generalize hera_cal.utils.data_to_miriad

This is a followup from #163 that didn't make it into #175.

PR #195, opened 2018-03-01
adds write_vis to io module

nkern:


nkern:
adresses #177.

nkern:
a lot of this is code moved over from `abscal.data_to_miriad`, so it shouldn't require too detailed of a review

Issue #180, opened 2018-02-13
Deprecate hera_cal.firstcal.UVData_to_dict in favor of hera_cal.io.load_vis


jsdillon:
Use `nested_dict = True` to get the legacy behavior of UVData_to_dict

Issue #179, opened 2018-02-13
Deprecate hera_cal.abscal_funcs.UVData2AbsCalDict in favor of hera_cal.io.load_vis


jsdillon:
Keep in mind that the default parameters are slightly different.

PR #196, opened 2018-03-01
Issue180/deprecate uv data to dict

zakiali:
This PR changes the instances of the UVData_2_dict function in firstcal to use the io.load_vis function.

Also deprecates `abscal.UVData2AbscalDict` for `io.load_vis`

### DIFF
--- a/hera_cal/abscal_funcs.py
+++ b/hera_cal/abscal_funcs.py
@@ -802,100 +802,6 @@ def array_axis_to_data_key(data, array_index, array_keys, key_index=-1, copy_dic
         return new_data
 
 
-def UVData2AbsCalDict(datanames, pol_select=None, pop_autos=True, return_meta=False, filetype='miriad',
-                      pick_data_ants=True, return_wgts=False):
-    """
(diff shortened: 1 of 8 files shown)


nkern:
addresses #179 and #180

hera_cal/abscal_funcs.py
```diff
@@ -1563,7 +1462,7 @@ def avg_file_across_red_bls(data_fname, outdir=None, output_fname=None,
         uvd.read_miriad(data_fname)
 
     # get data
-    data, flags = UVData2AbsCalDict(uvd)
+    data, flags = io.load_vis(uvd, pop_autos=True, return_meta=False, pick_data_ants=True)
```
jsdillon: You don't need return_meta=False.

nkern: yeah, just in case defaults change doesn't hurt to have

nkern:
ah, you merged it by accident... let me go back and undo that...

nkern:
ok should be good to go now

nkern:
*once it passes travis

PR #201, opened 2018-03-05
Apply cal module

jsdillon:
Here's a simple alternative for applying calibration solutions for IDR2.1

Issue #197, opened 2018-03-02
Build update_cal module


jsdillon:
Based on the current architecture of the analysis pipeline, we need a module that takes in:

* (Un)Calibrated data
* Calibration solutions to apply
* Optionally, calibration solutions to remove.

This will then do the following:
1) Remove the old calibration solutions everywhere where the data is not flagged.
2) Apply the new calibration solutions everywhere where the data is not flagged.
3) Update the data flags so that any calibration flags are propagated to all visibilities involved.
4) Append to the history of the data file and write it out.

I've already got most of this infrastructure, but could probably use some help.

jsdillon:
Addressed by #206.

PR #206, opened 2018-03-07
Add apply cal module

jsdillon:
Here's a simple alternative for applying calibration solutions for IDR2.1. I've also added support for loading npz flags.

I don't think it makes sense to add that functionality to hera_cal.io, since the data available in the .npz files makes it impossible to create a DataContainer without the accompanying miriad file. We can think about whether we want to add support for npz flags in io.load_vis (in which case you also have the .uv) or whether we want to eventually deprecate npz flags or improve them.

I screwed up the history on #201, so I'm starting over here. 

This addresses #197.


hera_cal/apply_cal.py
```diff
@@ -0,0 +1,99 @@
...
+            if new_gains.has_key((i,pol[0])) and new_gains.has_key((j,pol[1])) and (old_gains == None
+                or (old_gains.has_key((i,pol[0])) and old_gains.has_key((j,pol[1])))):
+                    gigj_new = new_gains[(i,pol[0])] * np.conj(new_gains[(j,pol[1])])
+                    if old_gains is not None:
+                        gigj_old = old_gains[(i,pol[0])] * np.conj(old_gains[(j,pol[1])])
+                    else:
+                        gigj_old = np.ones_like(gigj_new)
+                    # update all the data, assuming it wasn't flagged to begin with
+                    if gain_convention == 'divide':
+                        data[(i,j,pol)][~data_flags[(i,j,pol)]] *= (gigj_old / gigj_new)[~data_flags[(i,j,pol)]]
```
zakiali: This line should be "/= " instead of "*="?

hera_cal/apply_cal.py
```diff
@@ -0,0 +1,99 @@
...
+                    elif gain_convention == 'multiply':
+                        data[(i,j,pol)][~data_flags[(i,j,pol)]] *= (gigj_new / gigj_old)[~data_flags[(i,j,pol)]]
+                    else:
+                        raise KeyError("gain_convention must be either 'divide' or 'multiply'.")
+                    # update data flags
+                    data_flags[(i,j,pol)][cal_flags[(i, pol[0])]] = True
+                    data_flags[(i,j,pol)][cal_flags[(j, pol[1])]] = True
+            else:
+                data_flags[(i,j,pol)] = np.ones_like(data_flags[(i,j,pol)]) #all flagged
```
zakiali: isn't this redundant? I think the first if statement takes care if this case.

hera_cal/apply_cal.py
zakiali: This does assume that the conventions of both the old and new gains are the same. Do we want to support them not being the same?

jsdillon:
Yeah, that option seems pretty dangerous to me...to allow for the production of files with some antennas calibrated and some not (but also not flagged) seems like it would create problems.

Issue #184, opened 2018-02-17
ensure accuracy of `lst_rephase` function in `lstbin`


nkern:
There is a function called `lst_rephase` in `hera_cal.lstbin` that attempts to rephase a time integration to that of a different time. It follows the `pyuvdata` methods for rephasing, but preliminary tests seem to suggest there could be something wrong at the 10% level. It isn't a core part of `lstbin` yet, but we hope at some point to make its use the default action in `lstbin`, and therefore need a way to test it against other rephasing algorithms.

PR #207, opened 2018-03-07
adds (verifies) `lst_rephase` function for phasing data to center of LST bin

nkern:
This verifies the `hera_cal.lstbin.lst_rephase` function, that allows one to rephase data to the center of an LST bin. This verification comes from a PAPER point source simulation, and from experiments on real data which shows reduction in STD of LST bins when using `lst_rephase`.

In summary, the rephasing algorithm uses `aipy.coords.eq2top_m` and `aipy.coords.top2eq_m` functions and Eqn 21 & 22 in Zhang, Y. et al. 2018: "Unlocking Sensitivity..."


nkern:
I don't think this PR is critical for IDR2.1, so we can wait to get this in if priority is on other areas, but at some point we will want this capability

jsdillon:
If Zaki has time to review it, I think it can go in. It's self-contained to LST-binning.

nkern:
addresses #184

hera_cal/lstbin.py
```diff
@@ -143,6 +158,32 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, lst_start=None, lst
...
+
+                # copy rephased data if desired
+                if copy_rephase:
+                    d = copy.deepcopy(d)
+
+                # form baseline dictionary and lst_rephase
+                bls = odict(map(lambda k: (k, antpos[k[0]] - antpos[k[1]]), d.keys()))
+                lst_shift = lst_grid[grid_indices] - l
+                lst_rephase(d, bls, freq_array, lst_shift, lat=lat)
+                d.phase_type = "LSTBIN"
```
zakiali: d is a DataContainer right? Do we want to explicitly have a phase attribute on data containers? I'm all for it considering we're going to have multiple types of phased data products now.

nkern: well, it gets a little tricky b/c if we have a `phase` attribute we also need a `phase_center` and `phase_epoch` as well, and the idea behind the `lst_rephase` function is that this can be different *for every single integration*, so this turns out to be a decent amount of data you need to carry around. This attribute is used only internally in the context of this single function, so I don't think its something we need to make part of the `DataContainer` object.

jsdillon:
Went over this with Nick. Looks good to me.

PR #208, opened 2018-03-08
Added smooth cal arg parser and drive script

jsdillon:


scripts/smooth_cal_run.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python2.7
+"""Command-line drive script for hera_cal.smooth_cal"""
```
plaplant: It would be nice to add some more documentation here about what is happening. As @nkern pointed out, there are a few steps going on, and it would be nice to give an overview of the steps going on.

scripts/smooth_cal_run.py
```diff
@@ -0,0 +1,16 @@
...
+from hera_cal.smooth_cal import Calibration_Smoother, smooth_cal_argparser
+import sys
+
+a = smooth_cal_argparser()
+
+# Run Calibration smoothing
+sc = Calibration_Smoother(binary_wgts=a.binary_wgts)
+sc.load_cal(a.cal_infile, prev_cal=a.prev_cal, next_cal=a.next_cal)
+sc.load_data(a.data, prev_data=a.prev_data, next_data=a.next_data)
+if not a.disable_time:
```
plaplant: Does smoothing in time and frequency commute? I know that convolution in general is a linear operator, but am concerned that things like adding weights/mirrors and doing a CLEAN might not be linear. It's fine if the order matters here, maybe just add to the documentation that we're smoothing in time first, then smoothing in frequency.

jsdillon: Yes, order does matter. Using the module interactively, it is possible to reverse the order. I don't see a need to implement that at the script level now. We can revisit if we have a reason to.

I've added a bunch of explanatory material to the script (the notebook also explains a lot and I expect users to go there first to get an understanding of how the script works). 

I'm going to merge unless anyone objects.

PR #209, opened 2018-03-08
Added arg parsing and a command line script for delay filter

jsdillon:
Also changed the default standoff to 15 ns following Zaki's recommendation.


adampbeardsley:
Generally looks good. I just have a couple requests to make things a little more robust and in-line with how we do this in QM.

1. Could you put the arg parser function in utils? This is pretty superficial, but would be nice to unify the interface with QM when possible.

2. It would also be nice to (at least have the option to) return the parser itself. I've been working on basically this same function in my xrfi_after_delay branch, and I need to parse multiple sets of arguments. By returning the parser, I can make it `parse_known_args`.

Alternatively I can just keep developing my version of the argparser, but I don't think we want multiple versions in the end.

jsdillon:
I can certainly do 2. I kinda don't like moving the arg parser outside the module. I find that very confusing, to be honest.

I'm also perfectly fine with multiple arg parsers, to be honest. Even though XRFI w/ Delay uses this module, it's fundamentally a different block in the analysis pipeline.

adampbeardsley:
Ok, so originally I was planning to combine the xrfi parser and the delay parser. But maybe it makes more sense to just have a single parser, independent of this one, to handle both. In that case (2) isn't needed, unless you want it.

PR #211, opened 2018-03-09
fixes io.write_cal give UVCal.ant_array definition

nkern:
realized b/c we didn't pull in my `apply_cal` module this was never fixed.
this fixes `write_cal` given the new definition of `UVCal.ant_array` object.

should be a quick review


jsdillon:
Looks fine.

PR #212, opened 2018-03-09
makeflow compatability for abscal_run()

nkern:
this PR adds an LST checking capability within `abscal_run()`, which is needed for how we will use it for IDR2 analysis runs w/ makeflow.


jsdillon:
Add unit test for match_times.

jsdillon:
Otherwise, looks good to me (assuming it passes tests).

PR #213, opened 2018-03-12
Io write cal check zero

nkern:
if gain_array has zeros in them, then they can incur nans when applied to data. This checks to see if zeros exist `gain_array`, and if so sets the values to 1.0 and makes sure those pixels are flagged

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -514,6 +514,9 @@ def write_cal(fname, gains, freqs, times, flags=None, quality=None, write_file=T
                 flag_array[j, :, :, :, i] = np.ones((Nspws, Nfreqs, Ntimes), np.bool)
                 quality_array[j, :, :, :, i] = np.ones((Nspws, Nfreqs, Ntimes), np.float)
 
+    # Check gain_array for values close to zero, if so, set to 1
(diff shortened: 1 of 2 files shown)


hera_cal/io.py
```diff
@@ -514,6 +514,11 @@ def write_cal(fname, gains, freqs, times, flags=None, quality=None, write_file=T
                 flag_array[j, :, :, :, i] = np.ones((Nspws, Nfreqs, Ntimes), np.bool)
                 quality_array[j, :, :, :, i] = np.ones((Nspws, Nfreqs, Ntimes), np.float)
 
+    # Check gain_array for values close to zero, if so, set to 1
+    zero_check = np.isclose(gain_array, 0)
```
jsdillon: I'd set the atol and rtol to be a bit more stringent. Like rtol = 1e-8 and atol = 1e-10.

nkern:
@zakiali that's a good point, and may be a better path to take. I figured that if the gains are zero then we can effectively treat them as flagged, in which case it doesn't matter what their value is, so I figured might as well treat them as 1.0. the other thing we can do is to not change their value but just flag them. lets chat at our meeting today about it

PR #220, opened 2018-03-13
enabled abscal_run to write blank and flagged calfits file if no model overlap

nkern:
this capability is for the makeflow pipeline which will break if there is no output calfits file.


nkern:
is the history writing for this calfits file critical to our IDR2 history-writing chain? if so i'm inclined to fix it in this PR. if not I'll make an issue to address later

jsdillon:
I’d do something very simple now... just capture the run script command.
We’ll revisit git hashes and all that later.

jsdillon:
Thanks!

PR #222, opened 2018-03-14
Xrfi after delay filter

adampbeardsley:
This was supposed to be synced with the respective hera_qm PR, but I forgot. Other than my editor auto-fixing whitespace, the biggest thing is that I made the outfile optional in the delay filter argparser.


hera_cal/delay_filter.py
```diff
@@ -111,15 +111,15 @@ def write_filtered_data(self, outfilename, filetype_out='miriad', add_to_history
...
+            io.update_vis(self.input_data, outfilename, filetype_in=self.filetype, filetype_out=filetype_out, data=data_out,
                           add_to_history=add_to_history, clobber=clobber, **kwargs)
 
 
 def delay_filter_argparser():
     '''Arg parser for commandline operation of hera_cal.delay_filter.'''
     a = argparse.ArgumentParser(description="Perform delay filter of visibility data.")
     a.add_argument("infile", type=str, help="path to visibility data file to delay filter")
-    a.add_argument("outfile", type=str, help="path to new visibility results file")
+    a.add_argument("outfile", nargs='?', default=None, type=str, help="path to new visibility results file")
```
plaplant: I think we should document (in this help string or elsewhere) that the script default is to append 'D' to the input filename if outfile is not specified.

PR #223, opened 2018-03-16
fixed bug in abscal.match_times

nkern:
@plaplant  noticed that if the model files start at a large LST and wrap around 2pi they get temporarily unwrapped (ex. 5 -> 2 radians turns into 5 -> 8.28 radians). if a data file starts at a small LST (ex. 1 radian) it wasn't also getting unwrapped to match the unwrapping of the model files. `abscal_funcs.match_times` now appropriately handles this case.


jsdillon:
Looks good to me. If you can think of a unit test to expose the issue...or even an assert statement so abscal will fail less quietly in this way, that would be good.

jsdillon:
I'll merge this in and make an issue for @nkern

PR #225, opened 2018-03-16
Make repo pep8 compliant

plaplant:
This PR makes all code in the repo pep8 compliant. This was done using the `autopep8` tool as follows, invoked from the top level directory of the repo:
```
autopep8 . --recursive --in-place --pep8-passes 2000 --verbose -a
```
Lingering issues (such as reformatting `if val == True:` to `if val:`) were addressed by hand. This also adds a check to the travis build, which will now error if non-compliant changes are added in the future.


jsdillon:
I’m generally fine with pep8, but I personally find the short line length
to often make things far less readable. Can we do a longer maximum line
length? Maybe 140 characters?

plaplant:
@jsdillon I agree, and I have explicitly ignored the max line length error when checking on travis (E501 in pep8 parlance). When I ran the autopep8 command, I also let the max line length be 1000 characters (though I didn't look closely to see that no lines were broken unnecessarily). I believe that this allows us to set our own line lengths, but if you find that not to be the case, we can adjust.

plaplant:
The code base has moved a lot since this was first opened, so I'm going to close this and apply the `autopep8` tool to the latest version.

PR #226, opened 2018-03-19
enabled propagation of calfits gain and meta in abscal_run

nkern:


hera_cal/abscal.py
```diff
@@ -809,7 +809,8 @@ def abscal_arg_parser():
     a = argparse.ArgumentParser(description="command-line drive script for hera_cal.abscal module")
     a.add_argument("--data_files", type=str, nargs='*', help="list of file paths of data to-be-calibrated.", required=True)
     a.add_argument("--model_files", type=str, nargs='*', help="list of data-overlapping miriad files for visibility model.", required=True)
-    a.add_argument("--calfits_fname", type=str, default=None, help="name of output calfits file.")
+    a.add_argument("--calfits_files", type=str, nargs='*', help="list of calfits files to multiply with abscal solution before writing to disk.")
```
jsdillon: I'd probably call this "--old_calfits_files" or "--calfits_infiles" or similar to clarify what this argument is.

(And then propagate the change below too).

hera_cal/abscal.py
```diff
@@ -881,13 +883,17 @@ def abscal_run(data_files, model_files, verbose=True, overwrite=False, write_cal
                 a list of paths to miriad file(s) containing complex
                 visibility data, or a path itself
 
+    calfits_files : type=list or string, path(s) to calfits files(s)
```
jsdillon: When I first read this, I thought you were supporting multiple calfits files that get multiplied together to make a single calfits file, but actually you're only supporting a single calfits file to multiply in _per output file_. That's totally fine, but it could use clarification in the documentation.

Relatedly, you probably want to say here that if a calfits file is provided, quals, total quals, and history get passed through.

hera_cal/io.py
```diff
@@ -380,7 +380,7 @@ def load_cal(input_cal, return_meta=False):
 
     Returns:
         if return_meta is True:
-            (gains, flags, quals, total_qual, ants, freqs, times, pols)
+            (gains, flags, quals, total_qual, ants, freqs, times, pols, history)
```
jsdillon: Is now the time to switch to returning gains, flags, and a meta dictionary? That would be a more extensible format. Alternatively, we can hold off for now and make it an issue so we can get moving on the pipeline. Thoughts?

hera_cal/abscal.py
```diff
@@ -982,113 +988,129 @@ def abscal_run(data_files, model_files, verbose=True, overwrite=False, write_cal
...
-            gain_list.append(AC.ant_phi_gain)
-
-        # collate gains
-        if len(gain_list) == 0:
-            raise ValueError("abscal_run executed without any calibration arguments set to True")
-        gain_dict = merge_gains(gain_list)
+            flag_dict = odict(map(lambda k: (k, np.ones((Ntimes, Nfreqs), np.bool)), gain_dict.keys()))
+
+        # load calfits file if provided
+        if calfits_files is not None:
```
jsdillon: I think a lot of this could be much more easily accomplished with `hera_cal.io.update_cal`.

hera_cal/io.py
```diff
@@ -413,26 +414,28 @@ def load_cal(input_cal, return_meta=False):
         raise TypeError('Input must be a UVCal object, a string, or a list of either.')
 
     #load gains, flags, and quals into dictionaries
-    gains, quals, flags = odict(), odict(), odict()
-    for i, ant in enumerate(cal.ant_array):
-        for ip, pol in enumerate(cal.jones_array):
+    gains, quals, flags, total_qual = odict(), odict(), odict(), odict()
+    for ip, pol in enumerate(cal.jones_array):
+        if cal.total_quality_array is not None:
+            total_qual[jonesnum2str[pol]] = cal.total_quality_array[0, :, :, ip].T
```
jsdillon: Good idea!

hera_cal/io.py
```diff
@@ -500,7 +505,10 @@ def write_cal(fname, gains, freqs, times, flags=None, quality=None, write_file=T
     gain_array = np.empty((Nants_data, Nspws, Nfreqs, Ntimes, Njones), np.complex)
     flag_array = np.empty((Nants_data, Nspws, Nfreqs, Ntimes, Njones), np.bool)
     quality_array = np.empty((Nants_data, Nspws, Nfreqs, Ntimes, Njones), np.float)
+    total_quality_array = np.empty((Nspws, Nfreqs, Ntimes, Njones), np.float)
     for i, p in enumerate(pol_array):
+        if total_qual is not None:
+            total_quality_array[:, :, :, 0] = total_qual[p].T[None, :, :]
```
jsdillon: I'm confused about this assignment. Shouldn't the dimension that's always index 0 be the spectral window dimension? Shouldn't it be `total_quality_array[0, :, :, i]`?

Also, I think it's preferable to use `np.newaxis` rather than None (it's just clearer what's going on).

jsdillon:
The more I think about it, we can make this whole PR much simpler if we use update_cal. You wouldn't have to update hera_cal.io (maybe change it to turn total_quality_array into a dictionary, but I'm not sure if that's even necessary). It would remove the need to change smooth_cal and and vastly simplify the changes to abscal.

hera_cal/io.py
nkern: yeah let's just make that its own PR seeing as its somewhat of a separate issue

hera_cal/io.py
nkern: yeah you're right, thanks for the catch

hera_cal/abscal.py
nkern: yeah i guess i was trying to limit I/O, so the idea was to this in memory before having to write out the calfits file...

jsdillon: Calfits I/O is trivial and will never be an issue compared to visibility I/O.

PR #230, opened 2018-03-22
changed default window for delay_filter to blackman-harris

jsdillon:


zakiali:
I approved, but did the tests pass locally for you? I would imagine answers changing if there are tests that are comparing values after delay filtering.

jsdillon:
Yes they did, but I'll wait for them to finish before I merge

Issue #232, opened 2018-03-23
Improve smooth_cal to take out a per-antenna delay, smooth, and then put that delay back in


jsdillon:
This is creating an issue where the antennas with long delays wrap faster than the smoothing scale (i.e. 10 MHz)... making them "unsmoothable."

jsdillon:
Addressed in #233

PR #233, opened 2018-03-24
Implemented delay rephasing in smooth_cal freq filter

jsdillon:
This addresses #232 and fixes the problem of the two antennas that looked bad in the .OCRS files but fine in the .OCR files that @nkern found today.

The root of the problem that @AaronParsons and I figured out today was that the smoother was trying to filter the real and imaginary components of the calibration solutions, which can have a lot of spectral structure when the antenna has a large delay (beyond the 10 MHz scale) but isn't what we actually want to smooth out. My solution is to find a per-antenna, per-integration delay using abscal.fft_dly, remove it, smooth, and then put the delay back in.

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -8,6 +8,7 @@ from copy import deepcopy
 import warnings
 import uvtools
 import argparse
+from hera_cal.abscal import fft_dly
 
 def drop_cross_vis(data):
     '''Delete all entries from a DataContiner that not autocorrelations in order to save memory.'''
@@ -79,6 +80,7 @@ def build_weights(unnorm_chisq_per_ant, autocorr, flags, binary_wgts = False):
 
 def freq_filter(gains, wgts, freqs, filter_scale = 10.0, tol=1e-09, window='none', skip_wgt=0.1, maxiter=100):
     '''Frequency-filter calibration solutions on a given scale in MHz using uvtools.dspec.high_pass_fourier_filter.
+    Befor filtering, removes a per-integration delay using abscal.fft_dly, then puts it back in after filtering.
     
     Arguments:
         gains: ndarray of shape=(Ntimes,Nfreqs) of complex calibration solutions to filter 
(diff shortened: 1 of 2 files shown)


hera_cal/tests/test_smooth_cal.py
```diff
@@ -78,10 +78,19 @@ def test_freq_filter(self):
         gains[3,5] = 10.0
         wgts = np.ones((10,10),dtype=float)
         wgts[3,5] = 0
-        freqs = np.linspace(100.,200.,10)
+        freqs = np.linspace(100.,200.,10)*1e6
```
nkern: needs an `endpoint=False` in the `linspace` call

hera_cal/tests/test_smooth_cal.py
```diff
@@ -78,10 +78,19 @@ def test_freq_filter(self):
...
         ff = sc.freq_filter(gains, wgts, freqs)
         np.testing.assert_array_almost_equal(ff, np.ones((10,10),dtype=complex))
 
+        #test rephasing
```
nkern: does this test that without the delay removal the gain smoothing fails? otherwise this isn't really testing that this method fixes the aforementioned problem...

hera_cal/smooth_cal.py
```diff
@@ -99,9 +101,11 @@ def freq_filter(gains, wgts, freqs, filter_scale = 10.0, tol=1e-09, window='none
     '''
     sdf = np.median(np.diff(freqs)) / 1e9 #in GHz
     filter_size = (filter_scale / 1e3)**-1 #Puts it in ns
-    filtered, res, info = uvtools.dspec.high_pass_fourier_filter(gains, wgts, filter_size, sdf, tol=tol, window=window, 
+    (dlys, phi) = fft_dly(gains, wgts, df=sdf*1e9, medfilt=False) #delays are in seconds
+    rephasor = np.exp(-2.0j * np.pi * np.outer(dlys, freqs))
```
zakiali: your exponent here and in line 89 of the test differ by a negative sign. Are you using the right convention?

jsdillon:
BTW, here's a demonstration that the fix worked:

screen shot 2018-03-24 at 9 56 31 am

hera_cal/tests/test_smooth_cal.py
jsdillon: I have verified that it does fail without the delay removal.

hera_cal/smooth_cal.py
jsdillon: In the test, I'm inserting a delay. It can be positive or negative, depending on the convention. Regardless, this undoes whatever is in there (since dlys can be negative or positive). I've checked that the other convention breaks things.

jsdillon:
@nkern @zakiali good to go?

jsdillon:
Thanks all!

PR #234, opened 2018-03-24
Replace interp2d w lst rephase

nkern:
in `hera_cal.abscal_run`

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -980,7 +980,7 @@ def abscal_run(data_files, model_files, calfits_infiles=None, verbose=True, over
         echo("loading {}".format(dfile), type=1, verbose=verbose)
         (data, data_flags, data_antpos, data_ants, data_freqs, data_times, data_lsts,
             data_pols) = io.load_vis(dfile, pop_autos=True, return_meta=True, pick_data_ants=False)
-        bls = odict(map(lambda k: (k, antpos[k[0]] - antpos[k[1]]), data.keys()))
+        bls = odict(map(lambda k: (k, data_antpos[k[0]] - data_antpos[k[1]]), data.keys()))
         Ntimes = len(data_times)
         Nfreqs = len(data_freqs)
         data_lsts[data_lsts < data_lsts[0]] += 2*np.pi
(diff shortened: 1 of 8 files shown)


hera_cal/abscal.py
```diff
@@ -997,19 +1000,19 @@ def abscal_run(data_files, model_files, calfits_infiles=None, verbose=True, over
...
+                for k in new_flags.keys():
+                    wgts[k][new_flags[k]] *= 0
 
             # reweight according to redundancy
             if reweight:
                 wgts = mirror_data_to_red_bls(wgts, model_antpos, tol=tol, weights=True)
 
             # instantiate class
-            AC = AbsCal(interp_model, data, wgts=wgts, antpos=antpos, freqs=data_freqs)
+            AC = AbsCal(new_model, data, wgts=wgts, antpos=antpos, freqs=data_freqs)
```
jsdillon: I think this will fail if rephase_model is False because then new_model will be undefined. Perhaps you need an `else: new_model = model` or something similar?

hera_cal/abscal_funcs.py
```diff
@@ -1113,6 +1112,66 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
     return DataContainer(new_model), DataContainer(new_flags)
 
 
+def rephase_vis(model, model_lsts, data_lsts, bls, freqs, flags=None, max_dlst=0.005, latitude=-30.72152):
```
jsdillon: `max_dlst = .005` corresponds to 69 seconds, which seems a bit long for a default. I'd bump it down to .002 or so.

hera_cal/abscal_funcs.py
```diff
@@ -1113,6 +1112,66 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
...
+    new_model = odict()
+    new_flags = odict()
+    for k in model.keys():
+        m = model[k][lst_nn, :]
+        new_model[k] = m
+        if flags is None:
+            new_flags[k] = np.zeros_like(m, np.bool)
+        else:
+            new_flags[k] = flags[k][lst_nn, :]
+        new_flags[k][flag_lst] = True
```
jsdillon: Should this be `new_flags[k][flag_lst,:] = True`. It might work as is, but I find being very explicit about dimensions helps both code readability and debugging.

hera_cal/abscal_funcs.py
nkern: well, for a 14.6 m baseline I don't think its terribly unreasonable. But for a longer baseline it probably isn't great. we probably should expose it.

hera_cal/abscal.py
nkern: good catch

PR #236, opened 2018-03-26
updated delay_filter_run such that residual and/or model can be written

nkern:
by default residual is written. model can be alternatively written w/ `--write_model`, or both can be written with `--write_both`.

This enables us to write to file the model and residual without having to re-run the `delay_filter`, which is a slow process.

This will only affect the IDR2.1 pipeline if it ever uses `delay_filter_run.py` with the `--write_model` flag, otherwise it shouldn't change anything in the pipeline.


jsdillon:
I'd rather have two different flags. One that is "write model" and one that is "write both." It should be possible to write only the model.

nkern:
good point. done.

nkern:
@adampbeardsley let me know if you have any comments, in particular if this will pose a significant problem for the `delay_filter_XRFI` algorithm, I suspect it won't, but let me know anyways..

PR #237, opened 2018-03-26
updated lstbin.lst_bin such that wrapped lst_arrays are handled better

nkern:
if the lst_array of a data_file was wrapped and lst_start was non-zero, the wrapped lsts were not being binned, so I included some checks and wrapping statements to catch this. a test is added to ensure its working properly.

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -196,7 +196,6 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, lst_start=None, lst
             if key in data:
                 pass
             elif switch_bl(key) in data:
-                print 'conjugating??'
                 # check to see if conj(key) exists in data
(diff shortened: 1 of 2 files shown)


hera_cal/lstbin.py
```diff
@@ -761,6 +763,10 @@ def make_lst_grid(dlst, lst_start=None, verbose=True):
 
     # shift grid by lst_start
     if lst_start is not None:
+        # enforce lst_start to be within 0-2pi, else replace with 0
+        if lst_start < 0 or lst_start > 2*np.pi:
+            abscal.echo("lst_start was < 0 or > 2pi, replacing with lst_start=0", verbose=verbose)
+            lst_start = 0.0
```
jsdillon: Might want to do `lst_start = lst_start % 2*np.pi`

PR #238, opened 2018-03-27
fixed bug in lstbin.lst_bin_files

nkern:
There was a bug in the wrapping and unwrapping of LST arrays in `lstbin.lst_bin_files`, as well as the in the implementation of `lst_rephase` (nothing wrong with the `lst_rephase` code itself). Also made some updates to `utils.get_miriad_times` that helps fix these problems.

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -611,8 +611,15 @@ def lst_bin_files(data_files, dlst=None, verbose=True, ntimes_per_file=60, file_
     miriad_kwargs['outdir'] = outdir
     miriad_kwargs['overwrite'] = overwrite
  
-    # get frequency, time and antennas position information from the zeroth data file
+    # get metadata from the zeroth data file
     d, fl, ap, a, f, t, l, p = io.load_vis(data_files[0][0], return_meta=True, pick_data_ants=False)
+
+    # push time and lst arrays forward by half an integration
+    abscal.echo("pushing time and lst arrays forward by half an integration b/c pyuvdata "\
+                "currently does not do this for us....", verbose=verbose)
+    t += np.median(np.diff(t)) / 2.0
+
+    # get frequency, time and antenna position information
     freq_array = copy.copy(f)
     antpos = copy.deepcopy(ap)
     start_jd = np.floor(t)[0]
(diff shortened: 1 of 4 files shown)


hera_cal/utils.py
```diff
@@ -411,10 +410,7 @@ def get_miriad_times(filepaths, add_int_buffer=False):
         stop = start + (uv['ntimes']-1) * int_time
         # add integration buffer to beginning and end if desired
         if add_int_buffer:
-            if i != 0:
```
jsdillon: The above comment needs to be updated.

hera_cal/lstbin.py
```diff
@@ -604,8 +597,15 @@ def lst_bin_files(data_files, dlst=None, verbose=True, ntimes_per_file=60, file_
...
     miriad_kwargs['overwrite'] = overwrite
  
-    # get frequency, time and antennas position information from the zeroth data file
+    # get metadata from the zeroth data file
     d, fl, ap, a, f, t, l, p = io.load_vis(data_files[0][0], return_meta=True, pick_data_ants=False)
+
+    # push time and lst arrays forward by half an integration
+    abscal.echo("pushing time and lst arrays forward by half an integration b/c pyuvdata "\
+                "currently does not do this for us....", verbose=verbose)
+    t += np.median(np.diff(t)) / 2.0
```
jsdillon: These kind of medians of diffs will give funny answers (nans, etc.) when nTimes is 1. I'd probably add in a statement that `assert(len(t) > 1)`.

hera_cal/lstbin.py
```diff
@@ -642,8 +642,16 @@ def lst_bin_files(data_files, dlst=None, verbose=True, ntimes_per_file=60, file_
...
                     d.phase_type = 'drift'
 
-                    # unwrap l
-                    l[l < start_lst] += 2*np.pi
+                    # unwrap l relative to itself
+                    l[l < l[0]] += 2*np.pi
+
+                    # push time and lst arrays forward by half an integration
+                    # b/c pyuvdata does not currently do this for us....
+                    l += np.median(np.diff(l)) / 2.0
```
jsdillon: Likewise here.

PR #239, opened 2018-03-31
fixed bug in abscal.apply_gains

nkern:
if an antenna existed in `data`, but was missing in gains it errored out. Now it just eliminates those visibilities from output `new_data`. Tests included to check for this.

### DIFF
--- a/hera_cal/abscal_funcs.py
+++ b/hera_cal/abscal_funcs.py
@@ -14,7 +14,7 @@ import functools
 import numpy as np
 from pyuvdata import UVCal, UVData
 from pyuvdata import utils as uvutils
-from hera_cal import omni, utils, firstcal, cal_formats, redcal, io
+from hera_cal import omni, utils, firstcal, cal_formats, redcal, io, apply_cal
 from hera_cal.datacontainer import DataContainer
 from scipy import signal
 from scipy import interpolate
--- a/hera_cal/abscal_funcs.py
+++ b/hera_cal/abscal_funcs.py
@@ -614,7 +614,8 @@ def merge_gains(gains):
 
 def apply_gains(data, gains, gain_convention='divide'):
     """
-    apply gain solutions to data
+    Apply gain solutions to data. If a requested antenna doesn't
+    exist in gains, eliminate associated visibilities from new_data.
 
     Parameters:
     -----------
@@ -659,6 +660,10 @@ def apply_gains(data, gains, gain_convention='divide'):
         # get gain keys
         g1 = (k[0], k[-1][0])
         g2 = (k[1], k[-1][1])
+
+        # ensure keys are in gains
+        if g1 not in gains or g2 not in gains:
+            continue
 
         # form visbility gain product
         vis_gain = gains[g1] * np.conj(gains[g2])
--- a/hera_cal/apply_cal.py
+++ b/hera_cal/apply_cal.py
@@ -25,9 +25,9 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
             'multiply' means V_true = gi gj* V_obs. Assumed to be the same for new_gains and old_gains.
     '''
     # get datatype of data_flags to determine if flags or wgts
-    if np.all([(df.dtype == np.bool) for df in data.values()]):
+    if np.all([(df.dtype == np.bool) for df in data_flags.values()]):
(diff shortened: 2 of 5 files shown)


jsdillon:
I feel pretty strongly that we should flag rather than eliminate visibilities from data, leaving the raw data unmodified. This would make abscal consistent with hera_cal.apply_cal and be less likely to break things later (if they assume that the number of visibilities is unchanged). Alternatively, `abscal` could call `recalibrate_in_place` which does that (and also allows for the propagation of input calibration flags, which isn't necessary now but might be some day): https://github.com/HERA-Team/hera_cal/blob/4d13364a271abcf935d883fb77511d60d20ea721/hera_cal/apply_cal.py#L7

nkern:
i think thats a smart suggestion!

hera_cal/abscal_funcs.py
```diff
@@ -614,7 +614,8 @@ def merge_gains(gains):
 
 def apply_gains(data, gains, gain_convention='divide'):
     """
-    apply gain solutions to data
+    Apply gain solutions to data. If a requested antenna doesn't
```
jsdillon: Can we just eliminate this function?

hera_cal/apply_cal.py
```diff
@@ -2,17 +2,19 @@
...
     '''Update data and data_flags in place, taking out old calibration solutions, putting in
     new calibration solutions, and updating flags from those calibration solutions. Previously 
     flagged data is left unmodified. Missing antennas from either the new gains or (if it's not None),
     the old gains are automatically flagged in the data's visibilities that involves those antennas.
     
     Arguments:
         data: DataContainer containing baseline-pol complex visibility data. This is modified in place.
-        data_flags: DataContainer containing data flags. This is modified in place.
+        data_flags: DataContainer containing data flags. This is modified in place. Can also be fed as a
+            data weights dictionary with float dtype.
```
jsdillon: If it is a float dtype, you should explain how the calibration operates (namely, that it treats weights == 0 as flags, but otherwise keeps weights the same).

hera_cal/apply_cal.py
```diff
@@ -21,6 +23,17 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
         gain_convention: str, either 'divide' or 'multiply'. 'divide' means V_obs = gi gj* V_true,
             'multiply' means V_true = gi gj* V_obs. Assumed to be the same for new_gains and old_gains.
     '''
+    # get datatype of data_flags to determine if flags or wgts
+    if data_flags[data_flags.keys()[0]].dtype == np.bool:
```
jsdillon: I'd recommend something like `if np.all([(df.dtype == np.bool) for df in data.values()]):` since it checks all entries and doesn't really take that much longer.

hera_cal/apply_cal.py
```diff
@@ -21,6 +23,17 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+        bool_flags = True
+    elif data_flags[data_flags.keys()[0]].dtype == np.float:
```
jsdillon: And likewise here.

hera_cal/apply_cal.py
```diff
@@ -21,6 +23,17 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+        bool_flags = False
+        wgts = data_flags
+        data_flags = DataContainer(dict(map(lambda k: (k, ~wgts[k].astype(np.bool)), wgts.keys())))
+    else:
+        raise ValueError("didn't recognize dtype of data_flags")
+
+    # loop over keys
     for (i,j,pol) in data.keys():
         if not np.all(data_flags[(i,j,pol)]):
```
jsdillon: I think this line needs to get updated to handle weights. This might break if all the weights are zero because the visibility is totally flagged.

hera_cal/abscal_funcs.py
nkern: i think we should keep it around for the time being, seeing as its a little different than `recalibrate_in_place`, it may have a use-case. Perhaps at some point after we get IDR2.1 finished we can do a cleaning-of-house for the whole repo?

jsdillon: Ok by me.

PR #242, opened 2018-04-04
Enable band edge flagging in apply_cal

jsdillon:
addresses #241

### DIFF
--- a/hera_cal/apply_cal.py
+++ b/hera_cal/apply_cal.py
@@ -74,7 +74,8 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
 
 
 def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibration=None, flags_npz=None, 
-              filetype='miriad', gain_convention='divide',  add_to_history='', clobber=False, **kwargs):
+              flag_nchan_low = 50, flag_nchan_high = 50, filetype='miriad', gain_convention='divide',  
+              add_to_history='', clobber=False, **kwargs):
     '''Update the calibration solution and flags on the data, writing to a new file. Takes out old calibration
     and puts in new calibration solution, including its flags. Also enables appending to history.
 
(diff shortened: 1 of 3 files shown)


hera_cal/apply_cal.py
```diff
@@ -74,7 +74,8 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
 
 
 def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibration=None, flags_npz=None, 
-              filetype='miriad', gain_convention='divide',  add_to_history='', clobber=False, **kwargs):
+              flag_nchan_low = 50, flag_nchan_high = 50, filetype='miriad', gain_convention='divide',
```
adampbeardsley: I think the default in the module should be to not flag band edges.

hera_cal/apply_cal.py
```diff
@@ -106,7 +109,12 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
         uvd.flag_array = np.logical_or(npz_flags['flag_array'], uvd.flag_array)
         add_to_history += ' FLAGS_NPZ_HISTORY: ' + str(npz_flags['history']) + '\n'
     data, data_flags = io.load_vis(uvd)
-    
+
+    # apply band edge flags
+    for bl in data_flags.keys():
+        data_flags[bl][:,0:flag_nchan_low] = True
+        data_flags[bl][:,-flag_nchan_high:] = True
```
adampbeardsley: setting `flag_nchan_high=0` will cause all data to be flagged here.

jsdillon: Good call. I'll fix this.

hera_cal/apply_cal.py
jsdillon: We want to set the default for now to 50 so that we can rerun part of the data without changing the makeflow. I have a separate PR to change the makeflow for future runs that will parameterize this there. If you want to open an issue to change it in the future, I'm OK with that, but I want it for the next week or so like this.

adampbeardsley: Can you rerun on a branch? It seems really odd to put temporary defaults into the master branch for a specific run.

Issue #204, opened 2018-03-07
Calibration and data times not quite matching 


jsdillon:
When developing smooth_cal, I built a function that checked the consistency of the data sets and calibration files that were loaded in.

One check I could not pass is the commented one here:
https://github.com/HERA-Team/hera_cal/blob/2cb03130577df697cfdc1537b19448600f48d7d1/hera_cal/smooth_cal.py#L176

Basically, I'm finding that the times in calfits files and the times in miriad files that are supposedly the same JD are not quite the same. 

Running `print sc.times - sc.data_times` after https://github.com/HERA-Team/hera_cal/blob/2cb03130577df697cfdc1537b19448600f48d7d1/scripts/notebooks/running_smooth_cal.ipynb
yields (in days):

```
[  0.00000000e+00   4.19095159e-09   8.84756446e-09   1.30385160e-08
   1.72294676e-08   2.14204192e-08   2.60770321e-08   3.02679837e-08
   3.44589353e-08   3.86498868e-08   4.37721610e-08   4.79631126e-08
   5.21540642e-08   5.63450158e-08   6.10016286e-08   6.51925802e-08
   6.93835318e-08   7.40401447e-08   7.82310963e-08   8.24220479e-08
   8.66129994e-08   9.12696123e-08   9.54605639e-08   9.96515155e-08
   1.03842467e-07   1.08499080e-07   1.12690032e-07   1.16880983e-07
   1.21071935e-07   1.25728548e-07   1.29919499e-07   1.34110451e-07
   1.38767064e-07   1.42958015e-07   1.47148967e-07   1.51339918e-07
   1.55996531e-07   1.60187483e-07   1.64378434e-07   1.68569386e-07
   1.73225999e-07   1.77416950e-07   1.81607902e-07   1.86264515e-07
   1.90455467e-07   1.94646418e-07   1.98837370e-07   2.03493983e-07
   2.07684934e-07   2.11875886e-07   2.16066837e-07   2.20723450e-07
   2.24914402e-07   2.29105353e-07   2.33296305e-07   2.37952918e-07
   2.42609531e-07   2.46800482e-07   2.51457095e-07   2.55648047e-07]
```

They start in agreement but the disagreement grows linearly at about 0.00037437 s per integration.

Any thoughts on what's going on here? @zakiali @adampbeardsley @dannyjacobs @plaplant @nkern ?

adampbeardsley:
[...]

PR #247, opened 2018-04-07
Warn and drop inconsistent prev/next files instead of erroring

jsdillon:
In getting assertion errors in smooth_cal because of a mysterious time gap between files: see https://github.com/HERA-Team/HERA_Commissioning/issues/113

I think the proper solution is that if prev/next data/cal is inconsistent with the file to be smoothed, we should just produce a warning and use mirroring (by pretending that a prev/next data/cal was never loaded in) instead of raising an AssertionError.

This PR implements that.


jsdillon:
This PR also fixes an issue where `xants` from antenna metrics causes an antenna to not be present in the previous/next calfits file (annoyingly, omnical does this). In that case, we add that antenna in but treat it as totally flagged. 

This appeared in our pipeline test on `/lustre/aoc/projects/hera/H1C_IDR2/IDR2_1/2458109/zen.2458109.12552.yy.HH.uv.abs.calfits` and `/lustre/aoc/projects/hera/H1C_IDR2/IDR2_1/2458109/zen.2458109.13297.yy.HH.uv.abs.calfits` because antenna 137 is (likely spuriously) identified as cross-polarized in `13297` but is not on the IDR2 list of bad antennas. See http://nbviewer.jupyter.org/github/HERA-team/HERA_plots/blob/master/data_inspect_2458109.ipynb

hera_cal/smooth_cal.py
```diff
@@ -206,30 +206,52 @@ def check_consistency(self):
...
-            #assert(np.all(self.next_times == self.next_data_times)) #TODO: revisit this
+            try:
+                assert(self.has_next_cal and self.has_next_data) #should have both or it doesn't make sense
+                assert(np.all(self.next_freqs == self.freqs))
+                assert(np.all(self.next_data_freqs == self.freqs))
+                assert(len(self.next_times) == len(self.next_data_times))
+                #test time contiguity
+                assert(np.abs(np.median(np.diff(self.times)) + self.times[-1] - self.next_times[0]) < 1e-6)
+                assert(np.abs(np.median(np.diff(self.times)) - np.median(np.diff(self.next_times))) < 1e-6)
+                #assert(np.all(self.next_times == self.next_data_times)) #TODO: revisit this
```
plaplant: Same as above

hera_cal/smooth_cal.py
```diff
@@ -206,30 +206,52 @@ def check_consistency(self):
...
+                for (ant,pol) in self.gains.keys():
```
plaplant: It seems like the corresponding for loop above was changed to be more flexible in terms of not requiring all antennas are consistent (i.e., the `assert` statements were replaced with an `if` statement). Should we also be less strict down here?

hera_cal/smooth_cal.py
```diff
@@ -206,30 +206,52 @@ def check_consistency(self):
...
+                    assert((ant, ant, pol+pol) in self.next_data) #assert data has autocorrelations
+            except AssertionError:
+                warnings.warn('next_cal and/or next_data is inconsistent with central cal/data and will not be used for time smoothing.')
+                self.has_next_cal, self.has_next_data = False, False
+
+            # if next_gain or next_data is missing an antenna (e.g. because it was in xants and omnical dropped it), treat it as flagged 
             for (ant,pol) in self.gains.keys():
-                assert((ant,pol) in self.next_gains.keys()) #assert next_gains has all the same keys
-                assert((ant, ant, pol+pol) in self.next_data) #assert data has autocorrelations
+                if (ant,pol) not in self.next_gains.keys() or (ant, ant, pol+pol) not in self.next_data.keys():
```
plaplant: I don't think the code will ever get down here if not all antennas match, because of the `assert` statements above. Seeing this loop, it makes me think we should get rid of those `assert`s.

hera_cal/smooth_cal.py
jsdillon: Yes. Good call. Meant to delete this loop.

hera_cal/smooth_cal.py
jsdillon: Good catch. Fixed.

hera_cal/smooth_cal.py
```diff
@@ -206,30 +206,52 @@ def check_consistency(self):
...
-            #assert(np.all(self.prev_times == self.prev_data_times)) #TODO: revisit this
+            try:
+                assert(self.has_prev_cal and self.has_prev_data) #should have both or it doesn't make sense
+                assert(np.all(self.prev_freqs == self.freqs))
+                assert(np.all(self.prev_data_freqs == self.freqs))
+                assert(len(self.prev_times) == len(self.prev_data_times))
+                #test time contiguity
+                assert(np.abs(np.median(np.diff(self.times)) - self.times[0] + self.prev_times[-1]) < 1e-6)
+                assert(np.abs(np.median(np.diff(self.times)) - np.median(np.diff(self.prev_times))) < 1e-6)
+                #assert(np.all(self.prev_times == self.prev_data_times)) #TODO: revisit this
```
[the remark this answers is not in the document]
jsdillon: Now that it's been solved (see #204), I've put back in a similar test and updated the testing calfits files so that they now pass.

jsdillon:
Ready for you to take another look @plaplant

Issue #246, opened 2018-04-07
Abscal key errors on 2458140


jsdillon:
Here's a log where it fails:

```
ABSCAL:


Error Suspected (no .out found) in ./2458140/zen.2458140.25241.xx.HH.uv.ABSCAL.xx.log
------------------------------------------------

Fri Apr  6 16:06:19 MDT 2018
[...]

PR #248, opened 2018-04-09
made default behavior in omni_abscal_run to use all_antenna_gains

nkern:
and enabled key check in abscal_run when merging out_gains with gain_dict


jsdillon:
This addresses #246

PR #249, opened 2018-04-09
fixed lst wrapping bug in lst_bin, also enabled output_file_selection…

nkern:
… in lst_bin_files


nkern:
the latter being a capability we need for the IDR2.1 makeflow pipeline parallelization

PR #256, opened 2018-04-16
scripts for IDR2 lstbin pipeline

nkern:
I wrote an lstbin pipeline using simple YAML parameter files and a `build_lstbin_pipe.sh` script that parses the YAML file and writes a PBS batch submission script.


nkern:
@plaplant I did it this way rather than using makeflow because I was unable to reconcile the fundamental difference between makeflow approach of "operate on individual input files" and the need for the lstbin pipeline which is "operate on individual output files".

nkern:
The intended workflow is as simple as
```
build_lstbin_pipe.sh lstbin_params.yaml
```
and then
```
qsub lstbin_pipe.sh
```

jsdillon:
One feature request: allow looping over all 4 polarizations with a single config file.

nkern:
Ok so this the `lstbin_params.yaml` now takes a
```
pols:
    - xx
    - yy
```
column and the result is that
```
build_lstbin_pipe.sh lstbin_params.yaml
```
now outputs
```
lstbin_pipe_xx.sh
lstbin_pipe_yy.sh
```
which you would `qsub` individually (only need to do this 4 times at most) and each job would produce a `lstbin_xx.out` log file.

hera_cal/lstbin.py
```diff
@@ -482,11 +482,8 @@ def lst_bin_arg_parser():
         "'2458042/zen.2458042.*.xx.HH.uv' '2458043/zen.2458043.*.xx.HH.uv' \n"
         "Consult lstbin.lst_bin_files() for further details on functionality.")
     a.add_argument('data_files', nargs='*', type=str, help="quotation-bounded, space-delimited, glob-parsable search strings to time-contiguous nightly data files")
-    a.add_argument("--lst_init", type=float, default=np.pi, help="starting point for universal LST grid")
```
plaplant: Why were these options removed? Just because these parameters are inferred from the files themselves now?

nkern: ah it was removed b/c it that capability was actually deprecated in previous PRs and I just caught it lingering around when making this PR...

Issue #257, opened 2018-04-16
Fix abscal's post-omni tip-tilt correction


jsdillon:
I noticed that tip-tilt correction does not produce flat relative phase slopes across the array between the abscal model and the omni+abscaled data. I believe that this is at the heart of the issue of the relative calibration between polarizations that leads to sources bouncing around, as Tasha saw by looking at Stokes-Q: https://files.slack.com/files-pri/T0KMWB8P3-F9WB73C6P/2458115.24482.uvocr_zoom.png

For example:

<img width="580" alt="screen shot 2018-04-16 at 3 49 11 pm" src="https://user-images.githubusercontent.com/5281139/38839549-0ba2d642-4190-11e8-8e07-1199316ffad1.png">

where we've averaged over redundant baselines. 

Although we pre-calibrate a delay slope over the array, we do not calibrate a global (i.e. frequency-independent) phase slope over the array. As a result, the abscaled data without the final tip-tilt corrections looks like:

<img width="590" alt="screen shot 2018-04-16 at 3 48 39 pm" src="https://user-images.githubusercontent.com/5281139/38840287-b4ccf5ec-4193-11e8-8903-283aaf6ee151.png">

This leads to phase wrapping that causes the fit of a slope at each frequency to not quite flatten the relative angle between the data and the abscal model as a function of position/baseline. 

The solution to the problem has two parts:

1. I think we should be fitting for the global phase slope across the array after the delay-slope calibration but before the per-frequency tip-tilt correction. Already, this has produced very promising results in the interactive notebook in my sandbox:

![screen shot 2018-04-16 at 4 36 30 pm](https://user-images.githubusercontent.com/5281139/38840420-65ccb6b6-4194-11e8-8df9-5917fa053146.png)

2. Second, I think we need to iterate our tip-tilt solver until it reaches some convergence is reached. I tried this without the global phase slope solver and it works pretty well. 

![screen shot 2018-04-16 at 4 38 39 pm](https://user-images.githubusercontent.com/5281139/38840473-b76df868-4194-11e8-95f6-41d1f37aebf6.png)

Doing both will likely make the convergence in step 2 faster, while the global fit can be quite fast (basically because I fit a median all frequencies).

jsdillon:
Addressed in #261.

Issue #254, opened 2018-04-11
tol is not documented in abscal_run


jsdillon:
https://github.com/HERA-Team/hera_cal/blob/ff7efd3b3a523a124ca8887451118014a5fc84e7/hera_cal/abscal.py#L855

jsdillon:
Addressed in #261.

Issue #253, opened 2018-04-11
As written, I don't think this supports a string, just a list of strings.


jsdillon:
https://github.com/HERA-Team/hera_cal/blob/ff7efd3b3a523a124ca8887451118014a5fc84e7/hera_cal/abscal.py#L880

jsdillon:
In fact, I think many of the inputs to abscal_run must be lists because you iterate over them without checking if they are strings.

jsdillon:
I think it's OK to force the input to be lists, but just clarify the documentation.

jsdillon:
Although it is confusing that output_calfits_fname is a string. Is there a situation where you'd load in many data files and then output one calfits file?

Issue #245, opened 2018-04-05
This line of the comment in abscal_func.py is not mathematically correct.


AaronParsons:
https://github.com/HERA-Team/hera_cal/blob/eb002532ebc4433683f60756a098518b15e60ea3/hera_cal/abscal_funcs.py#L101

nkern:
you mean because the RHS should read `angle(g_i_x - g_j_y)`? If so thats because thats just a typo in the doc-string (which should be fixed!). if you look in the code you'll see that its actually doing is

`ydata = odict([(k, np.angle(data[k]/model[k])) for k in keys])`

so the LHS of the doc-string is a correct representation of what its doing, and its then equating that to 

`Phi_i * r_i - Phi_j * r_j`, not `angle(g_i_x) * angle(g_j_y)`, which is definitely wrong b/c that's not even a linear system...

jsdillon:
Yeah, it's just that there's an issue in the docstring.

jsdillon:
Addressed in #261.

PR #261, opened 2018-04-20
Fix abscal tip tilt correction with global phase slope solver and iterative TT_phs_cal

jsdillon:
This PR addresses the issue with how abscal's post-omnical tip-tilt phase correction works, as I explained in #257. 

Along the way, I've also addressed #245, #253, and #254, which are mostly related to standardizing inputs and clarifying documentation in abscal.


hera_cal/abscal_funcs.py
```diff
@@ -585,6 +585,90 @@ def delay_slope_lincal(model, data, antpos, wgts=None, df=9.765625e4, medfilt=Tr
 
     return fit
 
+def global_phase_slope_logcal(model, data, antpos, wgts=None, verbose=True, tol=1.0):
+    """
+    Solve for a frequency-independent spatial phase slope using the equation 
+
+    median_over_freq(V_ij,xy^data / V_ij,xy^model) = dot(Phi_x, r_i) - dot(Phi_y, r_j)
```
nkern: is there supposed to be an `angle` somewhere on the LHS of this eqn?

hera_cal/abscal_funcs.py
```diff
@@ -506,7 +506,7 @@ def delay_slope_lincal(model, data, antpos, wgts=None, df=9.765625e4, medfilt=Tr
     -------
     fit : dictionary containing delay slope (T_x) for each pol [seconds / meter].
     """
-    echo("...configuring linsolve data for delay_lincal", verbose=verbose)
+    echo("...configuring linsolve data for delay_slope_lincal", verbose=verbose)
```
nkern: thanks for the fix.. i've been meaning to fix this specifically for months... haha

hera_cal/abscal_funcs.py
jsdillon: Good call. Fixed.

hera_cal/abscal_funcs.py
```diff
@@ -585,6 +585,90 @@ def delay_slope_lincal(model, data, antpos, wgts=None, df=9.765625e4, medfilt=Tr
...
+    
+    # build linear system
+    ls_data, ls_wgts = {}, {}
+    for rk in red_keys:
+        # build equation string 
+        eqn_str = '{}*Phi_ew_{} + {}*Phi_ns_{} - {}*Phi_ew_{} - {}*Phi_ns_{}'
+        eqn_str = eqn_str.format(antpos[rk[0]][0], rk[2][0], antpos[rk[0]][1], rk[2][0],
+                                 antpos[rk[1]][0], rk[2][1], antpos[rk[1]][1], rk[2][1])
+
+        # calculated frequency median of unflagged angle(data/model)
```
nkern: is it better to average the first data across freq, and then take `angle`?

hera_cal/abscal.py
```diff
@@ -44,17 +44,22 @@ class AbsCal(object):
         where T_dly is a delay slope in [ns / meter]
         and B_ij is the baseline vector between ant i and j.
 
-    5) Average amplitude linear calibration solves the equation:
+    5) frequency-independent phase slope calibration
+        median_over_freq(V_ij^data / V_ij^model) = dot(Phi, B_ji)
```
nkern: needs an `angle`

hera_cal/abscal_funcs.py
jsdillon: Haha, no problem.

Code reviews are one thing, but I definitely understand abscal a lot better now that I've had to *modify* it.

hera_cal/abscal.py
```diff
@@ -1075,66 +1202,74 @@ def abscal_run(data_files, model_files, calfits_infiles=None, verbose=True, over
...
+            io.write_cal(output_calfits_path, gain_dict, data_freqs, data_times, flags=flag_dict, quality=None, 
+                 total_qual=None, return_uvc=False, overwrite=overwrite, history=history)
 
     # form return tuple
     return_obj = ()
 
     # return gains if desired
     if return_gains:
-        return_obj += (gains,)
+        return_obj += (gain_dict,)
```
nkern: why was this changed?

hera_cal/tests/test_abscal.py
```diff
@@ -531,13 +555,13 @@ def test_omni_abscal_arg_parser(self):
         a = hc.abscal.omni_abscal_arg_parser()
 
     def test_abscal_run(self):
```
nkern: did you add an `abscal_run` call with `phs_slope_cal = True`

hera_cal/abscal.py
jsdillon: `gains` is no longer defined. It used to hold a list of `gain_dict`s, one per data file. But now there's just one data file.

hera_cal/tests/test_abscal.py
jsdillon: Done.

hera_cal/abscal.py
nkern: ah, oh yeah that's right.. okay cool

hera_cal/abscal_funcs.py
nkern: so what about phase wrapping issues... i.e. when the ratio of data / model isn't that close and oscillates between and around 0 and -2pi?

jsdillon: It's a fair question. Both methods suffer from phase wraps potentially... the hope is that the median fixes it either way as long as it doesn't phase wrap too much.

This is just my experience with one data file, which admittedly, is picked to be very similar to the calibrator. I think it's worth revisiting this if the fix doesn't stop sources from moving around in our images.

Also, the point of this is to just get things closer... the iterative TT will get us the rest of the way if there's lots of phase wrapping.

jsdillon:
Great. Thanks @nkern

PR #263, opened 2018-04-22
introduces reference antennas for abscal phase solvers

nkern:
such that the resultant gain phase of the reference antenna is identically zero across all frequencies

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -66,7 +66,8 @@ class AbsCal(object):
     pathologies, meaning that a delay calibration should generally precede a
     phs_logcal or a TT_phs_logcal bandpass routine.
     """
-    def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None, verbose=True):
+    def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None, 
+                 bl_cut=None, bl_taper_fwhm=None, verbose=True):
         """
         AbsCal object used to for phasing and scaling visibility data to an absolute reference model.
 
@@ -118,6 +119,12 @@ class AbsCal(object):
         freqs : ndarray of frequency array, type=ndarray, dtype=float
                 1d array containing visibility frequencies in Hz.
                 Needed for delay calibration.
+
+        bl_cut : float, eliminate all visibilities with baseline separation lengths
+            larger than bl_cut. This is assumed to be in ENU coordinates with units of meters.
+
+        bl_taper_fwhm : float, impose a gaussian taper on the data weights as a function of
+            bl separation length, with a specified fwhm [meters]
         """
         # set pols to None
         pols = None
@@ -138,6 +145,7 @@ class AbsCal(object):
 
         # get shared keys
         self.keys = sorted(set(model.keys()) & set(data.keys()))
+        assert len(self.keys) > 0, "no shared keys exist between model and data"
 
         # append attributes
         self.model = model
@@ -194,13 +202,43 @@ class AbsCal(object):
         self.bls = None
         if self.antpos is not None:
             # center antpos about reference antenna
-            self.antpos = odict(map(lambda k: (k, antpos[k] - antpos[self.refant]), antpos.keys()))
+            self.antpos = odict(map(lambda k: (k, antpos[k] - antpos[self.refant]), self.ants))
             self.bls = odict([(x, self.antpos[x[0]] - self.antpos[x[1]]) for x in self.keys])
             self.antpos_arr = np.array(map(lambda x: self.antpos[x], self.ants))
             self.antpos_arr -= np.median(self.antpos_arr, axis=0)
 
         # setup gain solution keys
         self._gain_keys = map(lambda p: map(lambda a: (a, p), self.ants), self.gain_pols)
+
+        # perform baseline cut
+        if bl_cut is not None:
+            assert self.antpos is not None, "can't request a bl_cut if antpos is not fed"
+
+            # restrict visibilities to bl_cut criterion
+            _model = odict()
+            _data = odict()
+            _wgts = odict()
+            for k in self.keys:
+                if np.linalg.norm(self.bls[k]) <= bl_cut:
+                    _model[k] = self.model[k]
+                    _data[k] = self.data[k]
+                    _wgts[k] = self.wgts[k]
+
+            _model = DataContainer(_model)
+            _data = DataContainer(_data)
+            _wgts = DataContainer(_wgts)
+
+            # re-init
+            self.__init__(_model, _data, refant=self.refant, wgts=_wgts, antpos=self.antpos, freqs=self.freqs, verbose=verbose)
+
+        # enact a baseline weighting taper
+        if bl_taper_fwhm is not None:
+            assert self.antpos is not None, "can't request a baseline taper if antpos is not fed"
+            # make gaussian taper func
+            taper = lambda ratio: np.exp(-0.5*ratio**2)
+            # iterate over baselines
+            for k in self.wgts.keys():
+                self.wgts[k] *= taper(np.linalg.norm(self.bls[k]) / bl_taper_fwhm)
 
     def amp_logcal(self, verbose=True):
         """
@@ -955,6 +993,8 @@ def abscal_arg_parser():
     a.add_argument("--gen_phs_cal", default=False, action='store_true', help='perform general antenna phase bandpass calibration')
     a.add_argument("--max_dlst", default=0.005, type=float, help="maximum allowed LST difference in model rephasing, otherwies model is flagged.")
     a.add_argument("--refant", default=None, type=int, help="antenna number integer to use as reference antenna.")
+    a.add_argument("--bl_cut", default=None, type=float, help="cut visibilities w/ baseline length large than bl_cut [meters].")
+    a.add_argument("--bl_taper_fwhm", default=None, type=float, help="enact gaussian weight tapering based on baseline length [meters] with specified FWHM.")
     return a
 
 
@@ -982,11 +1022,13 @@ def omni_abscal_arg_parser():
     a.add_argument("--TT_phs_conv_crit", type=float, default=1e-6, help="convergence criterion in Delta g / g for stopping iterative TT_phs_cal")
     a.add_argument("--max_dlst", default=0.005, type=float, help="maximum allowed LST difference in model rephasing, otherwies model is flagged.")
     a.add_argument("--refant", default=None, type=int, help="antenna number integer to use as reference antenna.")
+    a.add_argument("--bl_cut", default=None, type=float, help="cut visibilities w/ baseline length large than bl_cut [meters].")
+    a.add_argument("--bl_taper_fwhm", default=None, type=float, help="enact gaussian weight tapering based on baseline length [meters] with specified FWHM.")
     return a
 
 
 def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose=True, overwrite=False, write_calfits=True,
-               output_calfits_fname=None, return_gains=False, return_object=False, outdir=None,
+               bl_cut=None, bl_taper_fwhm=None ,output_calfits_fname=None, return_gains=False, return_object=False, outdir=None,
                match_red_bls=False, tol=1.0, reweight=False, rephase_model=True, all_antenna_gains=False,
                delay_cal=False, avg_phs_cal=False, delay_slope_cal=False, phase_slope_cal=False, abs_amp_cal=False,
                TT_phs_cal=False, TT_phs_max_iter=10, TT_phs_conv_crit = 1e-6, gen_amp_cal=False, gen_phs_cal=False, 
@@ -1036,6 +1078,12 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
     return_gains : type=boolean, if True, return AbsCal gain dictionary
 
     return_object : type=boolean, if True, return AbsCal object
+
+    bl_cut : float, eliminate all visibilities with baseline separation lengths
+        larger than bl_cut. This is assumed to be in ENU coordinates with units of meters.
+
+    bl_taper_fwhm : float, impose a gaussian taper on the data weights as a function of
+        bl separation length, with a specified fwhm [meters]
 
     match_red_bls : type=boolean, match unique data baselines to model baselines based on redundancy
 
@@ -1150,7 +1198,7 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
             wgts = mirror_data_to_red_bls(wgts, model_antpos, tol=tol, weights=True)
 
         # instantiate class
-        AC = AbsCal(new_model, data, wgts=wgts, refant=refant, antpos=antpos, freqs=data_freqs)
+        AC = AbsCal(new_model, data, wgts=wgts, refant=refant, antpos=antpos, freqs=data_freqs, bl_cut=bl_cut, bl_taper_fwhm=bl_taper_fwhm)
         total_gain_keys = flatten(map(lambda p: map(lambda k: (k, p), total_data_antpos.keys()), AC.gain_pols))
 
         gain_list = []
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -67,8 +67,7 @@ class AbsCal(object):
     phs_logcal or a TT_phs_logcal bandpass routine.
     """
 
-    def __init__(self, model, data, wgts=None, antpos=None, freqs=None,
-                 model_ftype='miriad', data_ftype='miriad', verbose=True):
+    def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None, verbose=True):
         """
         AbsCal object used to for phasing and scaling visibility data to an absolute reference model.
 
@@ -87,9 +86,6 @@ class AbsCal(object):
                 Optionally, model can be a path to a miriad or uvfits file, or a
                 pyuvdata.UVData object, or a list of either.
 
-        model_ftype : type=str, if model is a path to a file, this is its filetype
-                      options=['miriad', 'uvfits']
-
         data : visibility data of measurements, type=DataContainer
                keys are antenna pair + pol tuples (must match model), values are
                complex ndarray visibilities matching shape of model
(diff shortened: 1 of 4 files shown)


hera_cal/abscal.py
```diff
@@ -233,6 +238,8 @@ def phs_logcal(self, avg=False, verbose=True):
 
         Parameters:
         -----------
+        refant : antenna number integer to use as reference antenna, whose phase is set to zero
```
jsdillon: You don't actually need this here, since it's not a keyword argument of the function.

hera_cal/abscal.py
```diff
@@ -185,12 +182,20 @@ def __init__(self, model, data, wgts=None, antpos=None, freqs=None,
...
+            assert refant in self.ants, "refant {} not found in self.ants".format(refant)
+        self.refant = refant
 
         # setup antenna positions
         self.antpos = antpos
         self.antpos_arr = None
         self.bls = None
         if self.antpos is not None:
+            # center antpos about reference antenna
+            self.antpos = odict(map(lambda k: (k, antpos[k] - antpos[self.refant]), antpos.keys()))
```
jsdillon: Is this actually necessary? I thought absolute antenna position didn't matter, just relative antenna position.

hera_cal/abscal.py
```diff
@@ -465,12 +472,14 @@ def abs_amp_logcal(self, verbose=True):
         self._abs_eta = odict(map(lambda k: (k, copy.copy(fit["eta_{}".format(k[1])])), flatten(self._gain_keys)))
         self._abs_eta_arr = np.moveaxis(map(lambda pk: map(lambda k: self._abs_eta[k], pk), self._gain_keys), 0, -1)
 
-    def TT_phs_logcal(self, verbose=True, zero_psi=False, four_pol=False):
+    def TT_phs_logcal(self, refant=None, verbose=True, zero_psi=True, four_pol=False):
```
jsdillon: Why aren't you using self.refant?

hera_cal/abscal.py
```diff
@@ -515,6 +524,9 @@ def TT_phs_logcal(self, verbose=True, zero_psi=False, four_pol=False):
         self._TT_Phi = odict(map(lambda k: (k, copy.copy(np.array([fit["Phi_ew_{}".format(k[1])], fit["Phi_ns_{}".format(k[1])]]))), flatten(self._gain_keys)))
         self._TT_Phi_arr = np.moveaxis(map(lambda pk: map(lambda k: np.array([self._TT_Phi[k][0], self._TT_Phi[k][1]]), pk), self._gain_keys), 0, -1)
 
+        # project out an absolute phase such that reference antenna has identically zero phase
```
jsdillon: Did you mean to delete this comment?

hera_cal/abscal.py
```diff
@@ -1238,6 +1254,13 @@ def abscal_run(data_file, model_files, calfits_infile=None, verbose=True, overwr
         gain_dict = merge_gains(gain_list)
         flag_dict = odict(map(lambda k: (k, np.zeros((Ntimes, Nfreqs), np.bool)), gain_dict.keys()))
 
+        # ensure reference antenna phase has been projected out (i.e. set to zero)
+        for p in AC.gain_pols:
+            refant_phs = gain_dict[(AC.refant, p)] / np.abs(gain_dict[(AC.refant, p)])
```
jsdillon: I'd call this something other than `refant_phs`... it's more like a unit vector in the complex plane than a phase.

hera_cal/abscal_funcs.py
```diff
@@ -92,7 +92,7 @@ def abs_amp_logcal(model, data, wgts=None, verbose=True):
     return fit
 
 
-def TT_phs_logcal(model, data, antpos, wgts=None, verbose=True, zero_psi=False,
+def TT_phs_logcal(model, data, antpos, wgts=None, refant=None, verbose=True, zero_psi=True,
```
jsdillon: I'd add some documentation explaining how zero_psi interacts with refant.

hera_cal/abscal.py
```diff
@@ -969,10 +982,11 @@ def omni_abscal_arg_parser():
     a.add_argument("--TT_phs_max_iter", type=int, default=10, help="maximum number of iterations of TT_phs_cal allowed")
     a.add_argument("--TT_phs_conv_crit", type=float, default=1e-6, help="convergence criterion in Delta g / g for stopping iterative TT_phs_cal")
     a.add_argument("--max_dlst", default=0.005, type=float, help="maximum allowed LST difference in model rephasing, otherwies model is flagged.")
+    a.add_argument("--refant", default=None, type=int, help="antenna number integer to use as reference antenna.")
     return a
 
 
-def abscal_run(data_file, model_files, calfits_infile=None, verbose=True, overwrite=False, write_calfits=True,
+def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose=True, overwrite=False, write_calfits=True,
```
jsdillon: You've centered the antpos that lives inside AC, but I don't think you've centered `total_data_antpos` around the reference antenna. Is that a problem?

hera_cal/abscal_funcs.py
```diff
@@ -311,9 +325,24 @@ def phs_logcal(model, data, wgts=None, verbose=True):
...
+            ant2 += "*a1"
+        eqns.append((k, "{} - {}".format(ant1, ant2)))
+    eqns = odict(eqns)
     ls_design_matrix = odict()
 
+    # set reference antenna phase to zero
+    if refant is None:
+        refant = keys[0][0]
+    assert np.array(map(lambda k: refant in k, keys)).any(), "refant {} not found in data and model".format(refant)
+    ls_design_matrix['a1'] = 0.0
```
jsdillon: I'm confused by this. Doesn't this just make the coefficient of the reference anntenna's phase 0? This doesn't force it to be zero... it means that it could be anything and still satisfy the equations, no?

hera_cal/abscal_funcs.py
```diff
@@ -429,11 +462,26 @@ def delay_lincal(model, data, wgts=None, df=9.765625e4, solve_offsets=True, medf
...
+    eqns = odict(eqns)
 
     # setup design matrix dictionary
     ls_design_matrix = odict()
 
+    # set reference antenna delay to zero
+    if refant is None:
+        refant = keys[0][0]
+    assert np.array(map(lambda k: refant in k, keys)).any(), "refant {} not found in data and model".format(refant)
+    ls_design_matrix['a1'] = 0.0
```
jsdillon: Likewise here.

hera_cal/abscal_funcs.py
```diff
@@ -448,8 +496,15 @@ def delay_lincal(model, data, wgts=None, df=9.765625e4, solve_offsets=True, medf
...
-        ls_design_matrix = odict()
+        eqns = []
+        for i, k in enumerate(keys):
+            ant1 = "phi_{}_{}".format(k[0], k[2][0])
+            ant2 = "phi_{}_{}".format(k[1], k[2][1])
+            if refant == k[0]: ant1 += "*a1"
+            elif refant == k[1]: ant2 += "*a1"
+            eqns.append((k, "{}-{}".format(ant1, ant2)))
+        eqns = odict(eqns)
+        ls_design_matrix = odict([("a1", 0.0)])
```
jsdillon: And likewise here.

jsdillon:
Also, when you're done with this PR, do you want to expose the refant to the makeflow config by updating this PR and then pulling it in? https://github.com/HERA-Team/hera_opm/pull/25

Also, we should pick a reference antenna for (ideally) the whole IDR. Ideally it's one that's never flagged and near the center of the array.

hera_cal/abscal.py
nkern: ah yeah, that was a holdover from when i was allowing it to be passed to each function, thanks for the catch

hera_cal/abscal.py
nkern: same as before, if you look `self.refant` is actually used in the function. thanks for the catch

hera_cal/abscal.py
nkern: yup

hera_cal/abscal.py
nkern: i think `refant_phasor` is more appropriate

hera_cal/abscal.py
```diff
@@ -195,14 +202,45 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
-            self.antpos = odict(map(lambda k: (k, antpos[k] - antpos[self.refant]), antpos.keys()))
+            self.antpos = odict(map(lambda k: (k, antpos[k] - antpos[self.refant]), self.ants))
             self.bls = odict([(x, self.antpos[x[0]] - self.antpos[x[1]]) for x in self.keys])
             self.antpos_arr = np.array(map(lambda x: self.antpos[x], self.ants))
             self.antpos_arr -= np.median(self.antpos_arr, axis=0)
 
         # setup gain solution keys
         self._gain_keys = map(lambda p: map(lambda a: (a, p), self.ants), self.gain_pols)
 
+        # perform baseline cut
```
jsdillon: For namespace clarity and future utility, I'd break this out into its own function.

hera_cal/abscal.py
```diff
@@ -285,6 +321,11 @@ def delay_lincal(self, medfilt=True, kernel=(1, 11), time_ax=0, freq_ax=1, verbo
 
         time_avg : boolean, if True, average resultant antenna delays across time 
 
+        window : str, window to enact on data before FFT for dly solver, options=['blackmanharris', 'hann', None]
+            None is a top-hat window.
+
+        edge_cut : int, number of channels to flag at each band edge in FFT window
```
jsdillon: I think it's slightly clearer if the default is 0 rather than None, since that makes the type consistent.

hera_cal/abscal.py
jsdillon: Bumping this question.

hera_cal/abscal.py
```diff
@@ -1074,6 +1139,11 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
 
     max_dlst : type=float, maximum allowed LST difference in model rephasing, otherwies model is flagged.
 
+    window : str, window to enact on data before FFT for delay solver, options=['blackmanharris', 'hann', None]
+        None is a top-hat window.
+
+    edge_cut : int, number of channels to flag at each band edge in delay solvers
```
jsdillon: "flag" is a bit confusing, since we're not actually applying flags to that data. I'd say "exclude".

hera_cal/abscal_funcs.py
```diff
@@ -434,10 +439,11 @@ def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, solve_offse
         wgts[k][inf_select] = 0.0
 
         # get delays
-        dly, offset = fft_dly(ratio, wgts=wgts[k], df=df, medfilt=medfilt, kernel=kernel, time_ax=time_ax, freq_ax=freq_ax, solve_phase=solve_offsets)
+        dly, offset = fft_dly(ratio, wgts=wgts[k], df=df, medfilt=medfilt, kernel=kernel, time_ax=time_ax, 
+                              freq_ax=freq_ax, solve_phase=solve_offsets, window=window, edge_cut=edge_cut)
 
         # set nans to zero
-        rwgts = np.ones_like(dly, np.float)
+        rwgts = np.nanmean(wgts[k], axis=freq_ax, keepdims=True)
```
jsdillon: Strictly speaking, the mean of the weights isn't quite the optimal weight for the delay, but in practice it's probably fine. I'm not sure off the top of my head what would be optimal, but we can think about it some more if you care.

hera_cal/abscal_funcs.py
```diff
@@ -721,13 +732,10 @@ def global_phase_slope_logcal(model, data, antpos, wgts=None, refant=None, verbo
...
-        delta_phi[avg_flags[rk]] = np.nan
-        ls_data[eqn_str] = np.nanmedian(delta_phi, axis=1, keepdims=True)
-
-        # set weights based on redundancy of unflagged channels
-        for red in reds:
-            if rk in red or (rk[1],rk[0],rk[2][::-1]) in red:
-                ls_wgts[eqn_str] = np.sum([~flags[bl] for bl in red], axis=(0,2), keepdims=True)[0].astype(np.float)
+        avg_wgts[rk][np.isinf(delta_phi)+np.isnan(delta_phi)] = 0.0
+        delta_phi[np.isinf(delta_phi)+np.isnan(delta_phi)] = 0.0
+        ls_data[eqn_str] = np.median(delta_phi * avg_wgts[rk], axis=1, keepdims=True) / np.median(avg_wgts[rk], axis=1, keepdims=True)
```
jsdillon: This is definitely not correct. You can't do a weighted median like a weighted average. I think you want to put back in my loop over reds, but just loop over a sum of weights rather than a sum of ~flags.

hera_cal/abscal_funcs.py
```diff
@@ -1003,11 +1016,31 @@ def fft_dly(vis, wgts=None, df=9.765625e4, medfilt=True, kernel=(1, 11), time_ax
...
-    # fft
-    window = np.repeat(signal.windows.tukey(Nfreqs)[np.newaxis], Ntimes, axis=time_ax)
-    window = np.moveaxis(window, 0, time_ax)
-    window *= wgts
-    vfft = np.fft.fft(vis_smooth * window, axis=freq_ax)
+    # construct window
+    win = np.moveaxis(np.repeat(np.zeros(Nfreqs)[np.newaxis], Ntimes, axis=0), 0, time_ax)
+
+    if edge_cut is not None:
+        assert 2*edge_cut < Nfreqs, "edge_cut cannot be >= Nfreqs/2"
```
jsdillon: 2*edge_cut < Nfreqs+1 (since you need two frequencies to define a delay).

hera_cal/abscal_funcs.py
```diff
@@ -1003,11 +1016,31 @@ def fft_dly(vis, wgts=None, df=9.765625e4, medfilt=True, kernel=(1, 11), time_ax
...
+        win_slice = slice(edge_cut, Nfreqs-edge_cut)
+        win_Nfreqs = Nfreqs - 2*edge_cut
+    else:
+        win_slice = slice(None)
+        win_Nfreqs = Nfreqs
+
+    if window is None:
+        win[:, win_slice] = signal.windows.tukey(win_Nfreqs, 0.0)
```
jsdillon: In the documentation, you say 'None' means top-hat, but here you use Tukey. They should be consistent.

hera_cal/abscal_funcs.py
```diff
@@ -1003,11 +1016,31 @@ def fft_dly(vis, wgts=None, df=9.765625e4, medfilt=True, kernel=(1, 11), time_ax
...
+    elif window == 'blackmanharris':
+        win[:, win_slice] = signal.windows.blackmanharris(win_Nfreqs)
+    elif window == 'hann':
+        win[:, win_slice] = signal.windows.hann(win_Nfreqs)
+    else:
+        raise ValueError("didn't recognize window {}".format(window))
```
jsdillon: To make life easier for the user, I'd recommend telling them the allowable window strings here too.

hera_cal/abscal_funcs.py
```diff
@@ -1614,27 +1649,29 @@ def avg_data_across_red_bls(data, antpos, flags=None, broadcast_flags=True, medi
...
-    red_flags = odict()
+    red_wgts = odict()
 
     # iterate over reds
     for i, bl_group in enumerate(stripped_reds):
         # average redundant baseline group
         if median:
-            d = np.nanmedian(map(lambda k: data[k], bl_group), axis=0)
+            d = np.nanmedian(map(lambda k: data[k]*wgts[k], bl_group), axis=0)
+            d /= np.nanmedian(map(lambda k: wgts[k], bl_group), axis=0)
```
jsdillon: Once again, this formula is incorrect for the median. See, e.g. https://en.wikipedia.org/wiki/Weighted_median

hera_cal/abscal.py
nkern: ok!

hera_cal/abscal_funcs.py
nkern: a `singal.windows.tukey(Nfreqs, 0.0)` window is a top-hat window

hera_cal/abscal_funcs.py
nkern: i think i'll just opt to use a weighted mean here instead of the median, if that's ok

hera_cal/abscal_funcs.py
nkern: ok yeah we can discuss later

hera_cal/abscal_funcs.py
nkern: i think you mean 2*edge_cut < Nfreqs - 1

hera_cal/abscal_funcs.py
nkern: I think I'll just get rid of this capability seeing as I actually have never even used it, such that we just stick to weighted means..

hera_cal/abscal_funcs.py
jsdillon: Oh... can you at least make a comment to that effect?

hera_cal/abscal.py
```diff
@@ -1381,3 +1368,42 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         return return_obj
 
 
+
+def _cut_bls(datacontainers, bls, bl_cut):
```
jsdillon: Any reason this starts with an `_`?

Also, I'd recommend adding a unit test for this function.

hera_cal/abscal.py
jsdillon: Bumping again.

hera_cal/abscal_funcs.py
jsdillon: Yes, you're right.

hera_cal/abscal.py
nkern: yes. this is how you enact a reference antenna when solving for a phase slope, rather than a phase

jsdillon:
Ok, looks good to me. Let's pull it when when the tests pass. Nice work @nkern!

PR #274, opened 2018-04-29
More memory improvements to abscal

jsdillon:
Added extra garbage collection, reduced the number of gains being carried around in abscal, and allowed rephasing to happen in place.

Reduces the memory usage by 30% or more.

### DIFF
--- a/hera_cal/abscal_funcs.py
+++ b/hera_cal/abscal_funcs.py
@@ -1309,7 +1309,7 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
     return DataContainer(new_model), DataContainer(new_flags)
 
 
-def rephase_vis(model, model_lsts, data_lsts, bls, freqs, flags=None, max_dlst=0.005, latitude=-30.72152):
+def rephase_vis(model, model_lsts, data_lsts, bls, freqs, inplace=False, flags=None, max_dlst=0.005, latitude=-30.72152):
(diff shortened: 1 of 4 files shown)


hera_cal/abscal_funcs.py
```diff
@@ -1309,7 +1309,7 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
     return DataContainer(new_model), DataContainer(new_flags)
 
 
-def rephase_vis(model, model_lsts, data_lsts, bls, freqs, flags=None, max_dlst=0.005, latitude=-30.72152):
+def rephase_vis(model, model_lsts, data_lsts, bls, freqs, inplace=False, flags=None, max_dlst=0.005, latitude=-30.72152):
```
nkern: I think this function was originally `inplace` by default b/c this is required by the lstbinner. can we keep this default to True and make wherever abscal calls it feed it True?

jsdillon: To be clear, the function it calls (in hera_cal.utils) has default inplace=True. This function, in abscal_funcs, didn't expose that parameter until now.

I hesitate to change the default (which was effectively False before because it was hard-coded to be False) unless you think it's OK (it's your code after all).

jsdillon: Anyway, I'm gonna merge it in and if you don't like the default, we can change it later.

PR #280, opened 2018-05-01
added keyword argument passing to run_filter to enable

nkern:
more window functions, in particular the `tukey` window that takes a kwarg


zakiali:
tests are not passing. Looks like there's a typo in the options for delay filter

PR #284, opened 2018-05-04
Fixes broadcasting of weights for average red data

jsdillon:
This was causing the fits in `global_phase_slope_logcal` to be quite off because most baselines got completely flagged.

@nkern You might want to add some more unittests for this function. The different weighting schemes are not actually tested to show that they are working.


nkern:
thanks @jsdillon for the fix, i'll update doc and push to this branch

jsdillon:
I think the doc is right now, @nkern... it's the unittests that need some work.

PR #285, opened 2018-05-04
Make phase_slope_cal iterative

jsdillon:
This generally converges in a few iterations and it helps TT_phs_cal (which is slower) converge more quickly.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -986,7 +986,7 @@ def abscal_arg_parser():
     a.add_argument("--phase_slope_cal", default=False, action='store_true', help='perform frequency-indepdendent phase slope calibration')
     a.add_argument("--abs_amp_cal", default=False, action='store_true', help='perform absolute amplitude calibration')
     a.add_argument("--TT_phs_cal", default=False, action='store_true', help='perform Tip-Tilt phase slope calibration')
-    a.add_argument("--TT_phs_max_iter", type=int, default=100, help="maximum number of iterations of TT_phs_cal allowed")
+    a.add_argument("--phs_max_iter", type=int, default=100, help="maximum number of iterations of phase_slope_cal or TT_phs_cal allowed")
     a.add_argument("--TT_phs_conv_crit", type=float, default=1e-6, help="convergence criterion in Delta g / g for stopping iterative TT_phs_cal")
     a.add_argument("--gen_amp_cal", default=False, action='store_true', help='perform general antenna amplitude bandpass calibration')
     a.add_argument("--gen_phs_cal", default=False, action='store_true', help='perform general antenna phase bandpass calibration')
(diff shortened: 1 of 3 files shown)


hera_cal/abscal.py
```diff
@@ -1020,8 +1020,8 @@ def omni_abscal_arg_parser():
     a.add_argument("--phase_slope_cal", default=False, action='store_true', help='perform frequency-indepdendent phase slope calibration')
     a.add_argument("--abs_amp_cal", default=False, action='store_true', help='perform absolute amplitude calibration')
     a.add_argument("--TT_phs_cal", default=False, action='store_true', help='perform Tip-Tilt phase slope calibration')
-    a.add_argument("--TT_phs_max_iter", type=int, default=100, help="maximum number of iterations of TT_phs_cal allowed")
-    a.add_argument("--TT_phs_conv_crit", type=float, default=1e-6, help="convergence criterion in Delta g / g for stopping iterative TT_phs_cal")
+    a.add_argument("--phs_max_iter", type=int, default=100, help="maximum number of iterations of phase_slope_cal or TT_phs_cal allowed")
+    a.add_argument("--phs_conv_crit", type=float, default=1e-6, help="convergence criterion in Delta g / g for stopping iterative phase_slope_cal or TT_phs_cal")
```
nkern: is there a good reason why the `phs_conv_crit` is being used for both the global phase slope and the freq-dependent phase slope? If one is much slower than another, as you suggested, it might make sense to have two different knobs for these algorithms? However, I'm fine w/ keeping as is and adjusting if we see that in practice we find we need to..

nkern:
Can we propagate the `edge_cut` kwarg to the `global_phase_slope` algorithm? I think it would boil down to doing something like this
```
ls_data[eqn_str] = np.nanmedian(delta_phi[:, edge_cut:-edge_cut], axis=1, keepdims=True)
ls_wgts[eqn_str] = np.sum(avg_wgts[rk][:, edge_cut:-edge_cut], axis=1, keepdims=True)
```
around line 740 in `abscal_funcs.py`. If you do this, you might also want to add a check at the beginning of the functions along the lines of
```
assert 2*edge_cut < Nfreqs - 1, "edge_cut cannot be >= Nfreqs/2 - 1"
```

hera_cal/abscal.py
jsdillon: I'm just trying to avoid a huge proliferation of parameters and it seemed pretty odd to set these differently, to be honest. The one you want accurate is the final TT_phs, but the global slope is faster, so there's no real savings if you speed up the faster one.

jsdillon:
Ok, added edge_cut. It shouldn't be necessary because of the median, but it certainly doesn't hurt and it might make things go faster...

hera_cal/abscal_funcs.py
```diff
@@ -737,8 +740,8 @@ def global_phase_slope_logcal(model, data, antpos, wgts=None, refant=None, verbo
         delta_phi[binary_flgs] *= np.nan
         avg_wgts[rk][np.isinf(delta_phi)+np.isnan(delta_phi)] = 0.0
         delta_phi[np.isinf(delta_phi)+np.isnan(delta_phi)] *= np.nan
-        ls_data[eqn_str] = np.nanmedian(delta_phi, axis=1, keepdims=True)
-        ls_wgts[eqn_str] = np.sum(avg_wgts[rk], axis=1, keepdims=True)
+        ls_data[eqn_str] = np.nanmedian(delta_phi[:, edge_cut:(len(delta_phi)-edge_cut)], axis=1, keepdims=True)
```
nkern: are you sure this is right?  I think `len(delta_phi)` will give you the `Ntimes`, which is not what you want. what's the problem with doing 
```
if edge_cut > 0:
    ls_data[eqn_str] = np.nanmedian(delta_phi[:, edge_cut:-edge_cut], axis=1)
else:
    ls_data[eqn_str] = np.nanmedian(delta_phi, axis=1)
```
?

jsdillon: Fixed. Good catch.

PR #288, opened 2018-05-08
Remove PBS generation script

plaplant:
A recent PR in hera_opm (https://github.com/HERA-Team/hera_opm/pull/31) adds the capability to run LST binning as a batch job on the NRAO cluster. Once that PR is merged in, we should remove these scripts, which are now deprecated, to avoid confusion.


jsdillon:
@plaplant  Can we delete this branch?

plaplant:
It seems like the scripts are gone from master, so that's fine by me.

Issue #289, opened 2018-05-08
Abscal: Add flagging for when the sun is within X degrees of the horizon 


jsdillon:
Perhaps we also want to flag when the sun is a few degrees below the horizon too, so let's expose that parameter to makeflow.

If only part of the file is flagged in this way, we might want to run all of abscal and then impose the flags after the fact... that way we can use averages to get the delay slope right.

nkern:
addressed by #290 . i don't think we want to flag after the fact and let data pass through abscal that we know could be contaminated by the sun, *especially* if we are doing any averaging across time

PR #290, opened 2018-05-10
added solar_flag to utils and abscal_run

nkern:
adds `solar_flag` function in `utils` to flag data based on altitude of sun, and adds this into `abscal_run` addresses #289

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -1369,6 +1369,11 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         gain_pols = AC.gain_pols
         gain_keys = gain_dict.keys()
 
+        # join gain_flag_dict with flag_dict
+        for k in flag_dict:
+            if k in gain_flag_dict:
+                flag_dict[k] += gain_flag_dict[k]
+
     # make blank gains if no modelfiles
     else:
         gain_pols = set(flatten(map(lambda p: [p[0], p[1]], data_pols)))
@@ -1377,11 +1382,6 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         flag_dict = odict(map(lambda k: (k, np.ones((Ntimes, Nfreqs), np.bool)), gain_keys))
         if refant is None:
             refant = gain_keys[0][0]
-
-    # join gain_flag_dict with flag_dict
-    for k in flag_dict:
-        if k in gain_flag_dict:
-            flag_dict[k] += gain_flag_dict[k]
 
     # make extra calfits metadata
     total_qual = odict(map(lambda p: (p, np.ones((Ntimes, Nfreqs), np.float)), gain_pols))
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -1196,8 +1196,6 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         utils.solar_flag(data_flags, int(np.floor(np.mean(data_times))), time_array=data_times, 
                          flag_alt=solar_horizon, inplace=True)
 
-    # get wgts
-    wgts = DataContainer(odict(map(lambda k: (k, (~data_flags[k]).astype(np.float)), data_flags.keys())))
 
     # ensure nomodelfiles is False
     if nomodelfiles == False:
@@ -1210,11 +1208,15 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         if rephase_model:
             new_model, new_flags = rephase_vis(model, model_lsts, data_lsts, bls, data_freqs, inplace=True,
                                                flags=model_flags, latitude=latitude, max_dlst=max_dlst)
-            # set wgts to zero wheree model is flagged
+            # update data flags w/ model flags
             for k in new_flags.keys():
-                wgts[k][new_flags[k]] *= 0
+                if k in data_flags:
+                    data_flags[k] += new_flags[k]
         else:
             new_model = model
+
+        # get wgts
+        wgts = DataContainer(odict(map(lambda k: (k, (~data_flags[k]).astype(np.float)), data_flags.keys())))
 
         # reweight according to redundancy
         if reweight:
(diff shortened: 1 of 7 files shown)


hera_cal/abscal.py
```diff
@@ -998,6 +998,8 @@ def abscal_arg_parser():
     a.add_argument("--bl_taper_fwhm", default=None, type=float, help="enact gaussian weight tapering based on baseline length [meters] with specified FWHM.")
     a.add_argument("--window", default=None, type=str, help="window to enact on data before FFT in delay solvers, options=[None, 'blackmanharris', 'hann']")
     a.add_argument("--edge_cut", default=0, type=int, help="number of channels to flag on each band-edge in delay and global phase solvers.")
+    a.add_argument("--solar_flag", default=False, action='store_true', help="FLag data when the sun is above some specified altitude.")
```
jsdillon: Can we make this a single parameter. If you want to make the default to never flag, regardless of where the sun is, perhaps just make the "solar_flag_altitude" default to 90 degrees.

hera_cal/abscal.py
```diff
@@ -1038,7 +1042,7 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
                match_red_bls=False, tol=1.0, reweight=False, rephase_model=True, all_antenna_gains=False, window=None, edge_cut=0,
                delay_cal=False, avg_phs_cal=False, avg_dly_slope_cal=False, delay_slope_cal=False, phase_slope_cal=False, abs_amp_cal=False,
                TT_phs_cal=False, phs_max_iter=100, phs_conv_crit=1e-6, gen_amp_cal=False, gen_phs_cal=False, 
-               latitude=-30.72152, max_dlst=0.005, history=''):
+               latitude=-30.72152, max_dlst=0.005, solar_flag=False, solar_horizon=0.0, history=''):
```
jsdillon: Be sure to add this to the documentation of abscal_run

hera_cal/utils.py
```diff
@@ -530,3 +654,52 @@ def lst_rephase(data, bls, freqs, dlst, lat=-30.72152, inplace=True, array=False
         
     if inplace == False:
         return data
+
+
+def data_to_gain_flags(flags, gain_keys=None):
```
jsdillon: This function already exists in smooth_cal: https://github.com/HERA-Team/hera_cal/blob/6085944949e0ab21a85093989754454b8b6d055b/hera_cal/smooth_cal.py#L23

Let's figure out if either version is better and then keep that one. I think utils is a fine place for it.

hera_cal/abscal.py
```diff
@@ -1187,8 +1191,11 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
     data_ants = np.unique(map(lambda k: k[:2], data.keys()))
     data_antpos = odict(map(lambda k: (k, data_antpos[k]), data_ants))
 
-    # get wgts
-    wgts = DataContainer(odict(map(lambda k: (k, (~data_flags[k]).astype(np.float)), data_flags.keys())))
+    # solar flag
+    if solar_flag:
```
jsdillon: Shouldn't we also flag on times when the sun is up in the model?

hera_cal/abscal.py
```diff
@@ -1355,6 +1369,11 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         gain_pols = AC.gain_pols
         gain_keys = gain_dict.keys()
 
+        # join gain_flag_dict with flag_dict
+        for k in flag_dict:
+            if k in gain_flag_dict:
+                flag_dict[k] += gain_flag_dict[k]
```
jsdillon: What's the logic here? Were we missing flags before?

hera_cal/abscal.py
nkern: Sure thing, i'm happy to do that. in that case it would be default of -90 degrees though..

hera_cal/utils.py
```diff
@@ -310,6 +312,128 @@ def JD2RA(JD, longitude=21.42830, latitude=-30.72152, epoch='current'):
...
+    # get Sun locations
+    alts = np.array(map(lambda t: crd.get_sun(Time(t, format='jd')).transform_to(a).alt.value, jds))
+
+    if array:
+        return alts
+    else:
+        return alts[0]
+
+ 
+def solar_flag(flags, start_jd, time_array=None, flag_alt=0.0, longitude=21.42830, latitude=-30.72152,
```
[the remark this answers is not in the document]
nkern: I don't see what's 1) complicated about it and 2) what isn't doing much: there is no repetition or redundancy, so I don't think it could be expressed in fewer lines and also I think it's perfectly reasonable to have a function that flags entire datacontainers as something an analyst might want handy... what's your argument for why we should *decrease* modularity on this front?

hera_cal/utils.py
nkern: ah didn't see that, thanks for pointing it out. looks like yours accomplishes what I need, so i'm happy to just use that. can you confirm that your function is doing the same thing as mine is here?

hera_cal/utils.py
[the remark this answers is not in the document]
nkern: or, put another way, what's your few line solution?

hera_cal/abscal.py
jsdillon: If you only flag data *above* some altitude and you want to never flag by default, then the default altitude should be 90, since the sun is never above 90.

hera_cal/utils.py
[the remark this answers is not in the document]
jsdillon: Let me take a stab at it.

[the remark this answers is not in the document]
jsdillon: Thoughts? It passes your unit tests.

hera_cal/abscal.py
nkern: the flags were always False, so I'm just propagating the solar flags

nkern:
@jsdillon there was a reason why the `solar_flag` method was a little convoluted, and that's because the `get_solar_alt` function is not particularly fast, and the solar altitude vs. JD is a slowly varying function, therefore you can get significant speed-up by evaluating the function a few times and then interpolating when you need to. I'm going to combine parts of your version and my version of the function

jsdillon:
Ok to merge, assuming tests pass.

PR #292, opened 2018-05-11
Enables partial I/O in io.load_vis based on new pyuvdata capabilities

nkern:
Allows us to perform partial IO on miriad and uvfits files into datacontainers based on recent pyuvdata support of this functionality.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -16,7 +16,9 @@ jonesnum2str = {-5: 'x', -6: 'y'}
 jonesstr2num = {'x': -5, 'y': -6}
 
 
-def load_vis(input_data, return_meta=False, filetype='miriad', pop_autos=False, pick_data_ants=True, nested_dict=False):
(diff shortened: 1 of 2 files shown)


jsdillon:
@nkern wanna go over this PR this afternoon? I'm free after 3:30.

jsdillon:
Another idea that @zaki suggested: we could load a list of tuples and if the tuples are:

- length 1 (or strings rather than tuples): interpret them as polarizations and return all baselines with that polarization 
- length 2: interpret them as antennas and return all baselines with that antenna
- length 3: interpret them as baselines

Then we just need to disallow mixing types and this lets us get rid of a whole bunch of the keyword arguments.

jsdillon:
@nkern can we close this PR and delete this branch?

jsdillon:
I'd like to close this PR now, if that's ok with you @nkern.

Issue #291, opened 2018-05-11
Refactor smooth cal to work on a whole day of data


jsdillon:
Use flag npz files instead of data files for io and memory savings
Also switch the default window to Tukey.

Issue #283, opened 2018-05-04
Smooth_Cal: Make sure that there are no time discontinuities within prev_cal or next_cal


jsdillon:
https://github.com/HERA-Team/hera_cal/blob/85adb9a2f8736c1ea30ce23f675795fd20de63b2/hera_cal/smooth_cal.py#L210

jsdillon:
Irrelevant now.

Issue #229, opened 2018-03-21
Fix weighting in smooth_cal


jsdillon:
Right now, weights are given by 

[(unnormalized omnical chi^2 per antenna) / (abscaled autocorrelation)^2]^-1

This creates major problems at the band edges, for example:

![screen shot 2018-03-20 at 7 00 04 pm](https://user-images.githubusercontent.com/5281139/37691704-f0be414c-2c70-11e8-8dec-def77d4a997f.png)

A better noise proxy would be

[(unnormalized omnical chi^2 per antenna) / (raw autocorrelation)^2]^-1

But even this creates unpleasant edge effects:

![screen shot 2018-03-20 at 7 04 41 pm](https://user-images.githubusercontent.com/5281139/37691805-a84f5094-2c71-11e8-8ea8-8b50f562bfda.png)

So for now, I'm going to suggest binary weights (i.e. flagged or not). We should come back to this. I'm going to go make that the default in hera_opm.

jsdillon:
Irrelevant now, but we can revisit non-binary weights in the future.

PR #293, opened 2018-05-11
Refactor smooth_cal to allow for full-day calibration smoothing

jsdillon:
This addresses #291.

By getting rid of the notion of "next" and previous files, this eliminates the need for #283.

By removing chi^2-based weighting in favor of binary weights based on flags, this also makes #229 obsolete.


hera_cal/smooth_cal.py
```diff
@@ -171,247 +133,128 @@ def time_filter(gains, wgts, times, filter_scale = 120.0, nMirrors = 0):
...
+                 self.cal_times[cal], _) = io.load_cal(cal, return_meta=True)
+            
+        # load flags files
+        self.npzs = sorted(flags_npz_list)
+        if len(self.npzs) > 0:
+            self.npz_flags, self.npz_freqs, self.npz_times = odict(), odict(), odict()
+            for cal, npz in zip(self.cals, self.npzs):
+                self.npz_flags[cal] = synthesize_ant_flags(io.load_npz_flags(npz))
+                npz_dict = np.load(npz)
+                self.npz_freqs[cal] = np.unique(npz_dict['freq_array'])
```
AaronParsons: Can freq_array have non-unique elements?

hera_cal/smooth_cal.py
```diff
@@ -171,247 +133,128 @@ def time_filter(gains, wgts, times, filter_scale = 120.0, nMirrors = 0):
...
+                self.npz_times[cal] = np.unique(npz_dict['time_array'])
+                    
+        # set up time grid
+        all_file_times = np.array(self.cal_times.values()).ravel()
+        self.dt = np.median(np.diff(all_file_times))
+        self.time_grid = np.arange(all_file_times[0], all_file_times[-1]+self.dt, self.dt)
```
AaronParsons: may be possible that all_file_times[-1]+self.dt has a round-off error that makes it epsilon larger, so that it becomes the endpoint.  Suggest adding self.dt/2 to ensure this doesn't happen.

hera_cal/smooth_cal.py
```diff
@@ -171,247 +133,128 @@ def time_filter(gains, wgts, times, filter_scale = 120.0, nMirrors = 0):
...
+        self.time_indices = {cal: np.searchsorted(self.time_grid, times) for cal, times in self.cal_times.items()}
```
AaronParsons: Does this work if datatypes of time_grid and times are different?  Maybe want to search for nearest value, rather than for exact match?

hera_cal/smooth_cal.py
```diff
@@ -171,247 +133,128 @@ def time_filter(gains, wgts, times, filter_scale = 120.0, nMirrors = 0):
...
+        self.ants = sorted(list(set([k for gain in self.gains.values() for k in gain.keys()])))
+        self.gain_grids = {ant: np.ones((len(self.time_grid), len(self.freqs)), 
+                                        dtype=np.complex) for ant in self.ants}
+        self.flag_grids = {ant: np.ones((len(self.time_grid), len(self.freqs)), 
+                                        dtype=bool) for ant in self.ants}
+        for ant in self.ants:
+            for cal in self.cals:
+                if self.gains[cal].has_key(ant):
+                    self.gain_grids[ant][self.time_indices[cal],:] = self.gains[cal][ant]
+                    self.flag_grids[ant][self.time_indices[cal],:] = self.cal_flags[cal][ant]
```
AaronParsons: The way this is written precludes the possibility of a non-one-to-one mapping from gains to time_indices.  Would be better to use something like, e.g., aipy.utils.add2array to handle both cases.  otherwise, as written, there's a non-deterministic mapping of gains into grids in the multi-to-one mapping case.

hera_cal/io.py
```diff
@@ -660,4 +660,28 @@ def update_cal(infilename, outfilename, gains=None, flags=None, quals=None, add_
     cal.write_calfits(outfilename, clobber=clobber)
 
 
+def load_npz_flags(npzfile):
```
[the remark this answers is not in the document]
AaronParsons: Also, what do you want to do with the history stored in the xrfi npz?

hera_cal/smooth_cal.py
jsdillon: Fixed.

hera_cal/smooth_cal.py
jsdillon: Fixed.

hera_cal/smooth_cal.py
jsdillon: It's OK.

hera_cal/smooth_cal.py
jsdillon: Let's just enforce one-to-one mapping for now.

hera_cal/io.py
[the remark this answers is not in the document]
jsdillon: I'll open an issue.

PR #303, opened 2018-05-23
added ability to overwrite vis_units attribute

nkern:
of output visibility file when using `apply_cal`. Also, the `apply_cal_argparser` shouldn't parse the arguments in the function, but should return the un-parsed object, which is parsed at the script level. This is necessary if one is using an argparser interactively, one can feed the object an empty list to instantiate the argument namespace with its default values (thanks to @adampbeardsley for that tip).

### DIFF
--- a/hera_cal/apply_cal.py
+++ b/hera_cal/apply_cal.py
@@ -70,7 +70,7 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
 
 def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibration=None, flags_npz=None, 
               flag_nchan_low=0, flag_nchan_high=0, filetype='miriad', gain_convention='divide',  
-              add_to_history='', clobber=False, **kwargs):
+              add_to_history='', clobber=False, vis_units=None, **kwargs):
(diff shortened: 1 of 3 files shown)


scripts/apply_cal.py
```diff
@@ -4,7 +4,8 @@
 from hera_cal import apply_cal as ac
 import sys
 
-args = ac.apply_cal_argparser()
+a = ac.apply_cal_argparser()
+args = a.parse_args()
 ac.apply_cal(args.infile, args.outfile, args.new_cal, old_calibration=args.old_cal, flags_npz=args.flags_npz,
```
jsdillon: Don't you want to add vis_units to this?

hera_cal/apply_cal.py
```diff
@@ -131,6 +132,10 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
         add_to_history += ' OLD_CALFITS_HISTORY: ' + old_uvc.history + '\n'
         old_calibration, _ = io.load_cal(old_uvc)
 
+    # parse vis_units
```
jsdillon: I don't think this needs to be an explicitly enumerated script argument of this function...just pass this in as part of the kwargs in the script. I think it would be better to make this part of the argparser and let the script take care of things.

Issue #259, opened 2018-04-17
Abscal metrics: compare calibrated visibilities to abscal model


jsdillon:
I'm not sure if this belongs in hera_qm or here, but we need a way to compare absolutely calibrated visibilities to the model visibilities they are absolutely calibrated to. In theory, the differences should be noise-like. This could be some kind of chi^2 metric, for example.

jsdillon:
Based on my digging into in #311, I think the way forward is to implement chi^2 and chi^2 per antenna in abscal. For now, these would be unnormalized and would simply be something like: chi^2 = Sum(|V_abscaled - V_model|^2), though we may want to revisit the normalization when we get access to better noise models (especially in H2C).

I think the right thing to do is to store these into both the quality arrays (for per antenna) and the total quality array of the .abs.calfits (and so the .smooth_abs.calfits would inherit them too). Then we'd want to feed both the .omni.calfits and .abs.calfits into xrfi and use both total quality arrays to flag.

My preliminary digging into the file involved in the particular instance of a bad abscal solution in #311 shows clear evidence of abscal problems in this chi^2:

<img width="620" alt="screen shot 2018-06-14 at 7 52 49 pm" src="https://user-images.githubusercontent.com/5281139/41492042-b9104ca8-70b9-11e8-88d7-30eb07b8b8bb.png">

And, by way of comparison, here's what the corresponding omnical chi^2 looks like:

<img width="623" alt="screen shot 2018-06-14 at 7 52 43 pm" src="https://user-images.githubusercontent.com/5281139/41492047-c46662fe-70b9-11e8-905f-e7c3de00aa97.png">

I think the next step is to implement this in abscal so that it's an ingredient available for use in investigating new xrfi algorithms. @adampbeardsley heads up that we'll probably want to use both omni.calfits and abs.calfits in CAL_XRFI going forward (or whatever will replace it).

PR #326, opened 2018-06-21
Make abscal compare data and model to populate quality_array with chi^2

jsdillon:
This addresses #259.

### DIFF
--- a/hera_cal/utils.py
+++ b/hera_cal/utils.py
@@ -730,7 +730,7 @@ def synthesize_ant_flags(flags, threshold=0.0):
     return ant_flags
 
 
-def chisq(data, model, data_wgts, gains=None, gain_flags=None, by_pol=False,
+def chisq(data, model, data_wgts, gains=None, gain_flags=None, by_antpol=False,
           chisq=None, nObs=None, chisq_per_ant=None, nObs_per_ant=None):
     """Computes chi^2 defined as:
     
@@ -754,7 +754,8 @@ def chisq(data, model, data_wgts, gains=None, gain_flags=None, by_pol=False,
             which is interpreted as all gains are 1.0 (ie..e the data is already calibrated)
         gain_flags: optional dictionary mapping ant-pol keys like (1,'x') to a boolean flags waterfall
             with the same shape as the data. Default: None, which means no per-antenna flagging.
-        by_pol: if True, chisq and nObs are dictionaries mapping visibility polarizations to numpy arrays.
+        by_antpol: if True, chisq and nObs are dictionaries mapping antenna polarizations to numpy arrays.
+            Additionally, if by_antpol is True, cross-polarized visibilities are ignored.
         chisq: optional chisq to update (see below)
         nObs: optional nObs to update (see below). Must be specified if chisq is specified and must be 
             left as None if chisq is left as None.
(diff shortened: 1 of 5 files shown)


hera_cal/utils.py
```diff
@@ -731,3 +731,114 @@ def synthesize_ant_flags(flags, threshold=0.0):
             ant_flags[ap] = (ant_Nflag[ap] / ant_Nvis[ap]) > threshold
 
     return ant_flags
+
+
+def chisq(data, model, data_wgts, gains=None, gain_flags=None, by_antpol=False,
+          chisq=None, nObs=None, chisq_per_ant=None, nObs_per_ant=None):
+    """Computes chi^2 defined as:
+
+    chi^2 sum_ij(|data_ij - model_ij * g_i conj(g_j)| * wgts_ij)
```
nkern: missing equals sign

hera_cal/utils.py
```diff
@@ -731,3 +731,114 @@ def synthesize_ant_flags(flags, threshold=0.0):
...
+        data_wgts: dictionary or DataContainer mapping baseline-pol keys like (1,2,'xx') to real
+            weights with [0] axis time and [1] axis frequency. Weights are interpeted as 1/sigma^2
+            where sigma is the noise on the data (but not necessarily the model if gains are provided).
+            Flags are be expressed as data_wgts equal to 0, so (~flags) produces binary weights.
+        gains: optional dictionary mapping ant-pol keys like (1,'x') to a waterfall of complex gains to
+            be multiplied into the model (or, equivalently, divided out from the data). Default: None,
+            which is interpreted as all gains are 1.0 (ie..e the data is already calibrated)
+        gain_flags: optional dictionary mapping ant-pol keys like (1,'x') to a boolean flags waterfall
+            with the same shape as the data. Default: None, which means no per-antenna flagging.
+        by_antpol: if True, chisq and nObs are dictionaries mapping antenna polarizations to numpy arrays.
```
nkern: this kwarg name is a little confusing, but I don't have a better suggestion offhand. just something to think about. being a little more verbose with the kwarg name isn't a terrible thing...

hera_cal/utils.py
```diff
@@ -731,3 +731,114 @@ def synthesize_ant_flags(flags, threshold=0.0):
...
+
+    # build containers for chisq_per_ant and nObs_per_ant if not supplied
+    if chisq_per_ant is None and nObs_per_ant is None:
+        chisq_per_ant = {}
+        nObs_per_ant = {}
+    elif (chisq_per_ant is None) ^ (nObs_per_ant is None):
+        raise ValueError('Both chisq_per_ant and nObs_per_ant must be specified or nor neither can be.')
+
+    for bl in data.keys():
+        if model.has_key(bl) and data_wgts.has_key(bl) and (not by_antpol or bl[2][0] == bl[2][1]):
```
nkern: can you add a comment what the last conditional is doing? the first two are obvious but I don't quite understand why the two pols entries need to be the same?

hera_cal/utils.py
```diff
@@ -731,3 +731,114 @@ def synthesize_ant_flags(flags, threshold=0.0):
...
+
+            # include gain flags in data weights
+            assert np.isrealobj(data_wgts[bl])
+            if gain_flags is not None:
+                wgts = data_wgts[bl] * ~(gain_flags[ant1]) * ~(gain_flags[ant2])
+            else:
+                wgts = copy.deepcopy(data_wgts[bl])
+
+            # calculate chi^2
+            chisq_here = np.array(np.abs(model_here - data[bl]) * wgts, dtype=float)
```
nkern: perhaps use `np.asarray` here? the `np.array` means that the data will be copied, whereas `np.asarray` will not make a copy if the data already conforms to the desired standard.

nkern: and perhaps use `dtype=np.float64` just to be precise as to the data type

hera_cal/utils.py
jsdillon: fixed

hera_cal/utils.py
jsdillon: I'll make it `split_by_antpol`

hera_cal/utils.py
jsdillon: I've added a comment. It's because if you're doing a separate chi^2 for x and y, it doesn't make sense to include xy and yx visibilities. This is meant to more closely match omnical. For now, it doesn't matter; we don't have xy or yx abscal models.

hera_cal/utils.py
jsdillon: Done.

PR #328, opened 2018-06-28
Lstbin starting lst

nkern:
enables capability to enforce start of LST grid to `lst_start` kwarg, rather than automatically adjusting to LST of first data record.

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -483,6 +483,7 @@ def lst_bin_arg_parser():
     a.add_argument('data_files', nargs='*', type=str, help="quotation-bounded, space-delimited, glob-parsable search strings to time-contiguous nightly data files")
     a.add_argument("--dlst", type=float, default=None, help="LST grid bin width")
     a.add_argument("--lst_start", type=float, default=0, help="starting LST for binner as it sweeps across 2pi LST")
+    a.add_argunent("--fixed_start", action='store_true', default=False, help="If True, make the start of the LST grid equal to lst_start, rather than the LST of the first data record.")
     a.add_argument("--ntimes_per_file", type=int, default=60, help="number of LST bins to write per output file")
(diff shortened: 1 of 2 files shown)


hera_cal/lstbin.py
```diff
@@ -483,6 +483,7 @@ def lst_bin_arg_parser():
     a.add_argument('data_files', nargs='*', type=str, help="quotation-bounded, space-delimited, glob-parsable search strings to time-contiguous nightly data files")
     a.add_argument("--dlst", type=float, default=None, help="LST grid bin width")
     a.add_argument("--lst_start", type=float, default=0, help="starting LST for binner as it sweeps across 2pi LST")
+    a.add_argunent("--fixed_start", action='store_true', default=False, help="If True, make the start of the LST grid equal to lst_start, rather than the LST of the first data record.")
```
jsdillon: I think `fixed_start` is a confusing parameter name that doesn't really say what it means. I think instead it should be something like `no_lst_trimming` would be much clearer.

jsdillon:
(Pending test passage, of course.)

hera_cal/lstbin.py
nkern: What about `fixed_lst_start`. There is a separate function in `hera_pspec` called `trim_lst`, so I'd prefer to stay away from that specific name..

nkern:
@jsdillon I can't merge this, I think b/c there are unreviewed additions after you approved? can you reapprove and merge? thanks

jsdillon:
It was set so that reviews aren't dismissed when there's updates... my prior approval still stood. It just said "Required statuses must pass before merging" but all the checks had passed. Annoying.

Anyway, I'm undoing some of @plaplant's changes to the permissions (honestly, I find the bureaucracy tiresome and I want to leave it to the team's good judgement whether changes require further reviews), but I'm keeping the safeguard that only you, me, Aaron, Zaki, Adam, and Paul can push to master.

PR #331, opened 2018-07-10
DO NOT MERGE: Partial io

jsdillon:


jsdillon:
@nkern I found the last remaining failing unit test in the LST-binner and fixed it. Now we pass Travis as long as we import an old pyvudata. I think we can now review this PR in parallel with getting the whole up to date with pyuvdata's master branch.

I'll add reverting the hacks here to make travis work as an issue.

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+from hera_cal.utils import split_pol, conj_pol
+import collections
 
-polstr2num = {'I': 1, 'Q': 2, 'U': 3, 'V': 4, 'RR': -1, 'LL': -2, 'RL': -3, 'LR': -4, 'xx': -5, 'yy': -6, 'xy': -7, 'yx': -8}
-polnum2str = {val: key for key, val in polstr2num.items()}
 
-jonesnum2str = {-5: 'x', -6: 'y'}
-jonesstr2num = {'x': -5, 'y': -6}
+class HERACal(UVCal):
+    '''HERAData is a subclass of pyuvdata.UVCal meant to serve as an interface between
```
nkern: Typo: `HERACal`

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+from hera_cal.utils import polnum2str, polstr2num, jnum2str, jstr2num
+from hera_cal.utils import split_pol, conj_pol
+import collections
 
-polstr2num = {'I': 1, 'Q': 2, 'U': 3, 'V': 4, 'RR': -1, 'LL': -2, 'RL': -3, 'LR': -4, 'xx': -5, 'yy': -6, 'xy': -7, 'yx': -8}
-polnum2str = {val: key for key, val in polstr2num.items()}
 
-jonesnum2str = {-5: 'x', -6: 'y'}
-jonesstr2num = {'x': -5, 'y': -6}
+class HERACal(UVCal):
```
nkern: Although it falls in line given the naming convention of HERAData, I'm not a fan of the name `HERACal` given its similarity with the name of the software package. But I don't really care that much, so if you want to keep it I'm fine w/ it

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+
+        elif self.filetype in ['miriad', 'uvfits']:
+            for meta in self.HERAData_metas:
+                setattr(self, meta, None)  # no pre-loading of metadata
+        else:
+            raise NotImplementedError('Filetype ' + self.filetype + ' has not been implemented.')
+
+    def reset(self):
+        '''Resets all standard UVData attributes, potentially freeing memory.'''
+        super(HERAData, self).__init__()
```
nkern: just wondering if you find that this actually does free-up the amount of memory you expect?

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+
+    def get_metadata_dict(self):
+        ''' Produces a dictionary of the most useful metadata. Used as object
+        attributes and as metadata to store in DataContainers.
+
+        Returns:
+            metadata_dict: dictionary of all items in self.HERAData_metas
+        '''
+        antpos, ants = self.get_ENU_antpos()
+        ants = sorted(ants)
```
nkern: This is surely a bug...no? Why would you sort ants and not sort antpos and then directly zip them?

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+        antpairs = self.get_antpairs()
+        bls = [antpair + (pol,) for antpair in antpairs for pol in pols]
+
+        times_by_bl = {antpair: np.array(self.time_array[self._blt_slices[antpair]])
+                       for antpair in antpairs}
+        lsts_by_bl = {antpair: np.array(self.lst_array[self._blt_slices[antpair]])
+                      for antpair in antpairs}
+
+        locs = locals()
+        return {meta: eval(meta, {}, locs) for meta in self.HERAData_metas}
```
nkern: I hear its not good practice to use the built-in `eval` function, as its not safe. Perhaps try
```
locs = locals()
return odict([(meta, locs[meta]) for meta in self.HERAData_metas])
```
just a suggestion though...

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+
+    def _determine_pol_indexing(self):
+        '''Determine the mapping between polnums and indices
+        in the polarization axis of the data_array.'''
+        self._polnum_indices = {}
+        for i, polnum in enumerate(self.polarization_array):
+            self._polnum_indices[polnum] = i
+
+    def _get_slice(self, data_array, key):
+        '''Return a copy of the Nint by Nfreq waterfall or waterfalls for a given key. Abstracts
```
nkern: It took me a little while after reading this to realize that `data_array` has the _full_ data size, and is not just the data_array specified by `key`. Perhaps add to the docstring that this function selects out the `key` part of `data_array`, just to be a little clearer..

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+                return np.array(data_array[self._blt_slices[tuple(key[0:2])], 0, :,
+                                           self._polnum_indices[polstr2num(key[2])]])
+            except KeyError:
+                return np.conj(data_array[self._blt_slices[tuple(key[1::-1])], 0, :,
+                                          self._polnum_indices[polstr2num(conj_pol(key[2]))]])
+        else:
+            raise KeyError('Unrecognized key type for slicing data.')
+
+    def _set_slice(self, data_array, key, value):
+        '''Update data_array with Nint by Nfreq waterfall(s). Abstracts away both baseline
```
nkern: same here in terms of elaborating on what `data_array` should be

hera_cal/io.py

hera_cal/io.py
jsdillon: I get that, though I haven't heard the objection from anyone else so it might just be a matter of taste. If I hear more complaints, it's not too hard to change.

hera_cal/io.py
jsdillon: Not as much or as consistently as I'd like, but yes it does clear up some memory. I tested this with %memit

hera_cal/io.py
jsdillon: Good call. Fixed.

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+            for meta in self.HERAData_metas:
+                setattr(self, meta, None)  # no pre-loading of metadata
+        else:
+            raise NotImplementedError('Filetype ' + self.filetype + ' has not been implemented.')
+
+    def reset(self):
+        '''Resets all standard UVData attributes, potentially freeing memory.'''
+        super(HERAData, self).__init__()
+
+    def get_metadata_dict(self):
```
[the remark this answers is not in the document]
jsdillon: I don't really see the point of passing `center`... no one ever doesn't center.

On pick_data_ants, I think you're right.

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+        self._blt_slices = {}
+        for ant1, ant2 in self.get_antpairs():
+            indices = self.antpair2ind(ant1, ant2)
+            if len(indices) == 1:
+                self._blt_slices[(ant1, ant2)] = slice(indices[0], indices[0] + 1, self.Nblts)
+            elif not (len(set(np.ediff1d(indices))) == 1):
+                raise NotImplementedError('UVData objects with non-regular spacing of '
+                                          'baselines in its baseline-times are not supported.')
+            else:
+                self._blt_slices[(ant1, ant2)] = slice(indices[0], indices[-1] + 1,
```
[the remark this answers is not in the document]
jsdillon: Yes, because it uses `np.where`.

hera_cal/io.py
```diff
@@ -7,26 +7,537 @@
...
+
+    def _determine_blt_slicing(self):
+        '''Determine the mapping between antenna pairs and
+        slices of the blt axis of the data_array.'''
+        self._blt_slices = {}
+        for ant1, ant2 in self.get_antpairs():
+            indices = self.antpair2ind(ant1, ant2)
+            if len(indices) == 1:
+                self._blt_slices[(ant1, ant2)] = slice(indices[0], indices[0] + 1, self.Nblts)
+            elif not (len(set(np.ediff1d(indices))) == 1):
```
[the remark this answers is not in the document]
jsdillon: Done.

hera_cal/io.py
jsdillon: Done.

hera_cal/io.py
jsdillon: Done.

jsdillon:
Ok, I think I'm good to go. @nkern do you approve?

jsdillon:
Actually, @nkern can you take a look at the LST-binning tests?

hera_cal/io.py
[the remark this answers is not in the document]
nkern: Okay that sounds fine, and just FYI `center=True` is the default, so if you don't want to center you need to explicitly pass `center=False`

PR #342, opened 2018-07-14
copied old pyuvdata.utils.top2eq_m funcs to hera_cal.utils for rephasing

nkern:
As a quick fix to recent deprecation of `pyuvdata.utils.top2eq_m` and `eq2top_m`, I copied over the old functions to `hera_cal.utils` for the time being. At some point in the future we may consider using some of the new functions used for phasing in `pyuvdata`.

Also updated various modules and tests given `pyuvdata`'s updated handling of the `time_array` and `lst_array` attributes (`pyuvdata` standard is bin-center, while miriad is bin-left, and `pyuvdata` only recently started actually doing this conversion properly).


nkern:
I agree w/ your point 2. in principle, except it seems that `pyuvdata` didn't have any unittests for these two functions directly, but just had unittests for other functions that used these functions, which is the same in our case: we have unittests for phasing code that rely on these functions

jsdillon:
Fair enough. I approve once you get tests working.

nkern:
So the `omni.py` test failure seems like whats happening, is that an old file was written to disk and saved in the `data/` directory, and now a test in `test_omni.py` compares a function output to that file, but with the new way of writing the `uvw_array` in `pyuvdata.UVData`, the test fails b/c the `uvw`s are slightly off. I can just overwrite that test file with the new output, but want to make sure with @zakiali that's a reasonable solution

zakiali:
Is the test comparing the pyuvdata objects read in from file and one generated from scratch? If that's the case I think it's reasonable to rewrite the file.

nkern:
@zakiali it seems like its one generated from a previous commit hash, saved to disk, and all future commits are compared against it

zakiali:
Ok. Seems arbitrary, so I'm fine with using a new test file.

jsdillon:
> At some point in the future we may consider using some of the new functions used for phasing in pyuvdata.

@nkern do we want to make an issue of eliminating these in favor of the new pyuvdata phasing code?

Issue #336, opened 2018-07-11
Get apply_cal ready for IDR3 / H2C RTP


jsdillon:
This includes:

* Making it python 2 and python 3 compatible (#318)
* Implementing partial data loading (#319) using the HERAData class in #331

jsdillon:
Addressed in #344.

PR #344, opened 2018-07-26
Partial I/O for apply_cal

jsdillon:
This addresses #336.

I think I've also made apply_cal and its tests and script python 3 compatible, but I haven't double checked that yet.


hera_cal/apply_cal.py
```diff
@@ -71,84 +78,103 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
 
-    recalibrate_in_place(data, data_flags, new_gains, new_flags, old_gains=old_calibration, gain_convention=gain_convention)
-    io.update_vis(data_infilename, data_outfilename, filetype_in=filetype, filetype_out=filetype, data=data,
-                  flags=data_flags, add_to_history=add_to_history, clobber=clobber, **kwargs)
+    # partial loading and writing using uvh5
+    if nbl_per_load is not None:
+        if not (filetype_in is 'uvh5') or not (filetype_out is 'uvh5'):
+            raise NotImplementedError('Partial writing is not implemented for non-uvh5 I/O.')
+        hd = io.HERAData(data_infilename, filetype='uvh5')
+        for attribute, value in kwargs.items():
```
plaplant: What types of kwargs would be passed in? Should we have the `__setattr__` statement be in a try/except block?

hera_cal/apply_cal.py
```diff
@@ -71,84 +78,103 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+            hd.__setattr__(attribute, value)
+        for data, data_flags, _ in hd.iterate_over_bls(Nbls=nbl_per_load):
+            for bl in data_flags.keys():
+                # apply band edge flags
+                data_flags[bl][:, 0:flag_nchan_low] = True
+                data_flags[bl][:, data_flags[bl].shape[1] - flag_nchan_high:] = True
+                # apply npz flags
+                if flags_npz is not None:
+                    data_flags[bl] = np.logical_or(data_flags[bl], npz_flag_dc[bl])
```
plaplant: Should a test be done to see if `npz_flag_dc` has flags corresponding to `bl`? Basically, could this cause a `KeyError`?

hera_cal/apply_cal.py
```diff
@@ -71,84 +78,103 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+        hd = io.HERAData(data_infilename, filetype=filetype_in)
+        hd.read()
+        # apply npz flags
+        if flags_npz is not None:
+            hd.flag_array = np.logical_or(npz_flags['flag_array'], hd.flag_array)
+        data, data_flags, _ = hd.build_datacontainers()
+        for bl in data_flags.keys():
+            # apply band edge flags
+            data_flags[bl][:, 0:flag_nchan_low] = True
+            data_flags[bl][:, data_flags[bl].shape[1] - flag_nchan_high:] = True
```
plaplant: Same as above about indexing.

hera_cal/apply_cal.py
```diff
@@ -71,84 +78,103 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+            # apply npz flags
+            if flags_npz is not None:
+                data_flags[bl] = np.logical_or(data_flags[bl], npz_flag_dc[bl])
```
plaplant: Same question about keys in `npz_flag_dc`

hera_cal/apply_cal.py
```diff
@@ -71,84 +78,103 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+            hd.partial_write(data_outfilename, data=data, flags=data_flags,
+                             inplace=True, clobber=clobber, add_to_history=add_to_history)
+
+    # full data loading and writing
+    else:
+        hd = io.HERAData(data_infilename, filetype=filetype_in)
+        hd.read()
+        # apply npz flags
+        if flags_npz is not None:
+            hd.flag_array = np.logical_or(npz_flags['flag_array'], hd.flag_array)
```
plaplant: Can this fail if the contents of `npz_flags` is not the same shape as the `hd.flag_array`? Should a check be done to ensure that?

hera_cal/apply_cal.py
```diff
@@ -1,5 +1,12 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 the HERA Project
+# Licensed under the MIT License
```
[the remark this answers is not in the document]
jsdillon: Done.

hera_cal/apply_cal.py
```diff
@@ -71,84 +78,103 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+        old_hc = io.HERACal(old_calibration)
+        old_calibration, _, _, _ = old_hc.read()
+        add_to_history += ' OLD_CALFITS_HISTORY: ' + old_hc.history + '\n'
 
-    recalibrate_in_place(data, data_flags, new_gains, new_flags, old_gains=old_calibration, gain_convention=gain_convention)
-    io.update_vis(data_infilename, data_outfilename, filetype_in=filetype, filetype_out=filetype, data=data,
-                  flags=data_flags, add_to_history=add_to_history, clobber=clobber, **kwargs)
+    # partial loading and writing using uvh5
+    if nbl_per_load is not None:
+        if not (filetype_in is 'uvh5') or not (filetype_out is 'uvh5'):
```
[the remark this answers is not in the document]
jsdillon: Done.

hera_cal/apply_cal.py
jsdillon: Basically, it's any pyuvdata parameter. I'm just copying the functionality here:     https://github.com/HERA-Team/hera_cal/blob/c6e563e823d71cdc0b5f0b5baa15bd54438170c0/hera_cal/io.py#L831

I don't think we need a try/except block.

hera_cal/apply_cal.py
```diff
@@ -71,84 +78,103 @@ def recalibrate_in_place(data, data_flags, new_gains, cal_flags, old_gains=None,
...
+            raise NotImplementedError('Partial writing is not implemented for non-uvh5 I/O.')
+        hd = io.HERAData(data_infilename, filetype='uvh5')
+        for attribute, value in kwargs.items():
+            hd.__setattr__(attribute, value)
+        for data, data_flags, _ in hd.iterate_over_bls(Nbls=nbl_per_load):
+            for bl in data_flags.keys():
+                # apply band edge flags
+                data_flags[bl][:, 0:flag_nchan_low] = True
+                data_flags[bl][:, data_flags[bl].shape[1] - flag_nchan_high:] = True
```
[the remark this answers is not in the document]
jsdillon: That would break when flag_nchan_high is 0.

hera_cal/apply_cal.py
jsdillon: Same as above.

hera_cal/apply_cal.py
jsdillon: Yes this could cause a KeyError. I honestly think that's fine. I'll add a clarification to the flags_npz docstring.

hera_cal/apply_cal.py
jsdillon: Same response as above.

hera_cal/tests/test_apply_cal.py
```diff
@@ -80,57 +87,70 @@ def test_recalibrate_in_place(self):
...
+        new_data, new_flags, _ = hd.read()
+        self.assertEqual(hd.vis_units, 'Jy')
         for k in new_data.keys():
             for i in range(new_data[k].shape[0]):
                 for j in range(new_data[k].shape[1]):
                     if not new_flags[k][i, j]:
-                        self.assertAlmostEqual(new_data[k][i, j] / 25.0, data[k][i, j], 4)
-                    if j < 300 or j > 923:
+                        self.assertAlmostEqual(new_data[k][i, j] / 25.0 / data[k][i, j], 1.0, 4)
+                    if j < 450 or j > 623:
```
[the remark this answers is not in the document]
jsdillon: They come from `flag_nchan_low=450, flag_nchan_high=400` (623 = 1024 - 400 - 1). I'll add some comments.

hera_cal/apply_cal.py
jsdillon: Yes, but I think it's fine to just let it fail. I don't think we need a check...but the updated docstring should be enough.

hera_cal/apply_cal.py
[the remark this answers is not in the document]
plaplant: Ah, good point!

hera_cal/apply_cal.py
plaplant: Okay, it'd be nice to have a more graceful failure mode (like maybe it raises a warning but keeps going), but we can keep it this way if you want.

jsdillon:
Thanks for the thorough review @plaplant!

Issue #337, opened 2018-07-11
Get delay_filter ready for IDR3 / H2C RTP


jsdillon:
This includes:

* Making it python 2 and python 3 compatible (#318)
* Implementing partial data loading (#319) using the HERAData class in #331
* ~~~Replacing flagged data with the CLEAN model (but keeping it flagged) #282~~~

jsdillon:
Done in #347.

PR #347, opened 2018-08-02
Enable delay filter using partial I/O

jsdillon:
This addresses #337.


jsdillon:
Also @adampbeardsley this will affect your delay-filter based xrfi stuff. It will let you do partial data loading using **read_kwargs in `load_data`.

hera_cal/io.py
```diff
@@ -877,9 +898,37 @@ def update_vis(infilename, outfilename, filetype_in='miriad', filetype_out='miri
         raise TypeError("Input filetype must be either 'miriad', 'uvfits', or 'uvh5'.")
 
 
+def to_HERACal(input_cal):
+    '''Converts a string path, UVData, or HERAData object, or a list of any one of those, to a
```
zakiali: Fix doc string. I think this should say "Converts a string path, UVCal, or HERACal..."

Issue #349, opened 2018-08-06
fft_dly bug


jsdillon:
When fft_dly is called with solve_phase=False, `dlys /= df` is never called and the returned answer has the wrong units. 

While
```
        # test mock data
        tau = np.array([1.5e-8]).reshape(1, -1)  # 15 nanoseconds
        f = np.linspace(0, 100e6, 1024)
        df = np.median(np.diff(f))
        r = np.exp(1j * 2 * np.pi * f * tau)
        dly, offset = hc.abscal.fft_dly(r, df=df, medfilt=True, kernel=(1, 5))
        nt.assert_almost_equal(float(dly), 1.5e-8, delta=1e-9)
```
passes, 
```
        # test mock data
        tau = np.array([1.5e-8]).reshape(1, -1)  # 15 nanoseconds
        f = np.linspace(0, 100e6, 1024)
        df = np.median(np.diff(f))
        r = np.exp(1j * 2 * np.pi * f * tau)
        dly, offset = hc.abscal.fft_dly(r, df=df, medfilt=True, kernel=(1, 5), solve_phase=False)
        nt.assert_almost_equal(float(dly), 1.5e-8, delta=1e-9)
```
fails.

jsdillon:
Fixed in #350.

Issue #262, opened 2018-04-20
Improve abscal tests


jsdillon:
Looking over abscal's tests, I am noticing that most of the tests of the major calibration routines in abscal_funcs are only tested through their AbsCal object interfaces, and even then they just assert that a result of the right shape is generated. 

Ideally, we'd do more comprehensive end-to-end tests where we put in fake data and fake model with known relative calibration and get out exactly that relative calibration. This would be much more robust and would likely have revealed some of the issues we're dealing with now (i.e. #257). See, for example, https://github.com/HERA-Team/hera_cal/blob/f6ff057230434b9d64dfcff6df082e5643fa81dc/hera_cal/tests/test_redcal.py#L292

jsdillon:
Additionally, abscal's tests were not broken by the update in https://github.com/HERA-Team/linsolve/pull/21, which means that they they are not testing for non-unitary weights (redcal had the same problem, see #322).

jsdillon:
Another thing that's come up. Looking at fft_dly, I've realized that the tests are really deficient. They make sure that there's an output with the right shape, but otherwise.

- There's no test that the phase offsets are right
- There's no test that the windowing or the edge_cut works properly. 
- There's no test that shows that the median filter works.

PR #350, opened 2018-08-06
Fix fft_dly

jsdillon:
While fixing #349, I decided to rewrite a large part of fft_dly.

1. I fixed the issue in #349 where delay units weren't being corrected
2. I used a more standard, robust, and readable approach to quadratic interpolation
3. I removed the ability to specify the time and frequency dimension in order to make the code more readable and shorter.
4. I condensed and simplified window code.
5. I made medfilt and phases default false, since they drastically slow down the code.

In the process, I noticed that the unit tests really don't cover much of the functionality of fft_dly... only that it actually produces an output. Namely, there's no test that the phase offsets are right. There's no test that the windowing or the edge_cut works. And there's no test that shows that the median filter works. I'm going to add this all as to #262.

### DIFF
--- a/hera_cal/abscal_funcs.py
+++ b/hera_cal/abscal_funcs.py
@@ -365,8 +365,8 @@ def phs_logcal(model, data, wgts=None, refant=None, verbose=True):
     return fit
 
 
-def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, solve_offsets=True, medfilt=True, kernel=(1, 5),
-                 verbose=True, time_ax=0, freq_ax=1, antpos=None, four_pol=False, window=None, edge_cut=0):
+def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, solve_offsets=True, medfilt=True,
+                 kernel=(1, 5), verbose=True, antpos=None, four_pol=False, window=None, edge_cut=0):
     """
     Solve for per-antenna delays according to the equation
 
@@ -404,10 +404,6 @@ def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, solve_offse
     medfilt : type=boolean, median filter visiblity ratio before taking fft
 
     kernel : type=tuple, dtype=int, kernel for multi-dimensional median filter
-
-    time_ax : type=int, time axis of model and data
-
-    freq_ax : type=int, freq axis of model and data
 
     antpos : type=dictionary, antpos dictionary. antenna num as key, position vector as value.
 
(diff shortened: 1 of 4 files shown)


nkern:
@jsdillon thanks for these improvements and bug fixes! One thing you can do to increase readability is to get rid of all of the `np.moveaxis(np.median` calls in favor of `np.median(... keepdims=True)`. I wrote that stuff before I knew you could do `keepdims=True` :)

hera_cal/abscal_funcs.py
```diff
@@ -996,111 +986,73 @@ def array_axis_to_data_key(data, array_index, array_keys, key_index=-1, copy_dic
...
-
-    peak_shifts = np.array([get_peak(np.take(amp, i, axis=time_ax), np.take(argmax, i, axis=time_ax)[0]) for i in range(Ntimes)])
-    dlys += np.moveaxis(peak_shifts.reshape(-1, 1) * dfreq, 0, time_ax)
+    # fft w/ window and find argmax
+    vfft = np.fft.fft(data * win * wgts, axis=1)
+    amp = np.abs(np.fft.fftshift(vfft))
+    argmaxes = np.argmax(amp, axis=1)
+    fftfreqs = np.fft.fftshift(np.fft.fftfreq(Nfreqs, df))
+    dlys, peak_shifts = np.zeros((Ntimes, 1)), np.zeros((Ntimes, 1))
+    # use parabolic peak interpolation: https://ccrma.stanford.edu/~jos/sasp/Quadratic_Interpolation_Spectral_Peaks.html
```
nkern: A few comments on the new peak finder / shifter:
1) Is it faster or slower than the one before? The reason I adopted the previous approach was that it was about as fast as way as I could think of doing this, and its an analytic result. 2) It seems like what you've implemented is actually quite similar (`0.5 * something / something`), and 3) does it give the same or nearly similar result as the old function? I'm happy to go with this new implementation if the speed isn't significantly slower and its getting similar if not the same results as before

nkern:
you mentioned you fixed a bug on units, can you add a test to ensure that bug is fixed?

hera_cal/abscal_funcs.py
jsdillon: Timing on a (4354, 1024) complex array:

New version takes 365 ms, old version takes 332 ms. The results are identical to numerical precision, so I think we're just writing the same algorithm differently.

And for the record, the old version using the default kwargs takes 2m 57s (the 332 ms is without the medfilt or the phase finding).

jsdillon:
Regarding the `np.moveaxis` stuff, I wasn't planning on changing the phase-finding code. If you want to simplify it when you address #262, that's cool.

Regarding the bug fix on units, by setting `solve_phase=False` in this line:

`dly, offset = hc.abscal.fft_dly(r, df, medfilt=True, kernel=(1, 5), solve_phase=False)`

I exposed the error. The problem was effectively an indentation problem: the `dlys /= df` should have been run regardless of whether solve_phase was true or not.

PR #351, opened 2018-08-09
apply_cal.recalib to handle single spectra, and HERAData filetype checking

nkern:
1. Updated the way `recalibrate_in_place` applies calibration flags to enable broadcasting, such that we can take single-integration spectra and apply it to waterfalls (data update is already done in such a way that enables broadcasting)

2. Fixed the `filetype` checking in `HERAData` to not use `filetype is ___` but rather `filetype == ___`. This was causing a bug when `filetype is 'miriad'` was False, even though `filetype = 'miriad'`. See https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is-in-python why its better to use `==` in this case.


jsdillon:
Looks fine to me!

Issue #320, opened 2018-06-14
Remove _apply_gains from redcal and use the code in apply_cal


jsdillon:
This addresses part of #100.

jsdillon:
Addressed in #353.

PR #353, opened 2018-08-20
Update redcal apply to remove duplicated functionality

jsdillon:
This addresses #320.

Unfortunately, this wasn't quite as simple as I hoped. `apply_cal.recalibrate_in_place` requires data_flags and gain_flags and also assumes that missing gains are flagged. Here we want to make a different assumption, which is that missing gains are 1.0. I had to leave `divide_by_gains` and turn it into a wrapper around `apply_cal.recalibrate_in_place` to make it perform the way we want.


jsdillon:
Depending on how you plan implement firstcal and omnical, we might want to keep the original functionality as is and just close the issue. I'm torn...

jsdillon:
@AaronParsons I think the ability to use wgts as flags (and update them appropriately in place) is functionality we want in hera_cal. However, your point that implicit behavior based on type is dangerous is well-taken. In the spirit of "explicit is better than implicit", I made an explicit boolean flag to trigger this behavior.

jsdillon:
Thanks @AaronParsons

PR #356, opened 2018-08-24
Omni update

AaronParsons:
Might need to to innoculate hera_cal to pyuvdata polarization changes until those are settled in.


hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
...
+        ex_ants (optional): antenna numbers (as int's) to exclude in reds.
+        ubls (optional): baselines representing their redundant group to include in reds.
+        ex_ubls (optional): baselines representing their redundant group to exclude in reds.
+        pols (optional): polarizations to include in reds. e.g. 'xy' or 'yx'.
+        ex_pols (optional): polarizations to exclude in reds. e.g. 'xy' or 'yx'.
+    Return:
+        reds: list of lists of redundant baselines as antenna pair tuples.
+    '''
+    if pols is None: # if no pols are provided, deduce them from the reds
+        pols = set(gp[0][2] for gp in reds)
```
jsdillon: This isn't quite as general as it could be: in the case where two different pols are considered redundant, this doesn't necessarily get all pols.

I'd instead do `pols = set([bl[2] for gp in reds for bl in gp])`.

hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
     return add_pol_reds(pos_reds, pols=pols, pol_mode=pol_mode, ex_ants=ex_ants)
 
 
+def filter_reds(reds, bls=None, ex_bls=None, ants=None, ex_ants=None, ubls=None, ex_ubls=None, pols=None, ex_pols=None):
+    '''
+    Filter redundancies to include/exclude the specified bls, antennas, unique bl groups and polarizations.
+    Assumes reds indices are Antpol objects.
```
jsdillon: I'd clarify in the documentation that ex_* always overrides *, so if you include a pol and then ex it, it's not included.

hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
...
+        pols (optional): polarizations to include in reds. e.g. 'xy' or 'yx'.
+        ex_pols (optional): polarizations to exclude in reds. e.g. 'xy' or 'yx'.
+    Return:
+        reds: list of lists of redundant baselines as antenna pair tuples.
+    '''
+    if pols is None: # if no pols are provided, deduce them from the reds
+        pols = set(gp[0][2] for gp in reds)
+    if ex_pols:
+        pols = set(p for p in pols if p not in ex_pols)
+    reds = [gp for gp in reds if gp[0][2] in pols]
```
jsdillon: Same comment. This should be `reds = [[bl for bl in gp if bl[2] in pols] for gp in reds if np.any([bl[2] in pols for bl in gp])]`

hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
...
+    def expand_bls(gp):
```
jsdillon: I'd recommend adding a comment to explain each of these internal functions.

hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
...
+        gp3 = [(g[0],g[1],p) for g in gp if len(g) == 2 for p in pols]
+        return gp3 + [g for g in gp if len(g) == 3]
+    antpols = set(sum([list(split_pol(p)) for p in pols],[]))
+    def expand_ants(gp):
+        gp2 = [(g,p) for g in gp if not hasattr(g,'__len__') for p in antpols]
+        return gp2 + [g for g in gp if hasattr(g,'__len__')]
+    def split_bls(bls):
+        return [split_bl(bl) for bl in bls]
+    if ubls or ex_ubls:
```
jsdillon: I find this to be more difficult to parse than `if (ubls is not None) or (ex_ubls is not None):`. Even though lists and strings evaluate to `True`, it's a quirk of python that's not particularly intuitive.

jsdillon: Also, same for `if ex_pols` above and more examples below.

hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
...
+        bl2gp = {}
```
jsdillon: I think you could probably use a comment here explaining what this dictionary is mapping.

hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
     return add_pol_reds(pos_reds, pols=pols, pol_mode=pol_mode, ex_ants=ex_ants)
 
 
+def filter_reds(reds, bls=None, ex_bls=None, ants=None, ex_ants=None, ubls=None, ex_ubls=None, pols=None, ex_pols=None):
+    '''
+    Filter redundancies to include/exclude the specified bls, antennas, unique bl groups and polarizations.
+    Assumes reds indices are Antpol objects.
+    Args:
+        reds: list of lists of redundant baseline tuples, e.g. (ind1,ind2,pol)
+        bls (optional): baselines as antenna pair tuples (i,j) to include in reds.
```
jsdillon: You should clarify that this supports (i,j) and (i,j,pol) formats.

hera_cal/redcal.py
```diff
@@ -147,6 +150,71 @@ def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0,
...
+        bls = set(k for k in bls if k not in ex_bls and reverse_bl(k) not in ex_bls)
+    if ants:
+        ants = expand_ants(ants)
+        bls = set(join_bl(i,j) for i,j in split_bls(bls) if i in ants and j in ants)
+    if ex_ants:
+        ex_ants = expand_ants(ex_ants)
+        bls = set(join_bl(i,j) for i,j in split_bls(bls) if i not in ex_ants and j not in ex_ants)
+    bls.union(set(reverse_bl(k) for k in bls)) # put in reverse baselines, just in case
+    reds = [[key for key in gp if key in bls] for gp in reds]
+    return [gp for gp in reds if len(gp) > 1] # XXX do we want to filter off length one reds?
```
jsdillon: I'd say no. The trivial decrease in computation is not worth the extra bookkeeping.

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
 
     return _apply_gains(target, gains, (lambda x, y: x * y), target_type)
 
+class OmnicalSolver(linsolve.LinProductSolver):
+    def __init__(self, data, sol0, wgts={}, gain=.3, **kwargs):
+        """Set up a nonlinear system of equations of the form g_i * g_j.conj() * V_mdl = V_ij
+        to linearize via the Omnical algorithm from Liu et al. 2010.
```
jsdillon: It's not in that paper. Perhaps cite the memo?

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+                Variables with trailing underscores '_' are interpreted as complex conjugates (e.g. x*y_ 
+                parses as x * y.conj()).
+            sol0: Dictionary mapping all variables (as keyword strings) to their starting guess values.
+                This is the point that is Taylor expanded around, so it must be relatively close to the
+                true chi^2 minimizing solution. In the same format as that produced by 
+                linsolve.LogProductSolver.solve() or linsolve.LinProductSolver.solve().
+            wgts: Dictionary that maps equation strings from data to real weights to apply to each 
+                equation. Weights are treated as 1/sigma^2. All equations in the data must have a weight 
+                if wgts is not the default, {}, which means all 1.0s.
+            gain: The fractional step made toward the new solution each iteration.
```
jsdillon: I'd add a comment about what the default is, why it's reasonable, and what range of values are generally safe.

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+            **kwargs: keyword arguments of constants (python variables in keys of data that 
+                are not to be solved for)
+
+        Returns:
```
jsdillon: I generally skip this when there are no returns, but that's your call.

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+            None
+        """
+        linsolve.LinProductSolver.__init__(self, data, sol0, wgts=wgts, **kwargs)
```
jsdillon: Isn't this the circumstance in which you're supposed to use `super`?

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+    def _get_ans0(self, sol, keys=None):
+        '''Evaluate the system of equations given input sol. 
+        Specify keys to evaluate only a subset of the equations.'''
+        if keys is None:
+            keys = self.keys
+        _sol = {k+'_':v.conj() for k,v in sol.items() if k.startswith('g')}
+        _sol.update(sol)
+        return {k: eval(k, _sol) for k in keys}
+
+    def solve_iteratively(self, conv_crit=1e-10, maxiter=50, check_every=4, check_after=1, verbose=False):
```
jsdillon: Isn't `maxiter=50` a bit low for the default? I know it matches `LinProductSolver`, but we generically expect `OmnicalSolver` to take more iterations (even though each iteration is much faster). I'd probably bump this up to 200 or even 500.

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+        Args:
+            conv_crit: A convergence criterion (default 1e-10) below which to stop iterating. 
+                Converegence is measured L2-norm of the change in the solution of all the variables
+                divided by the L2-norm of the solution itself.
+            maxiter: An integer maximum number of iterations to perform before quitting. Default 50.
+            check_every: Compute convergence every Nth iteration (saves computation).  Default 4.
+            check_after: Start computing convergence only after N iterations.  Default 1.
+
+        Returns: meta, sol
+            meta: a dictionary with metadata about the solution, including
```
jsdillon: I'd clarify that each of these has the shape of the data.

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+    def solve_iteratively(self, conv_crit=1e-10, maxiter=50, check_every=4, check_after=1, verbose=False):
+        """Repeatedly solves and updates solution until convergence or maxiter is reached. 
+        Returns a meta object containing the number of iterations, chisq, and convergence criterion.
+
+        Args:
+            conv_crit: A convergence criterion (default 1e-10) below which to stop iterating. 
+                Converegence is measured L2-norm of the change in the solution of all the variables
+                divided by the L2-norm of the solution itself.
+            maxiter: An integer maximum number of iterations to perform before quitting. Default 50.
+            check_every: Compute convergence every Nth iteration (saves computation).  Default 4.
```
jsdillon: You should note that this also controls how often the weights are updated.

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+        dmdl_u = {k:v.flatten() for k,v in dmdl_u.items()}
+        # wgts_u hold the wgts the user provides.  dwgts_u is what is actually used to wgt the data
+        wgts_u = {k: (v * np.ones(chisq.shape, dtype=np.float32)).flatten() for k,v in self.wgts.items()}
+        sol_u = {k:v.flatten() for k,v in sol.items()}
+        iters = np.zeros(chisq.shape, dtype=np.int)
+        conv = np.ones_like(chisq)
+        update = np.where(chisq > 0)
+        for i in range(1,maxiter+1):
+            if verbose: print('Beginning iteration %d/%d' % (i,maxiter))
+            if (i % check_every) == 1:
```
jsdillon: Shouldn't this be `if (i == 1) or (i % check_every) == 1:`? Because otherwise, if `check_every` is 1, it'll never run.

hera_cal/redcal.py
```diff
@@ -269,6 +337,127 @@ def multiply_by_gains(target, gains, target_type='vis'):
...
+            dmdl_u = self._get_ans0(new_sol_u)
+            # check if i % check_every is 0, which is purposely one less than the '1' up at the top of the loop
+            if i < maxiter and (i < check_after or (i % check_every) != 0):
+                # Fast branch when we aren't expensively computing convergence/chisq
+                sol_u = new_sol_u
+            else:
+                # Slow branch when we compute convergence/chisq
+                new_chisq_u = sum([np.abs(v[update]-dmdl_u[k])**2 * wgts_u[k] for k,v in self.data.items()])
+                chisq_u = chisq[update]
+                gotbetter_u = (chisq_u > new_chisq_u)
```
jsdillon: This block with it's variety of similarly named variables (`gotbetter_u`, `where_gotbetter_u`, and `update_where`) could use a comment or two explaining what's going on.

hera_cal/redcal.py
```diff
@@ -297,37 +486,37 @@ def build_eqs(self, bls_in_data):
...
             wgts: dictionary of linear weights in the same format as data. Defaults to equal wgts.
             detrend_phs: takes out average phase, useful for logcal
-            sparse: represent the A matrix (visibilities to parameters) sparsely in linsolve
             **kwargs: other keyword arguments passed into the solver for use by linsolve
 
         Returns:
             solver: instantiated solver with redcal equations and weights
         """
-
+        # XXX ARP: concerned about detrend_phs.  Why is it necessary?
```
jsdillon: I honestly don't know. Where did it come from? If I recall correctly, you put this in here.

hera_cal/redcal.py
```diff
@@ -297,37 +486,37 @@ def build_eqs(self, bls_in_data):
...
+    def _solver(self, solver, data, wgts={}, detrend_phs=False, **kwargs):
         """Instantiates a linsolve solver for performing redcal.
 
         Args:
             solver: linsolve solver (e.g. linsolve.LogProductSolver or linsolve.LinProductSolver)
             data: visibility data in the dictionary format {(ant1,ant2,pol): np.array}
             wgts: dictionary of linear weights in the same format as data. Defaults to equal wgts.
             detrend_phs: takes out average phase, useful for logcal
-            sparse: represent the A matrix (visibilities to parameters) sparsely in linsolve
             **kwargs: other keyword arguments passed into the solver for use by linsolve
```
jsdillon: I see why you removed sparse, but I'd add a note to the kwargs like (e.g. "sparse")

hera_cal/redcal.py
```diff
@@ -357,7 +546,7 @@ def compute_ubls(self, data, gain_sols):
             ubl_sols[blgrp[0]] = np.average(d_gp, axis=0)  # XXX add option for median here?
         return ubl_sols
 
-    def logcal(self, data, sol0={}, wgts={}, sparse=False):
+    def logcal(self, data, sol0={}, wgts={}, sparse=False, mode='default'):
```
jsdillon: Can you explain `mode` in the docstring?

hera_cal/redcal.py
```diff
@@ -376,14 +565,14 @@ def logcal(self, data, sol0={}, wgts={}, sparse=False):
...
-        sol = ls.solve()
+        sol = ls.solve(mode=mode)
         sol = {self.unpack_sol_key(k): sol[k] for k in sol.keys()}
         for ubl_key in [k for k in sol.keys() if len(k) == 3]:
             sol[ubl_key] = sol[ubl_key] * self.phs_avg[ubl_key].conj()
         sol_with_fc = multiply_by_gains(sol, sol0, target_type='gain')
         return sol_with_fc
 
-    def lincal(self, data, sol0, wgts={}, sparse=False, conv_crit=1e-10, maxiter=50):
+    def lincal(self, data, sol0, wgts={}, sparse=False, mode='default', conv_crit=1e-10, maxiter=50, verbose=False):
```
jsdillon: Likewise here with `mode`

hera_cal/redcal.py
```diff
@@ -403,7 +593,32 @@ def lincal(self, data, sol0, wgts={}, sparse=False, conv_crit=1e-10, maxiter=50)
 
         sol0 = {self.pack_sol_key(k): sol0[k] for k in sol0.keys()}
         ls = self._solver(linsolve.LinProductSolver, data, sol0=sol0, wgts=wgts, sparse=sparse)
-        meta, sol = ls.solve_iteratively(conv_crit=conv_crit, maxiter=maxiter)
+        meta, sol = ls.solve_iteratively(conv_crit=conv_crit, maxiter=maxiter, verbose=verbose, mode=mode)
+        sol = {self.unpack_sol_key(k): sol[k] for k in sol.keys()}
+        return meta, sol
+
+    def omnical(self, data, sol0, wgts={}, gain=.3, conv_crit=1e-10, maxiter=50, check_every=4, check_after=1):
```
jsdillon: `gain` is missing from the docstring

hera_cal/tests/profile_logcal.py
```diff
@@ -0,0 +1,282 @@
...
+        for bls in reds:
+            ubl = sol[bls[0]]
+            self.assertEqual(ubl.shape, SHAPE)
+            for bl in bls:
+                d_bl = d[bl]
+                mdl = sol[(bl[0], 'x')] * sol[(bl[1], 'x')].conj() * ubl
+                np.testing.assert_almost_equal(np.abs(d_bl), np.abs(mdl), 5)
+                np.testing.assert_almost_equal(np.angle(d_bl*mdl.conj()), 0, 5)
+
+    def test_omnical(self):
```
jsdillon: Can we move this into test_redcal.py? Or perhaps duplicate it?

hera_cal/redcal.py
AaronParsons: Done.

hera_cal/redcal.py
AaronParsons: Done.

hera_cal/redcal.py
AaronParsons: Done.

hera_cal/redcal.py
AaronParsons: done.

hera_cal/redcal.py
AaronParsons: Done.

hera_cal/redcal.py
AaronParsons: I've never much liked super for it's lack of explicitness in single-inheritance cases.  Can discuss if you want.

hera_cal/redcal.py
AaronParsons: Done.

hera_cal/redcal.py
AaronParsons: Done.

hera_cal/redcal.py
AaronParsons: Done.

hera_cal/redcal.py
AaronParsons: done.

hera_cal/redcal.py
AaronParsons: Done.

AaronParsons:
@jsdillon Make an effort to port datacontainer to the new utils funcs.  Largely successful, but there's a failed test for __contains__ where I don't understand how it worked before, and I want to understand that before I code my own way of making it work now.  Or you could tweak it, if you want.

jsdillon:
@AaronParsons I believe I've finished what you've started. I'm happy with your changes based on my comments. LMK if you have any changes based on my commits from the last hour.

jsdillon:
@AaronParsons thanks! shall I merge, assuming tests pass on travis?

AaronParsons:
Just added tests for RedundantCalibrator.omnical.  I think those tests are sufficient to validate the OmnicalSolver, since it only exists for its implementation in that one function.

AaronParsons:
@jsdillon Yes, assuming tests pass (which they should) let's merge this in!

Issue #339, opened 2018-07-11
Get smooth_cal ready for IDR3 / H2C RTP


jsdillon:
This includes:

* Making it python 2 and python 3 compatible (#318)
* Parallelize over antennas (#338)
* Upgrade to a 2D filter (#298)
* Implement reference antenna selection in smooth_cal (#271)

jsdillon:
#338 remains outstanding, but #298 is done, #318 is done for smooth_cal, and #271 is enabled though we still need to figure out how we want reference antenna selection to happen at a opm/rtp level.

Given all that, I think it's fair to close this meta-issue and just worry about the individual issues.

Issue #298, opened 2018-05-17
Update smooth_cal to do a 2D filter in time and frequency


jsdillon:
Instead of a time filter followed by a frequency filter.

jsdillon:
Addressed in #357.

Issue #271, opened 2018-04-27
Move reference antenna selection / rephasing to smooth cal for RTP for H2C


jsdillon:
Preferably, we wouldn't use a single antenna but some kind of median over antennas.

The idea behind this is that smooth_cal can look at all the files we care about at one point and figure out a good consistent reference for smoothing the one particular file.

jsdillon:
This was enabled by #357, though we have to decide how we want to use it, whether we want to specify reference antennas by hand, and whether we want different reference antennas for each polarization.

jsdillon:
I think this is basically decided... at least until we implement 4pol abscal.

PR #357, opened 2018-08-25
Update smooth_cal for IDR3

jsdillon:
This PR implement filtering of calibration solutions in time and frequency simultaneously using a 2D CLEAN. This addresses #298. This filtering has two modes of operation: "rect", which restricts which Fourier modes can be occupied to a small rectangle around delay and fringe rate of 0. The other mode, "plus", only allows modes that have either delay of 0 or fringe rate of 0 or both. This produces calibrations solutions that are separable in time and frequency.

Additionally, this PR implements a method of selecting and enforcing a reference antenna based on which antenna has the smallest number of flags. This addresses #271.

Finally, we made some small tweaks to get ready for python 3 compatibility (#318).

This PR accomplishes most of #339, though it leaves parallelization over antennas for the future (when array jobs are implemented in hera_opm).


jsdillon:
@adampbeardsley I'm requesting your review since you reviewed the module when I first built it. LMK if you want me to find someone else.

hera_cal/smooth_cal.py
```diff
@@ -105,9 +106,109 @@ def time_filter(gains, wgts, times, filter_scale=1800.0, nMirrors=0):
...
+    fringes = np.fft.fftfreq(times.size, dt)
+    delay_scale = (freq_scale * 1e6)**-1  # Puts it in seconds
+    fringe_scale = (time_scale)**-1  # in Hz
+
+    # find per-integration delays, smooth on the time_scale of gain smoothing, and rephase 
+    taus = fft_dly(gains, df, wgts, medfilt=False, solve_phase=False)[0].astype(np.complex)  # delays are in seconds
+    if not np.all(taus == 0):  # this breaks CLEAN, but it means we don't need smoothing anyway
+        taus = uvtools.dspec.high_pass_fourier_filter(taus.T, np.sum(wgts, axis=1, keepdims=True).T,
+                                                      fringe_scale, dt, tol=tol, maxiter=maxiter)[0].T
+    rephasor = np.exp(-2.0j * np.pi * np.outer(np.abs(taus), freqs))
```
adampbeardsley: Should rephasing be an option, rather than being done automatically? I _think_ the idea is that if delays are off, this will center things so you can smooth around the most significant delay. But if, for example, you're calculating gain corrections, where you don't expect to have significant peaks at center delay, this might do weird things.

hera_cal/smooth_cal.py
```diff
@@ -105,9 +106,109 @@ def time_filter(gains, wgts, times, filter_scale=1800.0, nMirrors=0):
...
+def rephase_to_refant(gains, refant, flags=None):
+    '''Rephase all gains in place so that the phase of the reference antenna is 0 for all times and frequencies.
+
+    Arguments:
+        gains: Dictionary mapping antenna keys to gain waterfalls. Modified in place.
+        refant: Antenna key of antenna to make the reference antenna
+        flags: Optional dictionary mapping antenna keys to flag waterfall.
+            Used only to verify that all gains are flagged where the refant is flagged.
+    '''
+    refant_phasor = gains[refant] / np.abs(gains[refant])
```
adampbeardsley: Is it possible you will be dividing by zero here (e.g. in the case of flagged data)? You may want to avoid this with true_divide so you don't get blasted with warnings.

hera_cal/smooth_cal.py
```diff
@@ -105,9 +106,109 @@ def time_filter(gains, wgts, times, filter_scale=1800.0, nMirrors=0):
...
+            raise ValueError('The chosen reference antenna', refant, 'is flagged in at least one place where antenna',
+                             ant, 'is not, so automatic reference antenna selection has failed.')
+        else:
+            gains[ant] = gains[ant] / refant_phasor
+
+
 class CalibrationSmoother():
 
-    def __init__(self, calfits_list, flags_npz_list=[], antflag_thresh=0.0):
+    def __init__(self, calfits_list, flags_npz_list=[], pick_refant=False, antflag_thresh=0.0):
```
adampbeardsley: Please add `pick_refant` to the docstring.

hera_cal/smooth_cal.py
```diff
@@ -291,25 +424,22 @@ def smooth_cal_argparser():
     a.add_argument("--antflag_thresh", default=0.0, type=float, help="fraction of flagged pixels across all visibilities (with a common antenna) \
                    needed to flag that antenna gain at a particular time and frequency. 0.0 is aggressive flag broadcasting, while 1.0 is \
                    conservative flag broadcasting.")
+    a.add_argument("--pick_refant", default=False, action="store_true", help="TODO: write this")
```
adampbeardsley: TODO

hera_cal/smooth_cal.py
adampbeardsley: Also, should there be an option to manually select the refant?

hera_cal/tests/test_smooth_cal.py
```diff
@@ -68,6 +69,46 @@ def test_freq_filter(self):
...
+        gains *= np.exp(2.0j * np.pi * np.outer(150e-9 * np.ones(10), freqs))
+        ff, info = smooth_cal.time_freq_2D_filter(gains, wgts, freqs, times)
+        np.testing.assert_array_almost_equal(ff, gains)
+
+        # test errors
+        with self.assertRaises(ValueError):
+            ff, info = smooth_cal.time_freq_2D_filter(gains, wgts, freqs, times, filter_mode='blah')
+
+    def test_pick_reference_antenna(self):
+        flags = {ant: np.random.randn(10,10)>0 for ant in [(0, 'Jxx'), (1, 'Jxx')]}
```
adampbeardsley: I highly suggest not using completely random numbers in unit tests. An easy solution is to manually set the seed just before, just so it's reproducible if anything goes wrong.

hera_cal/smooth_cal.py
```diff
@@ -105,9 +106,109 @@ def time_filter(gains, wgts, times, filter_scale=1800.0, nMirrors=0):
...
+    then puts back in the delay rephasor. Uses aipy.deconv.clean to account for weights/flags.
+
+    Arguments:
+        gains: ndarray of shape=(Ntimes,Nfreqs) of complex calibration solutions to filter
+        wgts: ndarray of shape=(Ntimes,Nfreqs) of real linear multiplicative weights
+        freqs: ndarray of frequency channels in Hz
+        times: ndarray of shape=(Ntimes) of Julian dates as floats in units of days
+        freq_scale: frequency scale in MHz to use for the low-pass filter. freq_scale^-1 corresponds
+            to the half-width (i.e. the width of the positive part) of the region in fourier
+            space, symmetric about 0, that is filtered out.
```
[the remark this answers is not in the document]
jsdillon: Correct. Fixed.

hera_cal/smooth_cal.py
```diff
@@ -105,9 +106,109 @@ def time_filter(gains, wgts, times, filter_scale=1800.0, nMirrors=0):
...
+        time_scale: time scale in seconds. Defined analogously to freq_scale.
```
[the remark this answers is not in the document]
jsdillon: Confusingly, times is in days (because that's the units in a UVData object), but time_scale is in seconds. I've added a couple of notes to the documentation to highlight this fact.

hera_cal/smooth_cal.py
```diff
@@ -105,9 +106,109 @@ def time_filter(gains, wgts, times, filter_scale=1800.0, nMirrors=0):
...
+    dt = np.median(np.diff(times)) * 24.0 * 3600.0  # in seconds
+    delays = np.fft.fftfreq(freqs.size, df)
+    fringes = np.fft.fftfreq(times.size, dt)
+    delay_scale = (freq_scale * 1e6)**-1  # Puts it in seconds
+    fringe_scale = (time_scale)**-1  # in Hz
+
+    # find per-integration delays, smooth on the time_scale of gain smoothing, and rephase 
+    taus = fft_dly(gains, df, wgts, medfilt=False, solve_phase=False)[0].astype(np.complex)  # delays are in seconds
+    if not np.all(taus == 0):  # this breaks CLEAN, but it means we don't need smoothing anyway
+        taus = uvtools.dspec.high_pass_fourier_filter(taus.T, np.sum(wgts, axis=1, keepdims=True).T,
```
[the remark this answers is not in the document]
jsdillon: Done.

hera_cal/smooth_cal.py
jsdillon: Sorry, I don't follow. What do you mean by gain corrections? 

If what you're smoothing is basically flat, this will find the delay to be 0 and it will smooth around 0 delay. If the main delay is not 0, we need to smooth around that otherwise you end up getting destructive interference and introducing very strange spectral structure.

hera_cal/smooth_cal.py
jsdillon: AFAIK, we'll be putting 1.0s rather than 0s for flagged gains. But this is a true divide, since we're importing division from __future__

If we get lots of warnings, we can later decide to suppress or deal with them.

hera_cal/smooth_cal.py
jsdillon: Lol, oops. Fixed.

hera_cal/smooth_cal.py
jsdillon: Done. 

For manual refant selection, I'd prefer to punt on this. I'm not sure if we need it and how we implement it depends on whether we want to specify different reference antennas for each polarization or not. If we need it, we can add it pretty straightforwardly.

jsdillon:
@adampbeardsley ok, I think I'm ready for you to take another look.

hera_cal/tests/test_smooth_cal.py
jsdillon: Done.

hera_cal/smooth_cal.py
adampbeardsley: I don't know if I understand the bit about destructive interference, but I also don't understand my original question. I think I was operating with less than enough sleep.

hera_cal/smooth_cal.py
adampbeardsley: Ok, if we don't expect zero gains in a regular mode, then getting warnings is probably good.

hera_cal/smooth_cal.py
jsdillon: haha, ok.

PR #358, opened 2018-08-25
Updates to modules for H2C

nkern:
This PR makes the `abscal` and `lstbin` module compatible with `HERAData` objects

Implementation of partial loading in `abscal` and `lstbin` algorithms and Py3 compatibility are deferred to a future PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
   - conda install -q conda
+  - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   # create environment and install dependencies
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy astropy nose pip matplotlib coverage pandas scikit-learn h5py
   - source activate test-environment
-  - conda install -c conda-forge aipy pycodestyle
+  - conda install -c conda-forge healpy aipy pycodestyle
   - pip install coveralls
   - pip install git+https://github.com/HERA-Team/pyuvdata.git
   - pip install git+https://github.com/HERA-Team/omnical.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 python:
   # We don't actually use the Travis Python, but this keeps it organized.
   - "2.7"
-  - "3.6"
+#  - "3.6"
 
 install:
   # ensure that we have the full tag information available for version.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
   # We don't actually use the Travis Python, but this keeps it organized.
   - "2.7"
+  - "3.6"
+
 install:
   # ensure that we have the full tag information available for version.py
   - git fetch --unshallow --tags
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -71,7 +71,8 @@ class AbsCal(object):
     phs_logcal or a TT_phs_logcal bandpass routine.
     """
     def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
-                 min_bl_cut=None, max_bl_cut=None, bl_taper_fwhm=None, verbose=True):
+                 min_bl_cut=None, max_bl_cut=None, bl_taper_fwhm=None, verbose=True,
+                 filetype='miriad'):
         """
         AbsCal object used to for phasing and scaling visibility data to an absolute reference model.
 
@@ -87,7 +88,7 @@ class AbsCal(object):
                 these must be 2D arrays, with [0] axis indexing time
                 and [1] axis indexing frequency.
 
-                Optionally, model can be a path to a miriad file, a
+                Optionally, model can be a path to a pyuvdata-supported file, a
                 pyuvdata.UVData object or hera_cal.HERAData object,
                 or a list of either.
 
@@ -97,7 +98,7 @@ class AbsCal(object):
                 these must be 2D arrays, with [0] axis indexing time
                 and [1] axis indexing frequency.
 
-                Optionally, data can be a path to a miriad file, a
+                Optionally, data can be a path to a pyuvdata-supported file, a
                 pyuvdata.UVData object or hera_cal.HERAData object,
                 or a list of either. In this case, antpos, freqs
                 and wgts are overwritten from arrays in data.
@@ -136,6 +137,8 @@ class AbsCal(object):
 
         bl_taper_fwhm : float, impose a gaussian taper on the data weights as a function of
             bl separation length, with a specified fwhm [meters]
+
+        filetype : str, if data and/or model are fed as strings, this is their filetype
         """
         # set pols to None
         pols = None
@@ -143,12 +146,12 @@ class AbsCal(object):
         # load model if necessary
         if isinstance(model, list) or isinstance(model, np.ndarray) or isinstance(model, str) or issubclass(model.__class__, UVData):
             (model, model_flags, model_antpos, model_ants, model_freqs, model_lsts,
-             model_times, model_pols) = io.load_vis(model, pop_autos=True, return_meta=True)
+             model_times, model_pols) = io.load_vis(model, pop_autos=True, return_meta=True, filetype=filetype)
 
         # load data if necessary
         if isinstance(data, list) or isinstance(data, np.ndarray) or isinstance(data, str) or issubclass(data.__class__, UVData):
             (data, flags, data_antpos, data_ants, data_freqs, data_lsts,
-             data_times, data_pols) = io.load_vis(data, pop_autos=True, return_meta=True)
+             data_times, data_pols) = io.load_vis(data, pop_autos=True, return_meta=True, filetype=filetype)
             wgts = DataContainer(odict(map(lambda k: (k, (~flags[k]).astype(np.float)), flags.keys())))
             pols = data_pols
             freqs = data_freqs
@@ -1037,7 +1040,7 @@ def omni_abscal_arg_parser():
     return a
 
 
-def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose=True, overwrite=False, write_calfits=True,
+def abscal_run(data_file, model_files, filetype='miriad', refant=None, calfits_infile=None, verbose=True, overwrite=False, write_calfits=True,
                min_bl_cut=None, max_bl_cut=None, bl_taper_fwhm=None, output_calfits_fname=None, return_gains=False, return_object=False, outdir=None,
                match_red_bls=False, tol=1.0, reweight=False, rephase_model=True, all_antenna_gains=False, window=None, edge_cut=0,
                delay_cal=False, avg_phs_cal=False, avg_dly_slope_cal=False, delay_slope_cal=False, phase_slope_cal=False, abs_amp_cal=False,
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -26,7 +28,7 @@ install:
   - source activate test-environment
   - conda install -c conda-forge aipy pycodestyle
   - pip install coveralls
-  - pip install git+https://github.com/HERA-Team/pyuvdata.git@v1.3
+  - pip install git+https://github.com/HERA-Team/pyuvdata.git@9c94be0366eb41692f5501caac99981026a7ffe6
```
jsdillon: Can't we get up to date with master? I'd like to try to always be up to date with master and only build against a specific version or commit when master breaks us.

hera_cal/abscal.py
```diff
@@ -83,34 +81,38 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
 
         Parameters:
         -----------
-        model : visibility data of refence model, type=DataContainer
+        model : Visibility data of refence model, type=dictionary
```
jsdillon: Why this change? Are we never using any of DataContainer's special properties?

hera_cal/abscal.py
```diff
@@ -83,34 +81,38 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
 
-        wgts : weights of data, type=DataContainer, [default=None]
+        wgts : weights of the data, type=dictionary, [default=None]
                keys are antenna pair + pol tuples (must match model), values are real floats
                matching shape of model and data
 
         antpos : type=dictionary, dict of antenna position vectors in ENU (topo) frame in meters.
                  origin of coordinates does not matter, but preferably are centered in the array.
-                 keys are antenna integers and values are 2D ndarray position vectors,
+                 keys are antenna integers and values are 2D position East-North arrays,
```
jsdillon: Can't they also be East-North-Up with Up in position [2]?

hera_cal/abscal.py
```diff
@@ -122,25 +124,28 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
 
         bl_taper_fwhm : float, impose a gaussian taper on the data weights as a function of
             bl separation length, with a specified fwhm [meters]
         """
         # set pols to None
         pols = None
 
-        # check format of model if model is not a dictionary
+        # load model if necessary
         if isinstance(model, list) or isinstance(model, np.ndarray) or isinstance(model, str) or isinstance(model, UVData):
```
jsdillon: You might want to use `issubclass(UVData)` so `HERAData` works too.

hera_cal/abscal.py
```diff
@@ -122,25 +124,28 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
             (model, model_flags, model_antpos, model_ants, model_freqs, model_lsts,
              model_times, model_pols) = io.load_vis(model, pop_autos=True, return_meta=True)
 
-        # check format of data if data is not a dictionary
+        # load data if necessary
         if isinstance(data, list) or isinstance(data, np.ndarray) or isinstance(data, str) or isinstance(data, UVData):
```
jsdillon: And likewise here.

hera_cal/abscal.py
```diff
@@ -83,34 +81,38 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
-                Optionally, model can be a path to a miriad or uvfits file, or a
-                pyuvdata.UVData object, or a list of either.
+                Optionally, model can be a path to a miriad file, a
+                pyuvdata.UVData object or hera_cal.HERAData object,
+                or a list of either.
 
-        data : visibility data of measurements, type=DataContainer
-               keys are antenna pair + pol tuples (must match model), values are
-               complex ndarray visibilities matching shape of model
+        data :  Visibility data, type=dictionary
```
jsdillon: Same comment here.

hera_cal/abscal.py
```diff
@@ -83,34 +81,38 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
+                pyuvdata.UVData object or hera_cal.HERAData object,
+                or a list of either. In this case, antpos, freqs
+                and wgts are overwritten from arrays in data.
 
         refant : antenna number integer for reference antenna
             The refence antenna is used in the phase solvers, where an absolute phase is applied to all
             antennas such that the refant's phase is set to identically zero.
 
-        wgts : weights of data, type=DataContainer, [default=None]
+        wgts : weights of the data, type=dictionary, [default=None]
```
jsdillon: And here.

hera_cal/abscal.py
```diff
@@ -1457,15 +1459,22 @@ def cut_bls(datacontainer, bls, bl_cut):
...
-    Output (cut_datacontainers)
+    Output
     ------
     cut_datacontainer : DataContainer object with bl cut enacted
     """
+    if min_bl_cut is None:
+        min_bl_cut = 0.0
+    if max_bl_cut is None:
+        max_bl_cut = 1e10
     cut_datacontainer = odict()
```
jsdillon: Because data containers now carry around some metadata, I think this is a bad solution. Instead, I'd make a deepcopy and use `del`.

hera_cal/abscal.py
```diff
@@ -1446,9 +1448,9 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         return return_obj
 
 
-def cut_bls(datacontainer, bls, bl_cut):
+def cut_bls(datacontainer, bls, min_bl_cut=None, max_bl_cut=None):
```
jsdillon: It seems to me like `bls` should be an optional parameter, defaulted to None, an `if bls is None: bls = datacontainer.keys()`

hera_cal/abscal_funcs.py
```diff
@@ -1676,6 +1667,7 @@ def avg_file_across_red_bls(data_fname, outdir=None, output_fname=None,
 
     # write to file
     if write_miriad:
+        raise NotImplementedError("Correct averaging of integration_time and nsample not yet implemented")
```
jsdillon: This is a bit strange. You definitely want to comment out the code that isn't usable. But I wonder: why aren't we using `io.write_vis`? Can't we just eliminate this whole block?

hera_cal/abscal.py
nkern: I thought you wanted to deprecate `DataContainer`, so I thought it would just be apt to use a dictionary, which I think is actually the only thing we need, but I can add `or DataContainer` if you planned to keep it around...

hera_cal/abscal_funcs.py
nkern: I punted on writing b/c we need to implement the proper summing of the `nsample` and `integration_time` arrays which is not trivial and I didn't want to invest time into it now. I can comment out that code if you like

hera_cal/abscal.py
jsdillon: No, the whole point of `HERAData` is that it spits out DataContainers. We're definitely keeping them.

hera_cal/abscal_funcs.py
jsdillon: Yeah, I'd comment out the code. Otherwise our coverage will go way down.

Alternatively, if no one is using the code, we could also delete it.

hera_cal/abscal.py
jsdillon: Also, there may be other places where you turn dicts/odicts into DataContainers. We should avoid that if possible.

jsdillon:
@nkern I'd really like to get the pyuvdata update done before everything else. Can we narrow the scope of this PR and do the rest in a different PR?

hera_cal/abscal.py
nkern: bls is not a list of keys, its a dictionary of baseline vectors

hera_cal/abscal.py
```diff
@@ -83,28 +81,32 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
+        data :  Visibility data, type=dictionary or DataContainer
+                keys are antenna-pair + polarization tuples, Ex. (1, 2, 'xx').
+                values are complex ndarray visibilities.
+                these must be 2D arrays, with [0] axis indexing time
+                and [1] axis indexing frequency.
 
-                Optionally, model can be a path to a miriad or uvfits file, or a
-                pyuvdata.UVData object, or a list of either. In this case, wgts, antpos,
-                freqs, times and pols are overwritten with equivalent information from data object.
+                Optionally, data can be a path to a miriad file, a
```
jsdillon: Do we want to support uvh5 rather than miriad? Or just leave this as a legacy interface?

hera_cal/abscal.py
```diff
@@ -1446,28 +1450,42 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
...
+    max_bl_cut : float, maximum baseline separation [meters] to keep in data    
 
-    bl_cut : float, maximum baseline separation [meters] to keep in data
+    inplace : bool, if True edit data in input object, else make a copy.
 
-    Output (cut_datacontainers)
+    Output
     ------
-    cut_datacontainer : DataContainer object with bl cut enacted
+    datacontainer : DataContainer object with bl cut enacted
```
jsdillon: I think it's generally better to keep the returns of the same form whenever possible... so in this case, I'd recommend returning the cut datacontainer regardless. If people want to use inplace, it's OK to have a return and just call it without any assignment operator.

hera_cal/abscal.py
```diff
@@ -1446,28 +1450,42 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
...
+        only returned if not inplace
     """
-    cut_datacontainer = odict()
+    if inplace:
```
jsdillon: Shouldn't this be `if not inplace`?

hera_cal/abscal.py
```diff
@@ -1446,28 +1450,42 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
         return return_obj
 
 
-def cut_bls(datacontainer, bls, bl_cut):
+def cut_bls(datacontainer, bls, min_bl_cut=None, max_bl_cut=None, inplace=False):
```
jsdillon: Data containers (if they come from a file) now contain `antpos` inside them. Do we want to remove the dependence on `bls`?

hera_cal/abscal.py
```diff
@@ -1446,28 +1450,42 @@ def abscal_run(data_file, model_files, refant=None, calfits_infile=None, verbose
...
+        if k not in bls:
+            continue
+        if bl_len > max_bl_cut or bl_len < min_bl_cut:
+            del datacontainer[k]
 
-    assert len(cut_datacontainer) > 0, "no baselines were kept after baseline cut..."
+    assert len(datacontainer) > 0, "no baselines were kept after baseline cut..."
 
-    return DataContainer(cut_datacontainer)
+    if not inplace:
```
jsdillon: See above comment.

hera_cal/io.py
```diff
@@ -700,10 +697,11 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
...
     telescope_location : type=ndarray, telescope location in xyz in ITRF (earth-centered frame).
 
-    integration_time : type=float or ndarray, integration duration in seconds for data_array. This does not necessarily have
-        to be equal to the diff(time_array): for the case of LST-binning, this is not the duration of the LST-bin
-        but the integration time of the pre-binned data. If None, will use diff(time_array) by default.
-        
+    integration_time : type=float or ndarray, integration duration in seconds for data_array.
+        This does not necessarily have to be equal to the diff(time_array): for the case of
+        LST-binning, this is not the duration of the LST-bin but the integration time of the
+        pre-binned data. Default is median(diff(time_array)) in seconds.
```
jsdillon: I would clarify here that the total integration time is always `integration_time * nsamples`.

hera_cal/abscal.py
nkern: I added a filetype kwarg to support both

hera_cal/abscal.py
nkern: yup, good catch thanks

hera_cal/abscal.py
nkern: sure thing

hera_cal/io.py
nkern: sure thing

hera_cal/abscal.py
jsdillon: Where? I don't see it.

hera_cal/abscal.py
jsdillon: No?

nkern: ok

PR #368, opened 2018-09-24
fixed io.write_cal for 2pol case and added test

nkern:
quick bug fix to `io.write_cal` in case when `gains` contains multiple polarizations. Old code was extending `ants` instead of `jones` axis in this case. Tests are added.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -1007,7 +1007,7 @@ def write_cal(fname, gains, freqs, times, flags=None, quality=None, total_qual=N
     '''
 
     # get antenna info
-    ant_array = np.array(sorted(map(lambda k: k[0], gains.keys())), np.int)
+    ant_array = np.unique(map(lambda k: k[0], gains.keys())).astype(np.int)
     antenna_numbers = copy.copy(ant_array)
     antenna_names = np.array(map(lambda a: "ant{}".format(a), antenna_numbers))
     Nants_data = len(ant_array)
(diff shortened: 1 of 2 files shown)


hera_cal/io.py
```diff
@@ -1007,7 +1007,7 @@ def write_cal(fname, gains, freqs, times, flags=None, quality=None, total_qual=N
     '''
 
     # get antenna info
-    ant_array = np.array(sorted(map(lambda k: k[0], gains.keys())), np.int)
+    ant_array = np.unique(map(lambda k: k[0], gains.keys())).astype(np.int)
```
jsdillon: `np.unique([int(k[0]) for k in gains])` is much more readable and succinct in my mind, but it's a matter of taste.

hera_cal/tests/test_io.py
```diff
@@ -707,13 +706,21 @@ def test_write_cal(self):
...
+        gains2 = odict(map(lambda k: (k, gains[k][:1]), gains.keys()))
+        uvc = io.write_cal("ex.calfits", gains2, freqs, times[:1], return_uvc=True, outdir='./')
         self.assertAlmostEqual(uvc.integration_time, 0.0)
         self.assertEqual(uvc.Ntimes, 1)
         self.assertTrue(os.path.exists('ex.calfits'))
         os.remove('ex.calfits')
 
+        # test multiple pol
+        for k in gains.keys():
+            gains[(k[0], 'y')] = gains[k].conj()
```
jsdillon: Shouldn't this be 'Jyy'?

nkern: In the test it was (1, 'x') so I just mirrored it to y. I don't think it matters in the context of this test, but yes in principle one could say it should it Jyy

PR #370, opened 2018-10-05
DO NOT MERGE: Omni run

jsdillon:


hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+    nInt_to_load integrations at a time and skipping and flagging times when the sun is above solar_horizon.
+    
+    Arguments:
+        hd: HERAData object, instantiated with the datafile or files to calibrate. Must be loaded using uvh5.
+        nInt_to_load: number of integrations to load and calibrate simultaneously. Lower numbers save memory, 
+            but incur a CPU overhead.
+        pol_mode: polarization mode of redundancies. Can be '1pol', '2pol', '4pol', or '4pol_minV'.
+            See recal.get_reds for more information.
+        ex_ants: list of antennas to exclude from calibration and flag. Can be either antenna numbers or
+            antenna-polarization tuples. In the former case, all pols for an antenna will be excluded.
```
AaronParsons: Perhaps expose other arguments used in filter_reds, just for full flexibility?

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+    # get reds and then intitialize omnical visibility solutions to all 1s and all flagged
+    all_reds = get_reds({ant: hd.antpos[ant] for ant in ant_nums}, pol_mode=pol_mode,
+                        pols=set([pol for pols in pol_load_list for pol in pols]))
+    all_reds = filter_reds(all_reds, ex_ants=ex_ants)
+    rv['v_omnical'] = {red[0]: np.ones((nTimes, nFreqs), dtype=np.complex64) for red in all_reds}
+    rv['vf_omnical'] = {red[0]: np.ones((nTimes, nFreqs), dtype=bool) for red in all_reds}
+
+    # solar flagging
+    lat, lon, alt = hd.telescope_location_lat_lon_alt_degrees
+    solar_alts = utils.get_sun_alt(hd.times, latitude=lat, longitude=lon)
```
AaronParsons: Is hd.times compatible with baseline-dependent time cadences?

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+    solar_flagged = solar_alts > solar_horizon
+    if verbose and np.any(solar_flagged):
+        print(len(hd.times[solar_flagged]), 'integrations flagged due to sun above', solar_horizon, 'degrees.')
+    
+    # loop over data, performing partial loading 
+    for pols in pol_load_list:
+        if verbose:
+            print('Now calibrating', pols, 'polarization(s)...')
+        reds = filter_reds(all_reds, ex_ants=ex_ants, pols=pols)
```
AaronParsons: As per above, consider exposing other filter_reds args?

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+                    rv['chisq_per_ant'][ant][tinds, :] = cal['chisq_per_ant'][ant]
+                for bl in cal['v_omnical'].keys():
+                    rv['v_omnical'][bl][tinds, :] = cal['v_omnical'][bl]
+                    rv['vf_omnical'][bl][tinds, :] = cal['vf_omnical'][bl]
+                if pol_mode in ['1pol', '2pol']:
+                    for antpol in cal['chisq'].keys():
+                        rv['chisq'][antpol][tinds, :] = cal['chisq'][antpol]
+                else:  # duplicate chi^2 into both antenna polarizations
+                    for antpol in rv['chisq'].keys():
+                        rv['chisq'][antpol][tinds, :] = cal['chisq']
```
AaronParsons: Have we aggregated all of the metadata we want out of cal?  What about iterations, etc.?  Might help debugging or knowing if you're fully converged.

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+
+    return rv
+
+
+def redcal_run(input_data, firstcal_suffix='.first.calfits', omnical_suffix='.omni.calfits', omnivis_suffix='.omni.vis', 
+               outdir=None, ant_metrics_file=None, clobber=False, nInt_to_load=8, pol_mode='2pol', ex_ants=[], ant_z_thresh=4.0, 
+               solar_horizon=0.0, conv_crit=1e-10, maxiter=500, check_every=10, check_after=50, gain=.4,
+               append_to_history='', verbose=False):
+    '''Perform redundant calibration (firstcal, logcal, and omnical) an entire HERAData object, loading only 
+    nInt_to_load integrations at a time and skipping and flagging times when the sun is above solar_horizon.
```
AaronParsons: Emphasize in comments that this does iterative antenna flagging.  We probably want to be able to run this function without partial data loading on existing data as well.  Can we make nInt_to_load default to -1 (all) and if so, uses the redundantly_calibrate instead of redcal_partial_io_iteration?

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+        check_after: start computing omnical convergence only after N iterations (saves computation).
+        gain: The fractional step made toward the new solution each omnical iteration. Values in the
+            range 0.1 to 0.5 are generally safe. Increasing values trade speed for stability.
+        append_to_history: string to add to history of output firstcal and omnical files
+        verbose: print calibration progress updates
+
+    Returns:
+        cal: the dictionary result of the final run of redcal_partial_io_iteration (see above for details)
+    '''
+    assert isinstance(input_data, str), 'input_data must be a single string path to a uvh5 visibility data file.'
```
AaronParsons: Why not, optionally, and hd object as well, in case you already have it open.

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+    hd = HERAData(input_data)
+    ex_ants = set(ex_ants)
+    if ant_metrics_file is not None:
+        for ant in load_metric_file(ant_metrics_file)['xants']:
+            ex_ants.add(ant)
+
+    # loop over calibration, removing bad antennas and re-running if necessary
+    while True:
```
AaronParsons: Should we specify a max antenna flagging iterations to cap the total compute time we can spend in this loop?

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+        print('Now saving omnical visibilities to', input_data + omnivis_suffix)
+    hd.read(bls=cal['v_omnical'].keys())
+    hd.update(data=cal['v_omnical'], flags=cal['vf_omnical'])
+    hd.history += append_to_history
+    hd.write_uvh5(os.path.join(outdir, os.path.basename(input_data) + omnivis_suffix), clobber=True)
+
+    return cal
+
+
+def redcal_argparser():
```
AaronParsons: We are bordering on a level where we might want to move to config files to control parameter passing to these scripts.

hera_cal/tests/test_redcal.py
```diff
@@ -27,6 +31,54 @@ def build_hex_array(hexNum, sep=14.7):
             i += 1
     return antpos
 
+def build_split_hex_array_with_outriggers(sep = 14.6, hexNum = 11, splitCore = True, splitCoreOutriggers = 4):
```
AaronParsons: Would like this stuff moved to hera_sim

hera_cal/omni.py
```diff
@@ -386,25 +386,20 @@ def remove_degen(info, g, v, g0, minV=False):
...
-    # Taking polarization non-aware stuff from omnical and reextracting the relevant info for remove_degen
-    info_antpos = info.get_antpos()
-    antpos = dict(zip([ant[0] for ant in ants],
-                      [np.append(info_antpos[ant[0], 0:2], [0]) for ant in ants]))
-
     # Put sols into properly formatted dictionaries and remove degeneracies
-    sol = {(i, antpol): g[antpol][i] for (i, antpol) in ants}
-    sol.update({(i, j, pol): v[pol][(i, j)] for (i, j, pol) in bl_pairs})
-    sol0 = {(i, antpol): g0[antpol][i] for (i, antpol) in ants}
+    sol = {(i % info.nant, antpol): g[antpol][i] for (i, antpol) in ants}
```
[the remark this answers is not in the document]
jsdillon: Yes. I just made the changes to get the unit tests to pass. They'll become irrelevant in #372.

hera_cal/redcal.py
```diff
@@ -2,13 +2,19 @@
...
+from hera_cal.io import HERAData, HERACal, write_cal, write_vis
 from hera_cal.apply_cal import calibrate_in_place
-import six
+from hera_qm.ant_metrics import per_antenna_modified_z_scores
+from hera_qm.metrics_io import load_metric_file
+import argparse
+import os
 
 
 def noise(size):
```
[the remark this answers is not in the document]
jsdillon: I agree in principle, but I'm wary of making hera_sim a dependency for hera_cal when there isn't really anyone directly responsible for maintaining it. For example, right now hera_sim doesn't have tests for sim_red_data. Also, the hera_cal sim_red_data and and the hera_sim one have diverged a bit. This one uses DataContainers and handles polarization strings properly. If you want to make an issue, go ahead.

hera_cal/redcal.py
```diff
@@ -92,13 +99,10 @@ def get_pos_reds(antpos, bl_error_tol=1.0, low_hi=False):
...
-
-    if low_hi:
-        red_groups = [[bl[::-1] for bl in r] if r[0][0] > r[0][1] else r for r in red_groups]
-
-    return red_groups
+    return [sorted(reds[delta]) if sorted(reds[delta])[0][0] == np.min(reds[delta])
+            else sorted([reverse_bl(bl) for bl in reds[delta]]) for delta in orderedDeltas]
 
 
 def add_pol_reds(reds, pols=['xx'], pol_mode='1pol', ex_ants=[]):
```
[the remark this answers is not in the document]
jsdillon: Done in 5aeb6d3ab97db22d6a587bb2d8eacdc9d088a6d3

hera_cal/redcal.py
```diff
@@ -136,7 +140,7 @@ def excluded(bl, pol):
     return redsWithPols
 
 
-def get_reds(antpos, pols=['xx'], pol_mode='1pol', ex_ants=[], bl_error_tol=1.0, low_hi=False):
```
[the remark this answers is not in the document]
jsdillon: Done in 5aeb6d3ab97db22d6a587bb2d8eacdc9d088a6d3

hera_cal/redcal.py
```diff
@@ -436,15 +482,15 @@ def __init__(self, reds):
         self.reds = reds
         self.pol_mode = parse_pol_mode(self.reds)
 
-    def build_eqs(self, bls_in_data):
-        """Function for generating linsolve equation strings. Takes in a list of baselines that
-        occur in the data in the (ant1, ant2, pol) format and returns a dictionary that maps
-        linsolve string to (ant1, ant2, pol) for all visibilities."""
+    def build_eqs(self, dc):
```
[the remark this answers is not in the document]
jsdillon: I agree. I added a keyword error in 7c0a1cccc71d63b6a5dee1336259dc2b0ba6a251 and also made the DataContainer an optional keyword parameter.

hera_cal/redcal.py
```diff
@@ -464,15 +510,15 @@ def _solver(self, solver, data, wgts={}, detrend_phs=False, **kwargs):
             solver: instantiated solver with redcal equations and weights
         """
         # XXX ARP: concerned about detrend_phs.  Why is it necessary?
```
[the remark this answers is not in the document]
jsdillon: Yes, logcal still uses it. If you want to make it an issue that's fine, but I'm inclined to keep it for now.

hera_cal/redcal.py
```diff
@@ -464,15 +510,15 @@ def _solver(self, solver, data, wgts={}, detrend_phs=False, **kwargs):
...
-        dtype = data.values()[0].dtype
+        dtype = data.values()[0].dtype  # TODO: fix this for python 3
         dc = DataContainer(data)
-        eqs = self.build_eqs(dc.keys())
+        eqs = self.build_eqs(dc)
         self.phs_avg = {}  # detrend phases within redundant group, used for logcal to avoid phase wraps
         if detrend_phs:
             for blgrp in self.reds:
-                self.phs_avg[blgrp[0]] = np.exp(-np.complex64(1j) * np.median(np.unwrap([np.log(dc[bl]).imag for bl in blgrp], axis=0), axis=0))
+                self.phs_avg[blgrp[0]] = np.exp(-1j * np.median(np.unwrap([np.log(dc[bl]).imag for bl in blgrp], axis=0), axis=0))
```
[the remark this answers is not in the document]
jsdillon: Fixed.

hera_cal/redcal.py
```diff
@@ -644,13 +693,12 @@ def omnical(self, data, sol0, wgts={}, gain=.3, conv_crit=1e-10, maxiter=50, che
         sol = {self.unpack_sol_key(k): sol[k] for k in sol.keys()}
         return meta, sol
 
-    def remove_degen_gains(self, antpos, gains, degen_gains=None, mode='phase'):
```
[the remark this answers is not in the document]
jsdillon: haha

hera_cal/redcal.py
```diff
@@ -464,15 +510,15 @@ def _solver(self, solver, data, wgts={}, detrend_phs=False, **kwargs):
...
-        eqs = self.build_eqs(dc.keys())
+        eqs = self.build_eqs(dc)
         self.phs_avg = {}  # detrend phases within redundant group, used for logcal to avoid phase wraps
         if detrend_phs:
             for blgrp in self.reds:
-                self.phs_avg[blgrp[0]] = np.exp(-np.complex64(1j) * np.median(np.unwrap([np.log(dc[bl]).imag for bl in blgrp], axis=0), axis=0))
+                self.phs_avg[blgrp[0]] = np.exp(-1j * np.median(np.unwrap([np.log(dc[bl]).imag for bl in blgrp], axis=0), axis=0))
                 for bl in blgrp:
-                    self.phs_avg[bl] = self.phs_avg[blgrp[0]]
+                    self.phs_avg[bl] = self.phs_avg[blgrp[0]].astype(dc[bl].dtype)
```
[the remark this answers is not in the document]
jsdillon: Doesn't hurt though.

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
     """
 
     return count_redcal_degeneracies(antpos, bl_error_tol=bl_error_tol) == 4
+
+
+def get_pol_load_list(pols, pol_mode='1pol'):
```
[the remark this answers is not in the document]
jsdillon: I added a couple in 59b41f6a467d7e19c892d3d7c24244ea6c531ff7

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+    
+    Arguments:
+        data: dictionary or DataContainer mapping baseline-pol tuples like (0,1,'xx') to 
+            complex data of shape 
+        reds: list of lists of redundant baseline tuples, e.g. (0,1,'xx'). The first
+            item in each list will be treated as the key for the unique baseline.
+        freqs: 1D numpy array frequencies in Hz. Optional if inferable from data DataContainer,
+            but must be provided if data is a dictionary, if it doesn't have .freqs, or if the
+            length of data.freqs is 1.
+        times_by_bl: dictionary mapping antenna pairs like (0,1) to times in days. Optional if
```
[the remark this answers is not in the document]
jsdillon: Yes. I clarified the doc string.

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+    make_sol_finite(log_sol)
+    rv['omni_meta'], omni_sol = rc.omnical(data, log_sol, conv_crit=conv_crit, maxiter=maxiter, 
+                                           check_every=check_every, check_after=check_after, gain=gain)
+
+    # update omnical flags and then remove degeneracies
+    rv['g_omnical'], rv['v_omnical'] = get_gains_and_vis_from_sol(omni_sol)
+    rv['gf_omnical'] = {ant: ~np.isfinite(g) for ant, g in rv['g_omnical'].items()}
+    rv['vf_omnical'] = {bl: ~np.isfinite(v) for bl, v in rv['v_omnical'].items()}
+    make_sol_finite(omni_sol)
+    rd_sol = rc.remove_degen(omni_sol, degen_sol=rv['g_firstcal'])
```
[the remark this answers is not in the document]
jsdillon: We do a little bit, but this is a problem that exists in the current pipeline that abscal already fixes. I don't think we need to add extra functionality here.

hera_cal/redcal.py
jsdillon: I'd like to revisit this question as part of a bigger upgrade I've had in mind that also enables calibrating off a subset of the baselines. This is a bit more work, because we probably also want to then use those calibration solutions to come up with model visibilities for all unique baselines, even if they weren't part of the redundant calibration. This can be done by averaging, but I'd like to leave it out for now. Maybe you want to make an issue?

hera_cal/redcal.py
jsdillon: Good question. No, I don't think so. This seems like an issue for later, since it's not totally clear how we'd handle BDA data in omnical. Do we upsample? Downsample? I'm not sure. Maybe make an issue?

hera_cal/redcal.py
jsdillon: See above. I'd like to punt on this.

hera_cal/redcal.py
jsdillon: It's a bit tricky. We'd need more code for building containers to hold other metadata and there isn't an obvious way to store it in calfits. I think we're OK for now, since this reproduces the omnical functionality we currently have, but maybe we should make this an issue?

hera_cal/redcal.py
jsdillon: We made a decision a while ago that we want to break backwards compatibility for non-uvh5 data. This is all based on new code in hera_cal.io that operates differently (and better) on uvh5 files. Paul already converted all of IDR2 data to uvh5 and we can do it with the rest of IDR3 when we're ready for the big run.

Letting redcal_run do an end-run around redcal_partial_io_iteration and just use redundantly_calibrate is not easy at all and I'm not sure what the use case is.

I'm happy to change the default nint_to_load, but I figured it would be more helpful to have it be a reasonable value.

jsdillon: Also, I forgot to update that docstring. I've fixed that.

hera_cal/redcal.py
jsdillon: This mucks with how we do all the file suffixes. I think we want to re-think how they are handled anyway, so perhaps you can make an issue and we can put on this for now?

hera_cal/redcal.py
jsdillon: Done.

hera_cal/redcal.py
jsdillon: If we do that, it should be repo-wide. And in practice, using hera_opm, we ultimately do that. For now, I think we're fine.

hera_cal/tests/test_redcal.py
jsdillon: See my above comment.

hera_cal/redcal.py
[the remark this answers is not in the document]
AaronParsons: Can we replace the "XXX" comment with a statement that logcal uses this code to avoid the pi/-pi wrapping point in phase?

hera_cal/redcal.py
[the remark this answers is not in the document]
AaronParsons: But it's sloppy and implies something is happening that shouldn't be....

hera_cal/redcal.py
```diff
@@ -789,3 +837,335 @@ def is_redundantly_calibratable(antpos, bl_error_tol=1.0):
...
+    rv['g_omnical'], rv['v_omnical'] = get_gains_and_vis_from_sol(rd_sol)
+    rv['g_omnical'] = {ant: g * ~rv['gf_omnical'][ant] + rv['gf_omnical'][ant] for ant, g in rv['g_omnical'].items()}
+
+    # compute chisqs
+    data_wgts = {bl: utils.predict_noise_variance_from_autos(bl, data, 
+                 dt=(np.median(np.ediff1d(times_by_bl[bl[:2]])) * 86400.))**-1 for bl in data.keys()}
```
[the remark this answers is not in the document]
AaronParsons: Is times_by_bl backwards compatible with our current data sets?  If so, I'll remove my objection.  If we have to do a repackaging of our current data sets to construct a times_by_bl, then my object stands.

86400 should be stored as a constant somewhere at the top (maybe even pulled from astropy or something) and referenced by name here.  No magic numbers in the middle of code.

hera_cal/redcal.py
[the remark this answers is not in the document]
AaronParsons: Could we provide a mechanism for providing an alternate degen sol, in case we want to push abscal into here someday?

hera_cal/redcal.py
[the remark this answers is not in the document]
jsdillon: Done.

hera_cal/redcal.py
AaronParsons: Let's just expose parameters.  We can put off work on solving for unlisted baselines as a future issue.

hera_cal/redcal.py
jsdillon: Ok, nevermind. This is done now.

hera_cal/redcal.py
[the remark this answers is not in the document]
jsdillon: Yes, times_by_bl is now backwards compatible. I've put in some extra code to make sure that all the appropriate metadata is calculated and saved to the HERAData object you pass into redcal_iteration, regardless of input data format.

And I also moved the 86400 up to the top, though I still find this quite silly.

hera_cal/redcal.py
jsdillon: Ok, I've exposed them all in one big bundle as **filter_reds_kwargs

hera_cal/redcal.py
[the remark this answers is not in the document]
jsdillon: I don't think we know how we'd want to tie abscal in here in the future, but I think we should wait until we know that before we make this edit. If we're bring in abscal, are we still even running firstcal? Maybe we want a way to bypass firstcal? I don't know.

I think this will be easier to do later and risks adding bloat to our functions, docstrings, and unit tests now. To me this falls under the umbrella overly general code that might be harder to maintain later.

hera_cal/redcal.py
AaronParsons: Moved to issue #378

hera_cal/tests/test_redcal.py
AaronParsons: Moved to Issue #379.

hera_cal/redcal.py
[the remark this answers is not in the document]
AaronParsons: Moved to Issue #379.

hera_cal/redcal.py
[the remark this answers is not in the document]
AaronParsons: Moved to Issue #380

hera_cal/redcal.py
AaronParsons: Moved to Issue #381.

hera_cal/redcal.py
AaronParsons: Moved to issue #382.

hera_cal/redcal.py
```diff
@@ -504,7 +506,6 @@ def _solver(self, solver, data, wgts={}, detrend_phs=False, **kwargs):
         Returns:
             solver: instantiated solver with redcal equations and weights
         """
-        # XXX ARP: concerned about detrend_phs.  Why is it necessary?
```
AaronParsons: But add explanation of why we use detrend_phs.  It's to center the phase bias around 0 to get maximal distance from the +/- pi wrap points.

hera_cal/redcal.py
```diff
@@ -1007,17 +1015,26 @@ def redcal_partial_io_iteration(hd, nInt_to_load=8, pol_mode='2pol', ex_ants=[],
...
+                                   np.arange(nInt_to_load, len(hd.times[~solar_flagged]), nInt_to_load))
+        else:
+            tind_groups = [np.arange(nTimes)[~solar_flagged]]  # just load a single group 
+        for tinds in tind_groups:
             if len(tinds) > 0:
                 if verbose:
                     print('    Now calibrating times', hd.times[tinds[0]], 'through', hd.times[tinds[-1]], '...')
-                data, _, _ = hd.read(times=hd.times[tinds], polarizations=pols)
+                if nInt_to_load == -1:  # don't perform partial I/O
+                    data, _, _ = hd.build_datacontainers()  # this may contain unused polarizations, but that's OK
```
AaronParsons: Can build_containers not work for for nInt_to_load = the whole file?

hera_cal/redcal.py
```diff
@@ -1007,17 +1015,26 @@ def redcal_partial_io_iteration(hd, nInt_to_load=8, pol_mode='2pol', ex_ants=[],
...
-    # loop over data, performing partial loading 
+    # loop over polarizations and times, performing partial loading if desired
     for pols in pol_load_list:
         if verbose:
             print('Now calibrating', pols, 'polarization(s)...')
         reds = filter_reds(all_reds, ex_ants=ex_ants, pols=pols)
-        for tinds in np.split(np.arange(nTimes)[~solar_flagged], 
-                              np.arange(nInt_to_load, len(hd.times[~solar_flagged]), nInt_to_load)):
+        if nInt_to_load > 0:  # split up the integrations to load nInt_to_load at a time
+            tind_groups = np.split(np.arange(nTimes)[~solar_flagged],
```
AaronParsons: Can split handle a case where nInt_to_load is the whole file, in order to avoid if/else?

PR #390, opened 2018-11-02
added utils.broadcast_flags for factorizing flags in time and freq

nkern:
broadcasts flags along time if a threshold is met, otherwise flags the integration. This will be useful for reflection modeling analyses etc..


jsdillon:
OK, I'm good to merge.

PR #391, opened 2018-11-03
Make repo python3 compatible

plaplant:
This PR is the much-anticipated python3-ification of the repo. In addition to the usual python3 "gotchas" of `print` being a function and strings behaving differently, many of the changes were around handling functions which now return generators/iterators instead of lists (like `map`, `dict.keys()`, etc.). Care needs to be taken to explicitly cast these function outputs to lists if, e.g., addressing them by index, or testing for membership of an item. @nkern @jsdillon these will be some of the biggest changes to keep in mind going forward, but hopefully our tests will catch those instances when they crop up.

As a part of bringing this up to spec, there was a circular dependency that had to be resolved, and some general housekeeping. Most of the big changes:
* The `echo` function was moved from `abscal_funcs.py` to `utils.py`.
* A new module `flag_utils.py` was created, and the `solar_flag` and `synthesize_ant_flags` functions were moved from `utils.py` to this new module.
* The `FRFilter` object was moved out of the top-level namespace, so now users can no longer do, e.g., `from hera_cal import FRFilter`. Instead, they can import from the `frf.py` module, as in `from hera_cal.frf import FRFilter`. (If this behavior is desired, it can be added back in, though it seemed odd that it was the _only_ class available from the top-level namespace.)

Some outstanding issues that should either be resolved as part of this PR, or added as issues to the repo:
* Many of the tests produce warnings, which is sloppy and makes it hard to distinguish genuine unexpected behavior from known issues. At the very least, the tests should be wrapped so that the expected warnings are handled (using, e.g., `pyuvdata.tests.checkWarning`), but ideally these errors (which are largely numerical floating point errors) should not be generated in the first place.
* Some tests (such as `abscal_tests.py:Test_AbsCal.test_abs_amp_logcal`) do not actually test the output values produced for certain cases, and instead just call functions and silently assume the output is good. While this is valid for testing that the code runs without producing an error, it does _not_ ensure the code is producing the expected output, which somewhat defeats the purpose of a unit test.

I'm happy to chip in on these issues, but I feel addressing them should be spread out among multiple contributors, because there is some significant work to be done.


nkern:
@plaplant I'm also not as concerned about the warnings because they still produce expected behavior: i.e. they go to infs or nans that we catch later. We can't actually make them nan's before the division operation due to limitations within `linsolve` and other dependencies. I feel that "bad / unexpected behavior" is generally supposed to be captured by the unit tests (not by warnings), so the warnings--while sometimes unsettling to see when running the code--are less concerning to me.

plaplant:
@nkern I'm referring to this call in `test_abscal.py`: https://github.com/HERA-Team/hera_cal/blob/d1b38c/hera_cal/tests/test_abscal.py#L316. (It's not he only instance, just the first I happened to find.) The `abs_amp_logcal` function is called, but nothing is explicitly done with the output before the end of the test. If this branch of the code is in fact being verified in other tests, that's great, but then the call should be removed from this test, because nothing is being explicitly checked with this output.

As far as warnings go, I agree that sometimes the fault lies with upstream repos. For instance, in the current tests, there are some warnings that are coming from scipy functions using numpy in deprecated ways, and we have no direct control over those issues. Nevertheless, and especially for floating point exceptions, I do believe that the input can in principle (and should) be handled in such a way that the warnings do not come up in the first place. These warnings absolutely _should_ be concerning, because they can mask actual errors that we might not be aware of (which actually happened when running RTP last year!), because we've become inoculated to pushing ahead in the face of warnings instead of taking a step back and being more careful.

Regardless, if preventing the warnings in the first place is too onerous, at the very least our tests should not pass the warnings through to the output. If they are predictable, then we've already developed infrastructure for handling them. I've started to wrap a few tests in `test_abscal.py` and `test_utils.py`, but there are many other locations where we need to catch the warnings.

nkern:
Thanks for all the work @plaplant ! 🎉

jsdillon:
Yeah, thanks so much!

Issue #354, opened 2018-08-20
Update code for QM's XRFI revamp


adampbeardsley:
A lot of interfaces are changing with QM's XRFI revamp, so I'm preemptively filing an issue here to address it. For testing purposes, we can use QM's `xrfi_revamp` branch before it gets merged into master.
Things to update:
`scripts/omni_xrfi.py`
`scripts/apply_cal.py` (currently uses a flag npz, which will be a UVFlag h5 file in the future)
`scripts/smooth_cal_run.py` (flag npz)
`hera_cal/apply_cal.py`
`hera_cal/io.py` (load_npz_flags should be replaced, or add a load_uvflag function)

There is an issue in QM that the last item should be moved to QM, but given that load_vis and load_cal are in `hera_cal.io`, I'm not so convinced.

adampbeardsley:
I just realized right after merging the xrfi revamp branch that this issue is suddenly much more urgent. We will probably need to build against hera_qm@f3e32c3cef36cbdf784d4028c5a1ef14c082a274 until we do this stuff.

jsdillon:
Resolved by #398

PR #398, opened 2018-11-16
Update flag I/O to use both new UVFlag objects (via .h5 files) and legacy .npz files

### LINKED ISSUE #374: Enable support for UVFlag objects based on h5 files in smooth_cal

jsdillon:
Also updated smooth_cal and apply_cal to conform to the new flag I/O.

This resolves #374 and #354 (which I think are basically duplicates).

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -559,7 +559,7 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
             UVFlag object in the 'flag' mode (could be by baseline, by antenna, or by 
             polarization). 'npz' provides legacy support for the IDR2.1 flagging npzs,
             but only for per-visibility flags.
-        return_meta: if True, return a metadata dictionary with, e.g., 'times' and 'freqs'
+        return_meta: if True, return a metadata dictionary with, e.g., 'times', 'freqs', 'history'
         
     Returns:
         flags: dictionary or DataContainer mapping keys to Ntimes x Nfreqs numpy arrays.
@@ -578,6 +578,7 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
         assert uvf.mode == 'flag', 'Must be in flag mode.'
         freqs = np.unique(uvf.freq_array)
         times = np.unique(uvf.time_array)
+        history = uvf.history
         pol_indices = {polnum2str(polnum): i for i, polnum in enumerate(uvf.polarization_array)}
         if uvf.type == 'baseline':  # one time x freq waterfall per baseline
             blt_slices = get_blt_slices(uvf)
@@ -598,6 +599,7 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
         pols = [polnum2str(p) for p in npz['polarization_array']]
         freqs = np.unique(npz['freq_array'])
         times = np.unique(npz['time_array'])
+        history = npz['history']
         nAntpairs = len(npz['antpairs'])
         assert npz['flag_array'].shape[0] == nAntpairs * len(times), \
             'flag_array must have flags for all baselines for all times.'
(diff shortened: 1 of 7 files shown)


hera_cal/io.py
```diff
@@ -128,6 +128,28 @@ def update(self, gains=None, flags=None, quals=None, total_qual=None):
                 self.total_quality_array[0, :, :, ip] = total_qual[pol].T
 
 
+def get_blt_slices(uvo):
+    '''For a pyuvdata-style UV object, get the mapping from antenna pair to blt slice.
+    
+    Arguments:
+        uvo: a "UV-Object" like UVData or baseline-type UVFlag
```
adampbeardsley: Just a suggestion, but it might be useful to put in a check that the input adheres to this. Otherwise it might be pretty difficult to diagnose issues.

hera_cal/io.py
```diff
@@ -537,6 +548,73 @@ def iterate_over_times(self, Nints=1, times=None):
...
+        meta: (only returned if return_meta is True)
+    '''
+    flags = {}
+    if filetype not in ['h5', 'npz']:
+        raise ValueError("filetype must be 'h5' or 'npz'.")
+    
+    elif filetype == 'h5':
+        from hera_qm import UVFlag
+        uvf = UVFlag(flagfile)
+        assert uvf.mode == 'flag', 'Must be in flag mode.'
```
adampbeardsley: To help debugging, you may want this assert message to be a little more detailed.

hera_cal/io.py
```diff
@@ -537,6 +548,73 @@ def iterate_over_times(self, Nints=1, times=None):
...
+        freqs = np.unique(uvf.freq_array)
+        times = np.unique(uvf.time_array)
+        history = uvf.history
+        pol_indices = {polnum2str(polnum): i for i, polnum in enumerate(uvf.polarization_array)}
```
adampbeardsley: Strictly speaking, UVFlag objects can have strings in their polarization array. For example, if you average over pol, the resulting polarization array could be `['-5,-6']`. We can avoid it by not collapsing the pols in our actual workflow, but you might want to have some sort of assert or if here.

hera_cal/io.py
```diff
@@ -537,6 +548,73 @@ def iterate_over_times(self, Nints=1, times=None):
...
+                for (ant1, ant2), blt_slice in blt_slices.items():
+                    flags[(ant1, ant2, polnum2str(polnum))] = uvf.flag_array[blt_slice, 0, :, ip] 
+            flags = DataContainer(flags)
+        elif uvf.type == 'antenna':  # one time x freq waterfall per antenna
+            for i, ant in enumerate(uvf.ant_array):
+                for ip, jnum in enumerate(uvf.polarization_array):
+                    flags[(ant, jnum2str(jnum))] = np.array(uvf.flag_array[i, 0, :, :, ip].T)
+        elif uvf.type == 'waterfall':  # one time x freq waterfall per visibility polarization
+            for ip, jnum in enumerate(uvf.polarization_array):
+                flags[jnum2str(jnum)] = uvf.flag_array[:, :, ip]
```
adampbeardsley: This is where the polarization_array is likely to break things.

hera_cal/io.py
jsdillon: I'm not sure how to do that without unnecessarily restricting what can go into the function (who knows what objects we'll build in the future that could use this method). It's probably a "for experts" method, so I'd consider making it `_get_blt_slices`, but otherwise I think it's OK.

hera_cal/io.py

hera_cal/io.py

hera_cal/io.py

jsdillon:
@adampbeardsley I believe I'd addressed your concerns. Wanna take a second look?

PR #399, opened 2018-11-17
Adds xtalk modeling code, and makes FRF and ReflectionFitter Partial IO Capable

nkern:
This PR
1) adds in crosstalk modeling and removal code
2) updates the `ReflectionFitter` object to inherit from the new `FRFilter` and `VisClean` objects
3) Refines minor parts of the cable reflection algorithm
4) Small updates to `io.py` to accommodate capabilities in `ReflectionFitter`


nkern:
The xtalk modeling code uses a PCA-based approach and operates on the delay-transformed data.

Here is an example on real LSTBinned IDR2 data

__Re(Visbility) delay transform across the entire band__
83_84_dfft
Non fringing structure corrupting a wide swatch of delays outside foreground wedge is highly apparent.


__SVD (or PCA) decomposition of waterfall__
83_84_pca
(UL) Eigenspectrum showing eigenvalues of each PC. (UR) Zeroth eigen-mode across frequency. (LL) Zeroth eigen-mode across time. (LR) Outer product of the two to give the zeroth Principal Component.
The SVD is done for all 500 < |delays| < 2000, so the foreground wedge is ignored in the PC decomposition.

__Zeroth PC removes the non-fringing structure (by-eye)__
83_84_removal
(Top) Original data. (Middle) Zeroth principal component of SVD from above. (Bottom) Residual. What's left on the left-hand side at -1300 ns is the (fringing) cable reflection! Which you couldn't see before, and wasn't picked out by the PC b/c the cable reflection is a fringing structure, whereas the PC is a non-fringing template.

__Smoothing the time eigen-mode w/ a GPR over RFI gaps__
83_84_pca_smoothed
Same as before, but I've smoothed the time eigen-mode (LL) with a GPR, thus allowing me to get rid of some the noise in the eigen-mode, and interpolate over RFI gaps (or skip-wgts).

nkern:
The next important question in order for this technique to be viable, is does it also remove EoR signal at the delays over which it operates? I should have an answer to that in the next couple of days, but I don't think that should prohibit us from merging this code in...

hera_cal/io.py
```diff
@@ -215,14 +222,16 @@ def __init__(self, input_data, filetype='uvh5'):
...
+            self.read(read_data=False, run_check=run_check, check_extra=check_extra,
+                      run_check_acceptability=run_check_acceptability)
             self.filepaths = temp_paths
 
             if len(self.filepaths) > 1:  # save HERAData_metas in dicts
                 for meta in self.HERAData_metas:
                     setattr(self, meta, {})
                 for path in self.filepaths:
-                    hc = HERAData(path, filetype='uvh5')
+                    hc = HERAData(path, filetype='uvh5', run_check=run_check, check_extra=check_extra,
```
jsdillon: for consistency, this should by `hd`. Probably my mistake.

hera_cal/io.py
```diff
@@ -183,14 +183,21 @@ class HERAData(UVData):
     HERAData_metas = ['ants', 'antpos', 'freqs', 'times', 'lsts', 'pols',
                       'antpairs', 'bls', 'times_by_bl', 'lsts_by_bl']
 
-    def __init__(self, input_data, filetype='uvh5'):
+    def __init__(self, input_data, filetype='uvh5', run_check=True,
```
jsdillon: Why do we need these parameters? Also, if they are extremely rarely used and only by experts, perhaps we want to just collapse them into `**uvh5_read_kwargs` to avoid cluttering up the documentation?

hera_cal/io.py
```diff
@@ -361,8 +370,9 @@ def build_datacontainers(self):
 
         return data, flags, nsamples
 
-    def read(self, bls=None, polarizations=None, times=None,
-             frequencies=None, freq_chans=None, read_data=True, return_data=True):
+    def read(self, bls=None, polarizations=None, times=None, frequencies=None,
+             freq_chans=None, read_data=True, return_data=True, run_check=True,
+             check_extra=True, run_check_acceptability=True):
```
jsdillon: Same question as above, though this function is already pretty complicated so adding three more parameters is less of an issue.

hera_cal/io.py
```diff
@@ -740,6 +762,8 @@ def to_HERAData(input_data, filetype='miriad'):
     if isinstance(input_data, str):  # single visibility data path
         return HERAData(input_data, filetype=filetype)
     elif isinstance(input_data, HERAData):  # already a HERAData object
+        input_data._determine_blt_slicing()
```
jsdillon: Why is this necessary? I think you might have explained it to me, but I honestly don't recall a scenario where a HERA data object would have been created and where this could have been run but wasn't.

hera_cal/io.py
```diff
@@ -1207,8 +1230,9 @@ def write_cal(fname, gains, freqs, times, flags=None, quality=None, total_qual=N
     Nants_telescope = len(antenna_numbers)
 
     # get polarization info
-    pol_array = np.array(sorted(set(map(lambda k: k[1], gains.keys()))))
-    jones_array = np.array(list(map(lambda p: jstr2num(p), pol_array)), np.int)
+    jones_array = np.array(list(set([jstr2num(k[1]) for k in gains.keys()])))
+    jones_array = jones_array[np.argsort(np.abs(jones_array))]
```
jsdillon: Why? Is this the standard ordering in pyuvdata? Maybe add a comment explaining.

hera_cal/frf.py
```diff
@@ -26,22 +28,21 @@ def timeavg_waterfall(data, Navg, flags=None, nsamples=None, rephase=False, lsts
     weighted by a boolean flag array (flags) and also optionally by an Nsample array (nsample),
     such that, for a single frequency channel, the time average is constructed as
 
-    avg_data = sum( data * flag_wgt * nsample ) / sum( flag_wgt * nsample )
+    avg_data = sum( data * flag_wgt * nsample**2 ) / sum( flag_wgt * nsample**2 )
```
jsdillon: Why? If the noise goes like 1/sqrt(nSamples), then inverse variance weighting would be proportional to nsamples, not nsamples**2.

hera_cal/frf.py
```diff
@@ -26,22 +28,21 @@ def timeavg_waterfall(data, Navg, flags=None, nsamples=None, rephase=False, lsts
...
-    one can also construct the averaged nsample for each averaging window.
+    the averaged nsample for each averaging window is computed and returned.
 
     Parameters
     ----------
     data : ndarray
         2D complex ndarray of complex visibility with shape=(Ntimes, Nfreqs)
-        The rows of data are assumed to be ordered chronologically, in either
-        asending or descending order.
+        The rows of data are assumed to be ordered chronologically.
```
jsdillon: Is equal spacing in time assumed?

hera_cal/frf.py
```diff
@@ -172,8 +169,8 @@ def timeavg_waterfall(data, Navg, flags=None, nsamples=None, rephase=False, lsts
             dlst = mean_l - lst
             d = utils.lst_rephase(d, bl_vec, freqs, dlst, lat=lat, inplace=False, array=True)
 
-        # form data weights : flag weights * nsample
-        w = fw * n
+        # form data weights : flag weights * nsample**2
```
jsdillon: Same question as above.

hera_cal/frf.py
```diff
@@ -220,6 +216,26 @@ def timeavg_data(self, t_avg, rephase=False, verbose=True):
         ----------
         t_avg : float
             Width of time-averaging window in seconds.
+
+        rephase : bool
+            If True, rephase data in averaging window to window-center.
+
+        data : DataContainer
```
jsdillon: It seems like the only way to instantiate a VisClean object is with data. What's the point giving this optional data?

If you're looking for more flexibility, I think the right thing to do is to make a separate function that performs time averaging on datacontainers and then have this function call that one. Otherwise you have to instantiate a weird empty visclean object and then feed it external data.

hera_cal/frf.py
```diff
@@ -230,21 +246,28 @@ def timeavg_data(self, t_avg, rephase=False, verbose=True):
...
         # setup lists
-        avg_data = odict()
-        avg_flags = odict()
-        avg_nsamples = odict()
+        avg_data = DataContainer({})
+        avg_flags = DataContainer({})
+        avg_nsamples = DataContainer({})
+
+        # setup averaging quantities
+        if data is None:
```
jsdillon: Again, same comment as before. This feels like a weird hijacking of the object rather than a natural use of the underlying functions.

hera_cal/frf.py
```diff
@@ -338,6 +361,8 @@ def write_data(self, outfilename, write_avg=True, filetype='miriad', add_to_hist
         # write data
         if filetype == 'miriad':
             new_hd.write_miriad(outfilename, clobber=True)
+        elif filetype == 'uvh5':
```
jsdillon: I'd also make the default uvh5.

hera_cal/frf.py
```diff
@@ -197,16 +194,15 @@ def timeavg_waterfall(data, Navg, flags=None, nsamples=None, rephase=False, lsts
     # wrap lsts
     avg_lsts = avg_lsts % (2 * np.pi)
 
-    return dict(avg_data=avg_data, win_flags=win_flags, avg_nsamples=avg_nsamples,
-                avg_lsts=avg_lsts, avg_extra_arrays=avg_extra_arrays)
+    return avg_data, win_flags, avg_nsamples, avg_lsts, avg_extra_arrays
 
 
 class FRFilter(VisClean):
```
jsdillon: I have a more general question. What's the point of making FRF a subclass of any VisClean? Is it so that you can time axis cleaning? Maybe I'm confused because I don't know what a FRF run function will look like.

hera_cal/io.py
nkern: Ah, this was because we are accessing the `UVData.read_uvh5*` function which has these kwargs so we should really expose them. One reason is that `UVData.read_uvh5` does a check on LST values by recomputing them entirely, which for certain file shapes (e.g. long LST ranges) is actually quite intensive (i.e. ~13 seconds per file!). I agree we should use a compact `**kwargs` structure though

hera_cal/frf.py
nkern: ah that's right...

hera_cal/frf.py
nkern: i don't think this is assumed, but also I don't think there is a scenario in our normal day-to-day where that wouldn't be the case..

hera_cal/frf.py
nkern: Oh the point is that you can have multiple kinds of `data` attached to the `VisClean` object, for example, there is the `self.data` which is the data loaded in, but after cleaning you also have `clean_model` and `clean_resid`, and you may want to time average those as well. If you had `VisClean.timeavg_data` hardcoded to only time-avg `self.data`, but wanted to timeavg `self.clean_model`, then you'd have to do some silly gymnastics to copy it over to `self.data`, while copy that over to `self._data` in meantime to preserve it... etc etc you get the picture. The default is `self.data`, but its nice to have it flexible for other scenarios. And adding another function to timeavg DataContainers doesn't solve that issue, I think, and frankly in my mind, isn't worth creating another entire function for, b/c its just a for loop..

hera_cal/frf.py
nkern: I don't think its going against the philosophy of the object. The point is to be able to build upon the objects inherent capabilities: `VisClean` can take `self.data` and from it create `self.clean_model` and `self.clean_resid` and `self.clean_data`. One can choose to `timeavg` any one of those by simply feeding them into the function as `timeavg(data=VC.clean_resid)`, for example. Now one thing you may be thinking is, well we could do `data=['data', 'model', 'resid'...]` as a predefined list of choices, which it then knows to pick out `self.data`, `self.model`, `self.resid` etc respectively, but that seems a little unnecessarily rigid. I guess what I can say is, I have already found it beneficial to work with the object in this way in my own analyses, rather than having it hardcoded.

hera_cal/frf.py
nkern: Yeah, I guess the point is so that you can do some cleaning on the data before time averaging it, if you like, perhaps for example to fill-in entire integrations that were flagged, which would throw power for a true FR filter (rather than a FR clean as `VIsClean` is doing).

hera_cal/frf.py
jsdillon: I understand now and am fine with these optional parameters for this function.

However, I still think having a separate function that operates on DataContainers is the right thing to do. As it stands (and correct me if I'm wrong), if you have a DataContainer without an associated file or HERAData/UVData object, there's no way to use this function because you have to instantiate a VisClean object to use it. 

It would be much more modular and more useful interactively if we had a stand-alone function that worked on DataContainers. The idea behind this function is that it uses all the metadata from the HERAData object correctly to perform the time averaging... and then does the for loop. It's the automatic handling of metadata that's the key value-add here. 

But if you want to supply the metadata and DataContainers yourself, you should be able to. That's the philosophy that Aaron has been pushing for modularity of analysis and I've been trying to follow. See, e.g. redcal where we have `redundantly_calibrate` which operates on DataContainers, `redcal_iteration` which wraps it and operates on HERAData objects, and `redcal_run` which wraps that and operates on files.

hera_cal/frf.py
jsdillon: I'm still not totally convinced, but I'm ok with you closing this thread if you address the one below: https://github.com/HERA-Team/hera_cal/pull/399/files#r247688297

hera_cal/reflections.py
```diff
@@ -51,169 +72,80 @@
...
+from .frf import FRFilter
+from . import vis_clean
+from .utils import echo
 
 
-class ReflectionFitter(object):
-    def __init__(self, data=None, **load_kwargs):
+class ReflectionFitter(FRFilter):
+    """
+    A subclass of FRFilter with added reflection modeling capabilities.
```
jsdillon: Do you want to mention that FRFilter is itself a subclass of VisClean? Perhaps also mention how to instantiate a ReflectionFitter object?

hera_cal/reflections.py
```diff
@@ -51,169 +72,80 @@
...
-                          alpha=0.1, zero_pad=0, overwrite=False, fthin=10, verbose=True):
-        """
-        Model reflections in (ideally RFI-free) visibility data. To CLEAN data of
-        flags see the self.dly_clean_data() function. Recommended to set zero_pad
-        to at least as large as Nfreqs.
+        To CLEAN data of RFI gaps see the self.vis_clean() function.
+        Recommended to set zeropad to at least as large as Nfreqs/2.
 
         Args:
+            data : DataContainer, data to find reflections in.
```
jsdillon: Do you want to make this an optional parameter and use self.data unless otherwise overwritten like you did with FRFilter?

hera_cal/reflections.py
```diff
@@ -51,169 +72,80 @@
...
         if keys is None:
-            keys = [k for k in data.keys() if k[0] == k[1]]
+            keys = [k for k in data.keys() if k[0] == k[1] and k[2][0] == k[2][1]]
 
-        # get window kwargs
-        win_kwargs = {}
-        if taper == 'tukey':
-            win_kwargs['alpha'] = alpha
+        # Take FFT of data
+        self.fft_data(data, flags=flags, keys=keys, assign='dfft', ax='freq', window=window, alpha=alpha,
```
jsdillon: In general, I don't really get the philosophy here. Why assign the results to `self.dfft` rather than returning them and operating on them locally? Do we need them to stick around in the namespace and the memory?

Also, you could further reduce the memory footprint by doing the fft_data inside the following `for k in keys` loop and then letting each dfft[k] go out of scope after each iteration of the loop.

hera_cal/reflections.py
```diff
@@ -225,34 +157,63 @@ def model_reflections(self, dly_range, keys=None, data='clean', edgecut=0, taper
...
-            self.peak_ratio[rkey] = sig
-            self.dfft[k] = dfft
-
-    def write_reflections(self, output_calfits, input_calfits=None, overwrite=False, add_to_history=''):
+
+            # fit for reflection delays and amplitude
+            amp, dly, inds, sig = fit_reflection_delay(self.dfft[k], dly_range, self.delays, return_peak=False)
+
+            # make significance cut
+            if np.max(sig) < ref_sig_cut:
```
jsdillon: Wouldn't it be more sensible to simply write the amplitude, delay, and phase as 0? I get the desire to not have to solve for the phase if that's the slow part, but I think it's weird to ask it to find reflections for some number of antennas and for the code to silently only solve for some antennas.

That way you can record ref_significance rather than losing it and also fill out ref_eps.

hera_cal/reflections.py
```diff
@@ -225,34 +157,63 @@ def model_reflections(self, dly_range, keys=None, data='clean', edgecut=0, taper
...
+                k = (a, uvutils.parse_jpolstr(p)) 
+                if k not in self.ref_gains:
+                    self.ref_gains[k] = np.ones((self.Ntimes, self.Nfreqs), dtype=np.complex)
+
+    def write_auto_reflections(self, output_calfits, input_calfits=None, overwrite=False,
+                               write_npz=False, add_to_history=''):
         """
+        Write auto reflection gain terms from self.ref_gains.
+
         Given a filepath to antenna gain calfits file, load the
```
jsdillon: This should clarify that if input_calfits is None, the output is only reflections (especially this this the mode we plan on using this for in IDR2.2)

hera_cal/reflections.py
```diff
@@ -225,34 +157,63 @@ def model_reflections(self, dly_range, keys=None, data='clean', edgecut=0, taper
...
         Returns:
-            uvc : HERACal object with new gains
+            uvc : UVCal object with new gains
         """
         # Create reflection gains
-        rgains = _form_gains(self.epsilon)
+        rgains = self.ref_gains
         flags, quals, tquals = None, None, None
 
         if input_calfits is not None:
```
jsdillon: Something is weird here. Why would they not match? It seems to me like if the number of times and/or the number of freqs doesn't match, it's an error.

hera_cal/reflections.py
jsdillon: Also, clarify where the flags/quals come from in both the case where you have an input_calfits and the case where you don't.

hera_cal/reflections.py
```diff
@@ -267,25 +228,439 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
         else:
-            time_array = self.reflection_times
+            time_array = self.times
             freq_array = self.freqs
             kwargs = {}
 
-        uvc = io.write_cal(output_calfits, rgains, freq_array, time_array, flags=flags, quality=quals, 
-                           total_qual=tquals, outdir=os.path.dirname(output_calfits), zero_check=False, 
-                           overwrite=overwrite, history=version.history_string(add_to_history), **kwargs)
+        echo("...writing {}".format(output_calfits))
```
jsdillon: Shouldn't we be passing in a verbose option? Or do you really want this to always print?

hera_cal/reflections.py
```diff
@@ -267,25 +228,439 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+        uvc = io.write_cal(output_calfits, rgains, freq_array, time_array, flags=flags,
+                           quality=quals, total_qual=tquals, zero_check=False,
+                           overwrite=overwrite, history=version.history_string(add_to_history),
+                           **kwargs)
+
+        if write_npz:
+            output_npz = os.path.splitext(output_calfits)[0] + '.npz'
+            if not os.path.exists(output_npz) or overwrite:
+                echo("...writing {}".format(output_npz))
```
jsdillon: Likewise.

hera_cal/reflections.py
```diff
@@ -267,25 +228,439 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
             if cal.Ntimes > self.Ntimes:
```
jsdillon: Something is weird here. Why would they not match? It seems to me like if the number of times and/or the number of freqs doesn't match, it's an error.

hera_cal/reflections.py
```diff
@@ -225,34 +157,63 @@ def model_reflections(self, dly_range, keys=None, data='clean', edgecut=0, taper
...
+            self.ref_eps[rkey] = eps
+            self.ref_amp[rkey] = amp
+            self.ref_phs[rkey] = phs
+            self.ref_dly[rkey] = dly
+            self.ref_significance[rkey] = sig
+            self.ref_flags[rkey] = np.min(flags[k], axis=1, keepdims=True)
+
+        # form gains
+        self.ref_gains = form_gains(self.ref_eps)
+        for a in self.ants:
```
jsdillon: I understand why you need this loop, but I find it strange. If you specify keys that don't include all autocorrelations, then the code assumes that those antennas have no reflection and reports that. It's a non-intuitive cascade of assumptions. I get why you need that, but perhaps you want to clarify in the documentation what is assumed if you don't use `keys=None`.

hera_cal/reflections.py
```diff
@@ -51,169 +72,80 @@
...
             overwrite : bool, if True, overwrite dictionaries
+
+        Result:
+            self.ref_eps : dict, reflection epsilon term
+            self.ref_amp : dict, reflection amplitude
+            self.ref_phs : dict, reflection phase [radians]
+            self.ref_dly : dict, reflection delay [nanosec]
+            self.ref_flags : dict, reflection flags [bool]
+            self.ref_significance : dict, reflection significance,
+                which is max|Vfft| / median|Vfft| near reflection peak.
```
jsdillon: Also, and most importantly, self.ref_gains.

hera_cal/reflections.py
```diff
@@ -267,25 +228,439 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+                np.savez(output_npz, delay=self.ref_dly, phase=self.ref_phs, amp=self.ref_amp,
+                         significance=self.ref_significance, times=self.times, freqs=self.freqs,
+                         lsts=self.lsts, antpos=self.antpos,
+                         history=version.history_string(add_to_history))
+
         return uvc
 
+    def pca_decomp(self, dly_range, dfft=None, flags=None, side='both', keys=None, overwrite=False, verbose=True):
+        """
+        Create a PCA-based model of the FFT data in dfft.
```
jsdillon: More detailed info needed here. What kind of PCA? What are the vectors? Give some context for what dly_range means.

hera_cal/io.py
nkern: Yeah I agree: I think that's also just a bug w/ selecting on HERAData to begin with that this would fix..

hera_cal/reflections.py
nkern: yeah good point

hera_cal/reflections.py
```diff
@@ -51,169 +72,80 @@
...
+            data : DataContainer, data to find reflections in.
             dly_range : len-2 tuple of delay range [nanoseconds] to search
                 within for reflections. Must be either both positive or both negative.
+            flags : DataContainer, flags of data. Default is None.
             keys : list of len-3 tuples, keys in data model reflections over. Default
                 is to use all auto-correlation keys.
-            data : str, data dictionary to find reflections in. Options are ['clean', 'data'].
-            edgecut : int, number of channels on band edges to exclude from modeling
-            taper : str, tapering function to apply across freq before FFT
+            edgecut_low : int, Nbins to flag but not window at low-side of the FFT axis.
```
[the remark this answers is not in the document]
nkern: how about
`            edgecut_low : int, Nbins to flag and exclude from windowing function on low-side of the band
`

hera_cal/reflections.py
```diff
@@ -51,169 +72,80 @@
...
-        # get keys: only use auto correlations to model reflections
+        self.ref_eps = {}
+        self.ref_amp = {}
+        self.ref_phs = {}
+        self.ref_dly = {}
+        self.ref_flags = {}
+        self.ref_significance = {}
+
+        # get flags
+        if flags is None:
```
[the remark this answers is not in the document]
nkern: Because the data is supposed to be CLEANed, and `self.flags` has all of the non-factorized flags in them. The `flags` kwarg should be the CLEAN flags. I'll update the argument names to be more clear about this!

hera_cal/reflections.py
nkern: In practice I've been wanting to use `dfft` for analysis (like when I make plots to compare before and after), and `self.fft_data` has some convenience to it over the `fft_data` function. The memory footprint concern is well-taken, but b/c this is running on autocorrs only I don't think its going to be an issue for us.

hera_cal/reflections.py
nkern: yeah that's a good idea

hera_cal/reflections.py
```diff
@@ -51,169 +72,80 @@
...
-            self.clean_times = np.mean(self.clean_times, keepdims=True)
-
-    def model_reflections(self, dly_range, keys=None, data='clean', edgecut=0, taper='none',
-                          alpha=0.1, zero_pad=0, overwrite=False, fthin=10, verbose=True):
-        """
-        Model reflections in (ideally RFI-free) visibility data. To CLEAN data of
-        flags see the self.dly_clean_data() function. Recommended to set zero_pad
-        to at least as large as Nfreqs.
+        To CLEAN data of RFI gaps see the self.vis_clean() function.
+        Recommended to set zeropad to at least as large as Nfreqs/2.
```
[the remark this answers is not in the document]
nkern: Yeah, the `zeropad` issue is a little finnicky: I get mixed results depending on the data I'm working on. I wanted to make it clear to the user (i.e. in the first couple lines of doc-string b/c I'm *always* assuming that even a novice user has read the doc-string) that one should experiment with a large zeropad, but I didn't want to make it the default b/c doing so changes the output dimensions of `dfft`, which I think is a little weird to do if not explicitly requested.

hera_cal/reflections.py
nkern: I'll update the doc string to make it more clear in general

hera_cal/reflections.py
nkern: There are cases where a calibration is a single integration that we want to broadcast across all times, or a single channel that we want to broadcast across all freqs. In the case where the broadcasting isn't allowed, it will error in `io.write_cal`.

hera_cal/reflections.py
nkern: Actually this was done so that when we write out gains, all antennas have a gain. Otherwise, a `calfits` without an antenna in `apply_cal` will flag the baseline when applied, which we don't want: we simply don't have a reflection for that antenna, which is fine. I'll clarify in the docs though

hera_cal/reflections.py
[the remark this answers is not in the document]
jsdillon: This is a tricky case. Maybe just add another note to the description of `zeropad` so it's harder to miss.

hera_cal/reflections.py
jsdillon: Is that something we ever want to do? Or do we want to force the user to do the broadcasting rather than do it silently?

hera_cal/reflections.py
```diff
@@ -51,169 +72,84 @@
...
+            window : str, tapering function to apply across freq before FFT. See dspec.gen_window for options
             alpha : float, if taper is Tukey, this is its alpha parameter
-            zero_pad : int, number of channels to pad band edges with zeros before FFT
+            zeropad : int, number of channels to pad band edges with zeros before FFT
             fthin : int, scaling factor to down-select frequency axis when solving for phase
+            ref_sig_cut : float, if max reflection significance is not above this, do not record solution
             overwrite : bool, if True, overwrite dictionaries
+
+        Result:
+            self.ref_eps : dict, reflection epsilon term with (ant, pol) keys and ndarray values
```
jsdillon: I'd also include the shape of the values. It's 1D and length nTimes, right?

hera_cal/reflections.py
```diff
@@ -442,85 +832,194 @@ def fit_reflection(data, dly_range, freqs, full_freqs=None, edgecut=0,
...
-    # construct reflection
-    if full_freqs is None:
-        full_freqs = freqs
-    eps = construct_reflection(full_freqs, ref_amps, ref_dlys, ref_phs)
+    return ref_phs
 
-    return (eps, ref_amps, ref_dlys, ref_phs, ref_dly_inds, ref_significance,
-            np.fft.fftshift(dfft, axes=1), np.fft.fftshift(dlys))
 
+def _svd_waterfall(data):
```
jsdillon: This function isn't necessary. If you want to clarify the shapes of the PCs and svals, just add a comment to the one place where you use it.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+                continue
+
+            # perform svd to get principal components
+            d = np.zeros_like(dfft[k])
+            d[:, select] = dfft[k][:, select]
+            u, s, v = _svd_waterfall(d)
+            Nmodes = len(s)
+
+            # append to containers
+            self.umodes[k] = u[:, :Nmodes]
```
jsdillon: I'd recommend putting a comment here about why you're only keeping Nmodes.

jsdillon: Also, you might want to consider calling these tmodes and dmodes... because fundamentally they are delay vs. time.

hera_cal/reflections.py
```diff
@@ -38,9 +38,30 @@
...
+where we see that the V_22 = v_2 v_2^* autocorrelation is now
+present in the V_12 cross correlation. V_22 is a non-fringing
+term but time-dependent term, meaning it will introduce an
+offset in the visibility at a specific delay that will have
+some time-dependent amplitude. This cannot be corrected
+via antenna-based calibration. However, we can in principle
+subtract this off at the visibility level by taking
+some form of a time-average and subtracting off the bias.
+The code here models the time and delay dependent behavior
+through a combination of SVD and fringe-rate filtering.
```
jsdillon: Given the very large number of things you potentially attach to this object over the course of your various reflection fitting, I'd consider doing two things:

1) Add a list of all the things you attach to the object to this docstring, what they are, and explain which functions generate them.
2) Add an `__init__` that uses `super` but also sets all these things to `None`

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+        self.form_PCs(keys, overwrite=overwrite)
+
+        # append relevant metadata
+        if hasattr(dfft, 'times'):
+            self.umodes.times = dfft.times
+            self.vmodes.times = dfft.times
+            self.svals.times = dfft.times
+            self.uflags.times = dfft.times
+
+    def form_PCs(self, keys=None, u=None, v=None, overwrite=False, verbose=True):
```
jsdillon: I don't understand the utility of this function. When would you want to use some other u and v than the one you found with pca_decomp but then store the result to self.pcomps. 

I'd cut most of it and fold the rest into self.pca_decomp or _form_PCs as necessary.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+            dly_range : len-2 tuple of positive delays in nanosec
+            dfft : DataContainer, holding delay-transformed data. Default is self.dfft
+            flags : DataContainer, holding dfft flags (e.g. skip_wgts). Default is None.
+            side : str, options=['pos', 'neg', 'both']
+                Specifies dly_range as positive delays, negative delays or both.
+            keys : list of tuples
+                List of datacontainer baseline-pol tuples to create model for.
+            overwrite : bool
+                If dfft exists, overwrite its values.
+        """
```
jsdillon: There should be documentation of what objects are produced by this function and attached self, their types and shapes, etc.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+    def subtract_model(self, keys=None, data=None, overwrite=False, ifft=True, ifftshift=True,
+                       window='none', alpha=0.2, edgecut_low=0, edgecut_hi=0, verbose=True):
+        """
+        Subtract pcomp_model from data.
+
+        FFT pcomp_model to frequency space, divide by window, and subtract from data.
+        Inserts FFT of pcomp_model into self.pcomp_model_fft and 
+        residual of data with self.data_pmodel_resid. 
+
+        Note: The windowing parameters should be the *same* as those that were used
```
jsdillon: I'm confused. The default for fft_data is ifft=True and the default for this is ifft=True. It seems to me like if you use both in their default configurations, they should work. However, to get the normalization right, you want one fft and one ifft. So perhaps the default there should be fft and the default here should be ifft. In either case, it's not always the "same" parameters but sometimes complementary ones.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+            self.pcomps[k] = _form_PCs(u[k], v[k])
+        if hasattr(u, 'times'):
+            self.pcomps.times = u.times
+
+    def build_model(self, keys=None, Nkeep=None, overwrite=False, increment=False, verbose=True):
+        """
+        Sum principal components to get a model.
+
+        Sum principal components dotted with singular values and add to 
+        the pcomp_model.
```
jsdillon: Same comment as above about documenting what gets attached.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+                window function across freq to divide by after FFT.
+            alpha : float
+                if window is Tukey, this is its alpha parameter.
+            edgecut_low : int
+                Nbins to flag but not window at low-side of band.
+            edgecut_hi : int
+                Nbins to flag but not window at high-side of band.
+            verbose : bool
+                If True, report feedback to stdout.
+        """
```
jsdillon: Again, document what gets attached to the object instead of the `Returns:`.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+            else:
+                self.pcomp_model[k] = model_fft
+
+    def subtract_model(self, keys=None, data=None, overwrite=False, ifft=True, ifftshift=True,
+                       window='none', alpha=0.2, edgecut_low=0, edgecut_hi=0, verbose=True):
+        """
+        Subtract pcomp_model from data.
+
+        FFT pcomp_model to frequency space, divide by window, and subtract from data.
+        Inserts FFT of pcomp_model into self.pcomp_model_fft and
```
jsdillon: I find this nomenclature confusing. `model_fft` and `pcomp_model` are in delay space, but `pcomp_model_fft` is in frequency space. That's not intuitive at all.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+            else:
+                model_fft = np.fft.fft(model, axis=-1)
+
+            # divide by a window
+            win = dspec.gen_window(window, model_fft.shape[1], alpha=alpha,
+                                   edgecut_low=edgecut_low, edgecut_hi=edgecut_hi)
+            model_fft = np.true_divide(model_fft, win, where=~np.isclose(win, 0.0))
+
+            # subtract from data
+            self.pcomp_model_fft[k] = model_fft
```
jsdillon: I'm getting concerned about the memory bloat of this code. By this point, you've stored at least half a dozen things as large as the data itself. Does it have to be this way?

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+
+        if not hasattr(self, 'pcomps'):
+            self.pcomps = DataContainer({})
+
+        for k in keys:
+            if k in self.pcomps and not overwrite:
+                echo("{} exists in pcomps and overwrite == False, skipping...".format(k), verbose=verbose)
+                continue
+
+            self.pcomps[k] = _form_PCs(u[k], v[k])
```
jsdillon: This seems like unnecessary memory bloat. You shouldn't ever have to store all the outer products. You should form them on the fly as necessary and sum them with the appropriate singular value weights.

hera_cal/reflections.py
```diff
@@ -442,85 +832,194 @@ def fit_reflection(data, dly_range, freqs, full_freqs=None, edgecut=0,
...
+    Dot singular values into principal components,
+    keeping a specified number of PCs. svals should
+    be rank ordered from highest to lowest.
+
+    Parameters
+    ----------
+    svals : 1D ndarray
+        singular values, with length Nvals
+
+    PCs : 3D ndarray
```
jsdillon: This doesn't make sense. You've artificially produced a much larger object in memory than you need. Do the outer products on the fly so you never need to store anything larger than the original data.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+            model_fft = np.true_divide(model_fft, win, where=~np.isclose(win, 0.0))
+
+            # subtract from data
+            self.pcomp_model_fft[k] = model_fft
+            self.data_pmodel_resid[k] = data[k] - model_fft
+
+    def interp_u(self, u, times, full_times=None, uflags=None, keys=None, overwrite=False, verbose=True,
+                 mode='gpr', gp_len=600, gp_nl=0.1, optimizer=None):
+        """
+        Interpolate u modes along time, inserts into self.umode_interp
```
jsdillon: Again, explain what umode_interp is, it's keys, it's values, their shapes.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+                If True, overwrite output data if it exists.
+            verbose : bool
+                If True, report feedback to stdout.
+            mode : str
+                Interpolation mode. Options=['gpr']
+            gp_len : length-scale of GPR in units of times
+            gp_nl : GPR noise-level in units of input u.
+        """
+        if not hasattr(self, 'umode_interp'):
+            self.umode_interp = DataContainer({})
```
jsdillon: Is DataContainer the right format for this? Like, if you give it reversed keys, do you want the complex conjugate? It's no longer nTimes x nFreqs or nTimes x nDelays, so I'm not sure if you want this to be a DataContainer as opposed to a dictionary.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+        # iterate over keys
+        for k in keys:
+            # check overwrite
+            if k in self.umode_interp and not overwrite:
+                echo("{} in umode_interp and overwrite == False, skipping...".format(k), verbose=verbose)
+                continue
+
+            if mode == 'gpr':
+                # setup GP kernel
+                kernel = 1**2 * gp.kernels.RBF(length_scale=gp_len) + gp.kernels.WhiteKernel(noise_level=gp_nl)
```
jsdillon: What's the 1**2 for?

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+                y = u[k][select, :]
+
+                # fit gp and predict
+                GP.fit(X, y.real)
+                ypredict_real = GP.predict(Xpredict)
+                GP.fit(X, y.imag)
+                ypredict_imag = GP.predict(Xpredict)
+
+                # append
+                self.umode_interp[k] = ypredict_real.astype(np.complex) + 1j * ypredict_imag
```
jsdillon: The `.astype(np.complex)` is unnecessary. The sum will handle the type conversion.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+
+    def project_autos_onto_u(self, keys, auto_keys, u=None, flags=None, index=0, auto_delay=0,
+                             overwrite=False, verbose=True):
+        """
+        Project autocorr onto u modes.
+
+        Projects the time dependent dfft of an autocorrelation
+        at a specified delay onto the specified u-mode, replacing
+        the u-mode with the projected autocorrelation. Inserts
+        results into self.umode_interp
```
jsdillon: again, explain self.umode_interp more fully.

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+                af = np.sum([flags[_ak] for _ak in ak], axis=0)
+            else:
+                if ak not in self.dfft:
+                    echo("{} not in self.dfft, skipping...".format(ak), verbose=verbose)
+                    continue
+                A = np.array([self.dfft[ak][:, select]]).T
+                af = flags[ak]
 
-def _form_gains(epsilon):
+            # form least squares estimate
```
jsdillon: You need to explain the reasoning behind this step. What's going on here and why?

hera_cal/reflections.py
```diff
@@ -267,25 +239,445 @@ def write_reflections(self, output_calfits, input_calfits=None, overwrite=False,
...
+
+        Projects the time dependent dfft of an autocorrelation
+        at a specified delay onto the specified u-mode, replacing
+        the u-mode with the projected autocorrelation. Inserts
+        results into self.umode_interp
+
+        Args:
+            keys : list of tuples
+                List of baseline-pol DataContainer tuples to operate on.
+            auto_keys : list of tuples
```
jsdillon: I don't get this at all. For a given baseline, aren't there two auto-correlations to look at (one for each antenna)? And shouldn't that be the default behavior?

hera_cal/reflections.py
jsdillon: Why does this put its results into the same place as `self.interp_u`? Is it an alternative to that? It's very unclear what this function does or when to use it.

hera_cal/reflections.py
```diff
@@ -309,85 +701,44 @@ def construct_reflection(freqs, amp, tau, phs, real=False):
...
+    eps = amp * np.exp(2j * np.pi * tau * freqs + 1j * phs)
     if real:
         eps = eps.real
 
     return eps
 
 
-def fit_reflection(data, dly_range, freqs, full_freqs=None, edgecut=0,
-                   taper='none', zero_pad=0, real=False, fthin=10, **win_kwargs):
+def fit_reflection_delay(dfft, dly_range, dlys, return_peak=False):
```
jsdillon: It seems like there's a lot of repeated code between this and fft_dly. Can you modify one or both of them to eliminate the repeated code?

hera_cal/reflections.py
```diff
@@ -398,37 +749,76 @@ def fit_reflection(data, dly_range, freqs, full_freqs=None, edgecut=0,
...
+    peak, _ = fit_reflection_delay(dfft, (-np.abs(dly_range).min(), np.abs(dly_range).min()), dlys, return_peak=True)
 
     # get reflection amplitude
-    ref_amps = ref_peaks / np.max(abs_dfft[:, np.abs(dlys) < np.abs(dly_range).min()], axis=1, keepdims=True)
+    ref_amps = np.true_divide(ref_peaks, peak, where=~np.isclose(peak, 0.0))
+
+    return ref_amps, ref_dlys, ref_dly_inds, ref_significance
+
+
+def fit_reflection_phase(dfft, dly_range, dlys, ref_amps, ref_dlys, fthin=1, Nphs=500,
```
jsdillon: This whole algorithm feels a bit janky. What kind of accuracy can you actually get? Presumably no better than 2pi over the number of phase bins...

I'd recommend trying `scipy.optimize.minimize` or `scipy.optimize.curve_fit` or similar. Should be fairly straightforward to fit a cosine rather than generating a grid of cosines.

hera_cal/reflections.py
```diff
@@ -442,85 +832,194 @@ def fit_reflection(data, dly_range, freqs, full_freqs=None, edgecut=0,
...
+def auto_reflection_argparser():
+    a = argparse.ArgumentParser(description='Model auto (e.g. cable) reflections from auto-correlation visibilities')
+    a.add_argument("clean_data", nargs='*', type=str, help="CLEAN data file paths to run auto reflection modeling on")
+    a.add_argument("--output_fname", type=str, help="Full path to the output .calfits file")
+    a.add_argument("--dly_range", type=float, nargs='*', help='lower and upper delay [nanosec] within which to search for reflections.')
+    a.add_argument("--filetype", type=str, default='uvh5', help="Filetype of datafile")
+    a.add_argument("--overwrite", default=False, action='store_true', help="Overwrite output file if it already exists")
+    a.add_argument("--write_npz", default=False, action='store_true', help="Write NPZ file with reflection params with same path name as output calfits.")
+    a.add_argument("--input_cal", type=str, default=None, help="Path to input .calfits to apply to data before modeling")
+    a.add_argument("--antenna_numbers", default=None, type=int, nargs='*', help="List of antenna numbers to operate on.")
```
jsdillon: When default=None means do everything, you should explain that.

hera_cal/reflections.py
nkern: The broadcasting is something I do routinely: my CASA calibration solutions are single spectra gains, so anytime I apply the CASA gains using `apply_cal` I rely on this broadcasting ability.

hera_cal/reflections.py
nkern: 1. Sure thing

2. Hmm, I don't see how setting them to `None` by default helps clarity at all: I don't think its useful to have a bunch of `None` attributes for a user who, given their use-case, may not even create those structures anyways. Something like what `pyuvdata` does, which is to have them by default set to an object that is empty but holds a doc-string and other metadata would be helpful, but I think is massive overkill for the stage of this code right now.

hera_cal/reflections.py
nkern: There is quite a lot of utility for doing that actually, and there are two functions provided below to do exactly that. But I agree that we can fold this into `build_model` to reduce memory bloat

hera_cal/reflections.py
nkern: There is a little bit of repeated code, but its just the peak shifting thats similar: the two functions are working with fundamentally different inputs, so there is no easy way to use one for both.

hera_cal/reflections.py
nkern: So the adopted algorithm is done for speed: this is the fastest thing I could get that hits the necessary precision, I experimented with `scipy` minimization algorithms at one point and found it significantly slower. Precision is indeed 2pi / `Nphs`, which is exposed to the user in case higher precision is required...

hera_cal/reflections.py
nkern: Yes, we do want all of the capabilities of `DataContainer` including automatic conjugation for these products

hera_cal/reflections.py
nkern: Agreed: the default for `fft_data` is `ifft=False`, I believe, whereas the default here is `ifft=True`.

hera_cal/reflections.py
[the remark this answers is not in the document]
jsdillon: how about `edgecut_low : int, Nbins to flag and exclude from windowing function on low-side of the band (thus shrinking the size of the window)`

hera_cal/vis_clean.py
```diff
@@ -100,7 +102,7 @@ def clear_containers(self, exclude=[]):
                 continue
             obj = getattr(self, key)
```
jsdillon: You can delete this line.

hera_cal/vis_clean.py
```diff
@@ -217,9 +219,11 @@ def write_data(self, data, filename, overwrite=False, flags=None, nsamples=None,
         echo("...writing to {}".format(filename), verbose=verbose)
 
     def vis_clean(self, keys=None, data=None, flags=None, wgts=None, ax='freq', horizon=1.0, standoff=0.0,
-                  min_dly=0.0, max_frate=None, tol=1e-6, maxiter=100, window='none',
+                  min_dly=0.0, max_frate=None, tol=1e-6, maxiter=100, window='none', zeropad=0,
```
jsdillon: `zeropad` is undocumented.

hera_cal/vis_clean.py
```diff
@@ -474,20 +498,214 @@ def fft(d, ifft=ifft, shift=fftshift, ax=-1):
...
+        # iterate over keys
+        for k in keys:
+            factorize_flags(flags[k], spw_ranges=spw_ranges, time_thresh=time_thresh, inplace=True)
+
+        if not inplace:
+            return flags
+
+
+def fft_data(data, delta_bin, wgts=None, axis=-1, window='none', alpha=0.2, edgecut_low=0,
+             edgecut_hi=0, ifft=False, ifftshift=False, fftshift=True, zeropad=0):
```
jsdillon: I don't think we need both a fftshift and an ifftshift kwarg. If fftshift is true and ifft is true, it should do an ifftshift automatically (though this should be explained in the docstring).

hera_cal/vis_clean.py
```diff
@@ -474,20 +498,214 @@ def fft(d, ifft=ifft, shift=fftshift, ax=-1):
...
+
+        fourier_axes.append(fax)
+
+    if len(axis) == 1:
+        fourier_axes = fourier_axes[0]
+
+    return data, fourier_axes
+
+
+def _zeropad_array(data, zeropad=0, axis=-1, undo=False):
```
jsdillon: I think this function is reinventing the wheel. Numpy already has zeropadding in `np.pad` which covers most of the functionality. I can see an argument for writing a function that undoes the effect of np.pad with a set of kwargs, but we should stick to standard libraries and syntax as much as possible.

jsdillon:
Ok, @nkern I'm done going through the rest of the code. I've closed some conversations where I think we've reached a satisfactory conclusion and opened a few relatively minor ones on vis_clean.py, which I hadn't looked at before we talked yesterday.

So, there's still a handful of open threads we should hash out. Also, your tests are failing. But otherwise, I think we're getting pretty close to convergence on this PR.

LMK if you have any questions.

hera_cal/reflections.py
nkern: I've done tests and the phase is actually not the limiting factor for sufficiently high `Nphs` (the default parameter). But sure, the quadratic interpolation can't hurt..

hera_cal/vis_clean.py
nkern: No its still needed. These are doing three fundamentally different things. One is an ifftshift before the FFT, the next is whether or not to use fft or ifft in the FFT, and the last is whether to shift it after the FFT. Just because I want to do an fftshift after the ifft doesn't mean I need to ifftshift it before the ifft.

hera_cal/vis_clean.py
nkern: Ah I didn't know the function existed. Regardless, this function seems to have two critical capabilities that `np.pad` doesn't: 1) support for an `axis` and 2) and `undo` kwarg.

I also don't think not using `np.pad` in this case is that big of a deal, honestly...

nkern:
thanks @jsdillon, I incorporated some of your new suggestions and addressed the others. 

I also made the spectral bin shifting its own function and made `fft_dly` and `fit_reflection_dly` use that, as you had suggested earlier.

About the failing test, its just ensuring that `np.fft.fft(np.fft.ifft(data)) == data`, which passes for me locally but for some reason isn't passing on travis. I just got rid of the check b/c I couldn't figure out why it wasn't passing on travis and it was indeed a somewhat trivial test.

I think I've tied up all the loose ends. LMK if you have any further concerns.

PR #401, opened 2018-12-06
Added PseudoContainer to HERAData

nkern:
Forgive me @jsdillon for the numerous PRs: but I think I may have solved an issue that was bugging us for a while, which is the fact that in the `DelayFilter` and `FRFilter` objects, upon instantiation, we automatically double the memory footprint by linking the input `UVData` object and __then__ making series of `DataContainers` from them. This gets around this by utilizing a new object `SoftContainer` that mimics some of the basic functionality of a `DataContainer` but doesn't make a copy of the data, instead it just references the data in the `HERAData` object via its `get_slice` method.

Once we implement this we should be able use the `SoftContainer` linked as `HERAData.data` and `HERAData.flags` in our `DelayFilter` and `FRFilter` objects and not have to double the memory footprint...

Btw, I'm loving the new `HERAData` object and its `blt_slice` caching...

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -405,6 +405,11 @@ class HERAData(UVData):
                                  times=times, frequencies=frequencies, freq_chans=freq_chans)
                 self.unphase_to_drift()
 
+        # link SoftContainers
+        self.data = SoftContainer(self, 'data_array')
+        self.flags = SoftContainer(self, 'flag_array')
+        self.nsamples = SoftContainer(self, 'nsample_array')
+
         # process data into DataContainers
         if read_data or self.filetype == 'uvh5':
             self._determine_blt_slicing()
@@ -414,7 +419,11 @@ class HERAData(UVData):
 
     def __getitem__(self, key):
         '''Shortcut for reading a single visibility waterfall given a baseline tuple.'''
-        return self.read(bls=key)[0][key]
+        try:
+            return (self._get_slice(self.data_array, key), self._get_slice(self.flag_array, key),
+                    self._get_slice(self.nsample_array, key))
+        except KeyError:
+            return self.read(bls=key)[0][key]
 
     def update(self, data=None, flags=None, nsamples=None):
         '''Update internal data arrays (data_array, flag_array, and nsample_array)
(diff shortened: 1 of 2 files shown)


jsdillon:
Glad you're liking the new `HERAData` objects, though we'll see how much you'll love them after this ;)

I don't think this PR does what you think it does, unfortunately. `_get_slice`, by construction, always returns a copy of the data, not a reference. This is because we want to support conjugation of reversed baselines, which will always make a copy, so we just make everything make a copy. This is definitely not a soft-link. It's just partial copying. So I think it's confusing to call it a SoftContainer. 

So what you're doing is loading in the whole data set, copying one baseline's (e.g.) data, perhaps manipulating it, and then putting it back into the HERAData object. The `DataContainer`s spit out by a `HERAData` object are always meant to be copies and the only current "official" way to update the original HERAData object is with `update()`. The fundamental design idea is that if you're concerned about memory, you should be doing partial data loading. Doubling the memory footprint when you're only loading a small fraction of the data was thought to be a reasonable sacrifice for a guarantee that the data in the HERAData object isn't changing out from under you. Aaron and I, let's say, "debated vigorously" about this and originally I designed HERAData to return slices linked to the data. 

That said, I'm not fundamentally against this PR @nkern, I just want to reconsider whether we can accomplish everything you want with partial data loading because we need to be moving everything that way soon.

nkern:
Ah I see, I hadn't realized it was returning a copy of the data. I guess I just assumed from the name `slice` that it was a reference. Okay, then yes I think this is not doing what I thought it was doing, unfortunately. 

Funny enough, I think this __is__ still actually doing what I intend, just not the way I intended it to. Think of it this way, even if it returns a copy of the data, that copy is manipulated in some way and then assigned a home: in the case of `DelayFilter`, its home ends up being in `DF.clean_model` or something. That home was always going to be a copy of the data, its just that now, I don't have __another__ copy of the original data living in `DF.data`. So I think I still reduced the end-result memory footprint here. But indeed, it isn't a softlink to the data so we shouldn't call it `SoftContainer`.

One compromise is that we could enable an advanced-option on `read` that makes the `SoftContainer` a true soft-link by making the `blt_slice` a reference for unconjugated keys and a copy for conjugated keys. The default could be copies both ways, like how it is now. That would of course expose us to the possibility of data-changing-from-under-us, but I'm actually entirely okay with that, as an advanced user.

Let me know what you think. If you want we could schedule a time to chat early next week so we can weigh the options, and also I can give you a quick run-down of my other PRs to limit the amount of time you need to spend on them :) thanks again..

nkern:
Actually, @jsdillon I think I may have exposed a bug in your implementation... Before, I actually made a unit test to confirm that my `SoftContainers` were indeed "soft" and it passes. I.e. I can indeed update the underlying data via my SoftContainer. I would double check your assumption that `get_slices` returns copies of the data..

jsdillon:
Yes, let's definitely do this over video chat. I'm free late tonight Pacific time.

nkern:
[Edit 12/8] We are not merging in Soft (or Pseudo) Containers for the time being, but will merge in the the __getitem__ updates as a separate PR

PR #404, opened 2018-12-19
made abscal take calibration solutions

nkern:
for future RTP where we only store raw data and calibration solutions


nkern:
Closing this PR: moving these additions to a larger PR that preps abscal and lstbin for IDR3 push

Issue #317, opened 2018-06-12
[IDR2.1] propagating noise through calibration pipe


nkern:
Following the 6/7 pspec meeting, and based on the weird spectral structure in the variance plots from the IDR2.1 LSTbinner, we decided we need to get a handle on how the calibration pipeline affects noise in our raw data. To this end, we decided to start by getting a post-correlator noise model, generating some fake data that follow that model, and propagating that through the analysis pipeline through to the LSTbinner.

@jsdillon volunteered to start working on the noise model, with me and @plaplant working on running it through the analysis pipe.

jsdillon:
I think that this has largely been addressed by the consistency between the noise observed in the variance of days LST-binned together, the noise observed by freq/time differencing, and the noise predicted from autocorrelations, as demonstrated in this (draft) IDR 2.2. memo: https://nbviewer.jupyter.org/github/HERA-Team/hera_sandbox/blob/master/jsd/IDR2_2/IDR2.2_Memo.ipynb

Issue #394, opened 2018-11-07
Write hera_cal git hash to file outputs 


jsdillon:

jsdillon:
This should probably wait for #391

plaplant:
@jsdillon is this still an issue?

jsdillon:
Yes, I believe so.

jsdillon:
This was discussed on 1/2/19 and it was decided to add this to all written file histories at the module level. I volunteered to do it.

PR #406, opened 2018-12-21
Add autos.py and noise.py modules

### LINKED ISSUE #405: Save out raw autocorrelations as a dataproduct in the IDR2.2 run
When combined with gain tables and flags having the autos available in the release will help us assess calibration and noise levels.

A  uvh5 file is preferred.

jsdillon:
This PR adds two related modules to hera_cal. The first, autos.py, includes code for extracting autocorrelations from a file and writing them to disk (and optionally applying calibration solutions and associated flags). This resolves #405.

The second is a module for loading autocorrelations and using them to produce visibility-like files that include per-antenna noise predictions. This new module, noise.py, 
I've moved `predict_noise_variance_from_autos` out of utils.py into the new module and also codified a interleaved noise variance estimator that I had been using in its unit tests and made it its own function with unit tests. This doesn't entirely finish #317, but it's a decent step toward putting it to rest.

I've also added scripts for command line/pipeline usage of both functions.


hera_cal/noise.py
```diff
@@ -0,0 +1,120 @@
...
+
+    Returns:
+        noise: DataContainer mapping autocorrelation baselines like (1, 1, 'xx') to predictions for the standard
+            deviation of noise for that antenna.
+    '''
+    noise = {}
+    for bl in autos:
+        if (bl[0] == bl[1]) and (split_pol(bl[2])[0] == split_pol(bl[2])[1]):
+            noise[bl] = np.sqrt(predict_noise_variance_from_autos(bl, autos, dt=dt, df=df))
+    return DataContainer(noise)
```
dannyjacobs: This redefines a visibility autocorrelation as a per antenna quantity with units sqrt(Jy).  This is bound to lead to confusion later on.  If DataContainer is assumed to be visibilities, then it is not the right vehicle for the result. If its more general (ie can also hold calibration solutions) then it should somehow indicate units.

hera_cal/noise.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def write_per_antenna_noise_std_from_autos(infile, outfile, calfile=None, gain_convention='divide', add_to_history='', clobber=False):
```
dannyjacobs: Despite the fact that this potentially replaces data with simulation, the default action is to not add any history remarks.  I'd prefer to see a default history message appended.

dannyjacobs: Same for the rest of the functions which generate files.

hera_cal/autos.py
```diff
@@ -0,0 +1,52 @@
...
+        outfile: string path to .uvh5 output data file containing only autocorrelations
+        calfile: optional string path to .calfits calibration file to apply to autocorrelations before writing.
+            Can also take a list of paths if the .calfits files can be combined into a single UVCal object.
+        gain_convention: str, either 'divide' or 'multiply'. 'divide' means V_obs = gi gj* V_true,
+            'multiply' means V_true = gi gj* V_obs.
+        add_to_history: appends a string to the history of the infile when writing the outfile
+        clobber: if True, overwrites existing file at outfile
+    '''
+    hd = io.HERAData(infile)
+    auto_bls = [bl for bl in hd.bls if (bl[0] == bl[1] and split_pol(bl[2])[0] == split_pol(bl[2])[1])]
```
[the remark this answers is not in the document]
jsdillon: the `.bls` attribute of a HERAData object is a list of tuples of the form `(0,1,'xx')`. So this just goes through and takes only the pairs that are not cross-correlations or cross-polarization.

I'll add a comment to explain.

hera_cal/noise.py
```diff
@@ -0,0 +1,120 @@
...
+import scipy
+
+from . import io
+from .utils import split_pol, join_pol
+from .apply_cal import calibrate_in_place
+from .datacontainer import DataContainer
+
+
+def interleaved_noise_variance_estimate(vis, kernel=[[1, -2, 1], [-2, 4, -2], [1, -2, 1]]):
+    '''Estimate the noise on a visibiltiy per frequency and time using sequential differecing.
```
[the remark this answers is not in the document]
jsdillon: Fixed and clarified.

hera_cal/noise.py
jsdillon: `predict_noise_variance_from_autos` returns `Noise variance predicted on baseline bl in units of data squared`, so basically Jy^2. That means that this noise function returns a standard deviation in Jy. I did this exactly for the reason you bring up, which is that DataContainers generally hold things with units of visibilities.

I added a comment to clarify in the `Returns` of `per_antenna_noise_std`

hera_cal/noise.py
jsdillon: Thus far, this is how we've been doing this throughout hera_cal and I'd like to remain consistent (we use the run scripts to add to the history rather than the functions themselves). We have #394 open to do a better job of histories and I think this is a broader issue we should revisit as a group. If you'd like to add to that issue or open a new one, go ahead. But I think that's separate from this PR and should be handled all at once in a consistent way. 

I'll add it to the agenda for Wednesday.

hera_cal/noise.py
```diff
@@ -17,12 +17,13 @@
 
 
 def interleaved_noise_variance_estimate(vis, kernel=[[1, -2, 1], [-2, 4, -2], [1, -2, 1]]):
```
dannyjacobs: Wait, is this differencing auto correlations, visibilities, or both?

jsdillon: It's meant for visibilities, but in theory you could find the "noise" on autocorrelations using it.

PR #408, opened 2018-12-22
Remove explicit test of the number of warnings produced by a pyuvdata io task.

### LINKED ISSUE #407: Remove explicit test of number of warnings from a pyuvdata i/o task.
A new pyuvdata warning breaks this line:

https://github.com/HERA-Team/hera_cal/blob/b54c4cf0f698b991f9a2cefdeeb797490b1fb1ce/hera_cal/tests/test_utils.py#L314

Now we're getting two warnings when one was expected. It seems crazy to me that we should be testing the number of warnings, so I think we should remove this part of the test.

jsdillon:
This closes #407.

### DIFF
--- a/hera_cal/tests/test_utils.py
+++ b/hera_cal/tests/test_utils.py
@@ -8,6 +8,7 @@ import nose.tools as nt
 import numpy as np
 import sys
 import os
+import warnings
 import shutil
 import glob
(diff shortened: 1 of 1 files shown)


jsdillon:
Thanks @adampbeardsley

PR #409, opened 2018-12-27
[delay_filter] get filled data uses clean_flags

nkern:
Fixes a typo (that was just introduced in a recent PR) in `DelayFilter.get_filled_data`where `filled_data` was deepcopy of data, when it should have been deepcopy of clean_model...


nkern:
@jsdillon For this quick fix, could you take a brief look at this and merge it whenever you are done? Its just a small change for a recent edit to `DF.get_filled_data`. the problem was when `run_filter` was run on only a subset of the data, the `for` loop would iterate over all `data` keys, now it should be iterating over only `clean_model` keys, which is more correct.

PR #410, opened 2018-12-28
vis_clean.write_data fix for subset of data write

nkern:
@jsdillon another small patch for `vis_clean.write_data`: when writing a subset of the input data it failed to re-compute `HERAData._determine_blt_slices`, which was a use-case I hadn't anticipated, but in making the IDR3 abscal model I stumbled upon. Anyways, its a quick fix. Thanks! Almost finished with the idr3 abscal model :)


jsdillon:
(Sorry for being slow in my review. Somehow I missed this PR. If I'm slow, feel free to ping me on slack.)

PR #411, opened 2019-01-02
updates to abscal, lstbin (and io) for IDR2.2 and IDR3

nkern:
Updates to `abscal` and `lstbin` for UVH5 compatibility and for applying calibration solutions on-the-fly


hera_cal/abscal.py
```diff
@@ -1026,7 +1045,9 @@ def omni_abscal_arg_parser():
     a = argparse.ArgumentParser(description="command-line drive script for hera_cal.abscal module")
     a.add_argument("--data_file", type=str, help="file path of data to-be-calibrated.", required=True)
     a.add_argument("--model_files", type=str, nargs='*', help="list of data-overlapping miriad files for visibility model.", required=True)
-    a.add_argument("--calfits_infile", type=str, help="path to calfits file to multiply with abscal solution before writing to disk.")
+    a.add_argument("--filetype", type=str, default='uvh5', help="Filetype of input filepaths.")
+    a.add_argument("--input_cal", type=str, help="Path to a calfits file to apply to the data before running abscal.")
+    a.add_argument("--secondary_cal", type=str, help="Path to a calfits file to multiply with abscal solution before writing to disk.")
```
nkern: get rid of this capability

hera_cal/abscal.py
```diff
@@ -78,7 +78,7 @@ class AbsCal(object):
     """
     def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
```
jsdillon: As a general comment, I find it odd that we support so many different ways of instantiating this object and it's very difficult to keep track of how weights and flags are propagated. Also, it's pretty strange that a user would specify the data as a DataContainer but the calibration as a file path or vice versa.

Perhaps this can wait for some future analysis overhaul/streamlining, but I think we can make the code a lot more readable and maintainable if we impose consistent interfaces externally. For example, either it's an abscal object's responsibility to figure out its own weights or it's not.

hera_cal/abscal.py
```diff
@@ -987,7 +1004,9 @@ def abscal_arg_parser():
     a = argparse.ArgumentParser(description="command-line drive script for hera_cal.abscal module")
     a.add_argument("--data_file", type=str, help="file path of data to-be-calibrated.", required=True)
     a.add_argument("--model_files", type=str, nargs='*', help="list of data-overlapping miriad files for visibility model.", required=True)
-    a.add_argument("--calfits_infile", type=str, help="path to calfits file to multiply with abscal solution before writing to disk.")
+    a.add_argument("--filetype", type=str, default='uvh5', help="Filetype of input filepaths.")
+    a.add_argument("--input_cal", type=str, help="Path to a calfits file to apply to the data before running abscal.")
+    a.add_argument("--secondary_cal", type=str, help="Path to a calfits file to multiply with abscal solution before writing to disk.")
```
jsdillon: Likewise, dump this too.

hera_cal/abscal.py
```diff
@@ -1088,10 +1109,13 @@ def abscal_run(data_file, model_files, filetype='miriad', refant=None, calfits_i
 
     filetype : str, filetype of input data and models
 
-    calfits_infile : type=str, path to calfits files containing gain solutions
+    secondary_cal : type=str, path to calfits files containing gain solutions
```
jsdillon: And get rid of this.

hera_cal/abscal_funcs.py
```diff
@@ -1602,8 +1602,8 @@ def avg_data_across_red_bls(data, antpos, wgts=None, broadcast_wgts=True, tol=1.
     return DataContainer(red_data), DataContainer(red_wgts), red_keys
 
 
-def avg_file_across_red_bls(data_fname, outdir=None, output_fname=None,
-                            write_miriad=True, output_data=False, overwrite=False,
+def avg_file_across_red_bls(data_fname, filetype='uvh5', outdir=None, output_fname=None,
```
jsdillon: This is never used anywhere in the codebase. Can we delete it?

hera_cal/io.py
```diff
@@ -388,23 +394,26 @@ def read(self, bls=None, polarizations=None, times=None,
...
+                if self.filetype == 'miriad':
+                    self.read_miriad(self.filepaths, bls=bls, polarizations=polarizations)
+                    if any([times is not None, frequencies is not None, freq_chans is not None]):
+                        warnings.warn('miriad does not support partial loading for times and frequencies. '
+                                      'Loading the file first and then performing select.')
+                        self.select(times=times, frequencies=frequencies, freq_chans=freq_chans)
+                elif self.filetype == 'uvfits':
+                    self.read_uvfits(self.filepaths, bls=bls, polarizations=polarizations,
+                                     times=times, frequencies=frequencies, freq_chans=freq_chans)
+                    self.unphase_to_drift()
```
jsdillon: Don't forget to add errors if you try to do a select and there's no self.filepaths

hera_cal/io.py
```diff
@@ -626,6 +635,92 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
...
+
+    file_starts = np.array(file_starts)
+    file_stops = np.array(file_stops)
+    int_times = np.array(int_times)
+
+    # make sure times don't wrap
+    file_starts[np.where(file_starts < 0)] += 2 * np.pi
+    file_stops[np.where(file_stops >= 2 * np.pi)] -= 2 * np.pi
+
+    if _array is False:
```
jsdillon: I think it's bad for the format of the output to depend silently on the format of the input. I'm fine with the returned starts and stops always being lists. Otherwise, I'd explicitly say in the documentation that it returns floats if a single file is input, otherwise it returns lists.

hera_cal/io.py
```diff
@@ -626,6 +635,92 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
...
+
+    # get Nfiles
+    Nfiles = len(filepaths)
+
+    # iterate over filepaths and extract time info
+    for i, f in enumerate(filepaths):
+        if filetype == 'miriad':
+            uv = aipy.miriad.UV(f)
+            # get integration time
+            int_time = uv['inttime'] * 2 * np.pi / (23.9344699 * 3600.)
```
jsdillon: I recommend using `astropy.units.si.sday.in_units(astropy.units.si.s)`. You get a slightly different answer, as it turns out (86164.09164 vs. 86164.09053).

hera_cal/io.py
```diff
@@ -626,6 +635,92 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
...
+    Miriad standard is bin start, so shift by int_time / 2 is done.
+    UVH5 standard is bin center, so lsts are left untouched.
+
+    Parameters:
+    ------------
+    filepaths : type=list, list of filepaths
+
+    filetype : str, options=['miriad', 'uvh5']
+
+    add_int_buffer : type=bool, if True, extend stop times by an integration duration.
```
jsdillon: do you want to add a comment for why you might need this option?

hera_cal/io.py
```diff
@@ -626,6 +635,92 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
...
+def get_file_lst_range(filepaths, filetype='uvh5', add_int_buffer=False):
+    """
+    Get a file's start, stop and dlst in radians.
+
+    Start and stop are bin center.
+
+    Miriad standard is bin start, so shift by int_time / 2 is done.
+    UVH5 standard is bin center, so lsts are left untouched.
+
+    Parameters:
```
jsdillon: This docstring has a different style from the rest. I don't really care, but maybe you do.

hera_cal/lstbin.py
```diff
@@ -422,23 +424,19 @@ def lst_align_arg_parser():
...
 
     Parameters:
     -----------
     data_files : type=list, list of paths to miriad files, or a single miriad file path
-
     file_ext : type=str, file_extension for each file in data_files when writing to disk
-
     dlst : type=float, LST grid bin interval, if None get it from first file in data_files
-
+    filetype : str, filetype of filepaths in data_files. options=['miriad', 'uvh5']
```
jsdillon: I believe this also supports uvfits without you having to do any extra work.

hera_cal/lstbin.py
```diff
@@ -422,23 +424,19 @@ def lst_align_arg_parser():
     return a
 
 
-def lst_align_files(data_files, file_ext=".L.{:7.5f}", dlst=None,
+def lst_align_files(data_files, file_ext=".L.{:7.5f}", dlst=None, filetype='uvh5',
```
jsdillon: It seems like this function and its associated arg_parser and script are never used. Can we just remove them?

hera_cal/lstbin.py
```diff
@@ -611,33 +608,26 @@ def lst_bin_files(data_files, dlst=None, verbose=True, ntimes_per_file=60, file_
     data_files : type=list of lists: nested set of lists, with each nested list containing
                  paths to miriad files from a particular night. These files should be sorted
```
jsdillon: Update: 'not just miriad'

hera_cal/abscal.py
nkern: Yeah

hera_cal/lstbin.py
nkern: sure

hera_cal/io.py
jsdillon: Ah, ok. Good point.

PR #413, opened 2019-01-04
Add git info writing to history for all modules

### LINKED ISSUE #394: Write hera_cal git hash to file outputs

jsdillon:
This PR adds a function in version.py called `history_string()` that produces a human-readable string that can be appended to histories of HERAData/HERACal/UVCal/UVData objects before writing them out. The history strings know about the the name of the function and the file that called them (by examining the stack), so histories look like:

> `------------`
> `This file was produced by the function redcal_run() in redcal.py using: `
> `    git_branch: git-hash-histories`
> `    git_description: v2.0-434-g986bed7-dirty`
> `    git_hash: 986bed748e8bd5b8ba2928d95b7709991ddb5cb1`
> `    git_origin: https://github.com/HERA-Team/hera_cal.git`
> `    version: 2.0`
> `------------`

Calls to `version.history_string()` have been added to all major file writing functions. In the future, I think this is an easy standard to stick to. I realize that this could have been accomplished slightly more simply by just modifying io.py. However, this has two advantages. First, some functions in hera_cal call pyuvdata writing directly, bypassing io.py. Second, this gave me an opportunity to harmonize how histories are written (and fix a few bugs along the way). 

This resolves #394.


jsdillon:
I like that idea @nkern. I'll do something like that.

PR #422, opened 2019-01-20
Abscal Updates: remove abscal_funcs.fft_dly and speed up get_file_lst_range

### LINKED ISSUE #367: Update fft_dly in abscal
First off, @AaronParsons has implemented an improved `fft_dly` in #366 and it would be good to have abscal use it instead of the one in abscal (deprecating the one in abscal)

Second, in this line

https://github.com/HERA-Team/hera_cal/blob/f4c742de572585c171a5a494ee253825c2ecb820/hera_cal/abscal_funcs.py#L442

I'd recommend using the ratio of data/np.abs(data) rather than the data itself to find delays. This is much more robust to RFI and might actually eliminate the need for the median filter. 

From what I can tell, this works quite well in firstcal: see https://github.com/HERA-Team/hera_cal/pull/366#issuecomment-423377739

jsdillon:
- Speeds up get_file_lst_range for uvh5 files dramatically by using h5py directly. This is useful because each file's abscal must go through all model files to check 
- Removes abscal_funcs.fft_dly in favor of utils.fft_dly, though edge_cut is added from the former to the latter. This closes #367.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -367,8 +367,8 @@ class AbsCal(object):
 
         # assign data
         model = self.model
-        data = copy.deepcopy(self.data)
-        wgts = copy.deepcopy(self.wgts)
+        data = self.data
+        wgts = self.wgts
 
         # get freq channel width
         df = np.median(np.diff(self.freqs))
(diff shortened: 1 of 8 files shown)


nkern:
This stuff looks good. I've made some edits to allow for passing `read_kwargs` to `abscal_run`, such that one can run on 4-pol data files but only load-in and calculate single-pol calibration solutions, if memory is an issue, and also tried to cut down on unnecessary `deepcopies` to help reduce memory load spikes.

I'm testing this now at NRAO to see if I can produce a couple of files worth of `redcal` + `abscal` data...

jsdillon:
And your changes look fine to me. I'm going to pull this in now and while you're working on getting abscal running, I'm going to take over re-writing abscal_run to use partial I/O.

Issue #319, opened 2018-06-14
Move all hera_cal modules to use memory-efficient partial data loading


jsdillon:
Starting at some point in H2C, files will be far to large to load entirely into memory and then manipulate, which is the current paradigm.

We can wait on this until the UVH5 format is done and pyuvdata supports it, but after that we need to get going ASAP. And simultaneously, we should probably upgrade to python 3 (#318).

jsdillon:
We just have abscal and lst-binning left to do here, so I'm assigning this to you, @nkern

PR #430, opened 2019-01-24
Abscal run rewrite

### LINKED ISSUE #262: Improve abscal tests

### LINKED ISSUE #335: Get abscal ready for IDR3 / H2C RTP
This includes:

* Making it python 2 and python 3 compatible (#318)
* Implementing partial data loading (#319) using the HERAData class in #331 and deprecating old cal_formats (see #330) in favor of hera_cal.io
* Fix the issue where dead ants in the model flag the data (#327)
* Robustify tests (#262)
* Update the example notebook (#266)

### LINKED ISSUE #355: No "from module import *"
I know this happens in abscal for import abscal_funcs, for example.

### LINKED ISSUE #375: Properly normalize abscal chisq relative to the noise in the visibilities and DoF
See redcal for an analogous implementation in omnical. 

Let me know if you need help.

jsdillon:
* Rewrites and simplifies abscal run functions to only do post-redcal abscal.
* Enables partial I/O in abscal, partially addressing #319.
* Better normalizes abscal chi^2 relative to the noise generated from autocorrelations. Does not properly take into account the degrees of freedom, which is non-trivial to count. I think this is good enough for now, so I say this closes #375.
* Merges abscal.py and abscal_funcs.py. This closes #355.
* With the new tests, I'd say this also close #262.

I think with this, we can finally close #335.


hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+
+    Returns:
+        d2m_time_map: dictionary uniqely mapping times in the data to times in the model 
+            that are closest in LST. Each model time maps to at most one data time and 
+            each model time maps to at most one data time. Data times without corresponding
+            model times map to None.
+    '''
+    # first produce a map of indices using the LSTs
+    m2d_ind_map = {}  
+    for dind, dlst in enumerate(data_lsts):
```
nkern: should consider the effects of LST wrap in this FOR loop

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+            which are then accessible as class properties of AC
+        AC_kwargs: dictionary of kwargs to pass into AC_func
+        gain_funcs: list of functions to call to return gains after AC_func has been called
+        gain_args_list: list of tuples of arguments to pass to the corresponding gain_funcs
+        gain_flags: per-antenna flags to apply to AC.Data when performing recalibration
+        gain_convention: either 'divide' if raw data is calibrated by dividing it by the gains
+            otherwise, 'multiply'.
+        max_iter: maximum number of times to run phase solvers iteratively to avoid the effect
+            of phase wraps in, e.g. phase_slope_cal or TT_phs_logcal
+        phs_conv_crit: convergence criterion for updates to iterative phase calibration that compares
```
nkern: function isn't specific to phs, so should be `conv_crit`

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        AC_func(**AC_kwargs)
+        gains_here = merge_gains([gf(*gargs) for gf, gargs in zip(gain_funcs, gain_args_list)])
+        apply_cal.calibrate_in_place(AC.data, gains_here, AC.wgts, gain_flags, 
+                                     gain_convention=gain_convention, flags_are_wgts=True)
+        for k in gains_to_update.keys():
+            gains_to_update[k] *= gains_here[k]
+        if max_iter > 1:
+            crit = np.median(np.linalg.norm([gains_here[k] - 1.0 for 
+                                             k in gains_here.keys()], axis=(0, 1)))
+            echo("phase_slope_cal convergence criterion: " + str(crit), verbose=verbose)
```
nkern: put `AC_func` in here instead of `phase..`

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+
+    # instantiate Abscal object
+    wgts = DataContainer({k: (~flags[k]).astype(np.float) for k in flags.keys()})
+    AC = AbsCal(model, data, wgts=wgts, antpos=data.antpos, freqs=data.freqs, 
+                refant=pick_reference_antenna(synthesize_ant_flags(flags))[0],
+                min_bl_cut=min_bl_cut, max_bl_cut=max_bl_cut)
+    AC.antpos = data.antpos
+
+    # Global Delay Slope Calibration
+    for time_avg in [True, False]:
```
nkern: haven't exposed `time_avg` kwarg, and why do you need the conditional?

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        max_bl_cut: maximum baseline separation [meters] to keep in data when calibrating. None (default) means no maximum.
+        edge_cut: integer number of channels to exclude at each band edge in delay and global phase solvers
+        tol: baseline match tolerance in units of baseline vectors (e.g. meters)
+        phs_max_iter: integer maximum number of iterations of phase_slope_cal or TT_phs_cal allowed
+        phs_conv_crit: convergence criterion for updates to iterative phase calibration that compares them to all 1.0s.
+        clobber: if True, overwrites existing abscal calfits file at the output path
+        add_to_history: string to add to history of output abscal file
+
+    Returns:
+        hc: HERACal object which was written to disk. Matches the input redcal_file with an updated history and:
```
nkern: this makes it seem like 5 things are returned, perhaps consider another way of phrasing this w/o making it seem like `gains` `flags` etc are returned

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+                           tol=1.0, phs_max_iter=100, phs_conv_crit=1e-6, clobber=True, add_to_history='', verbose=True):
+    '''Perform abscal on entire data files, picking relevant model_files from a list and doing partial data loading.
+    
+    Arguments:
+        data_file: string path to raw uvh5 visibility file
+        redcal_file: string path to calfits file that redundantly calibrates the data_file
+        model_files: list of string paths to externally calibrated data. Strings must be sortable 
+            to produce a chronological list in LST (wrapping over 2*pi is OK)
+        output_replace: tuple of two strings to find and replace in redcal_file to produce the output calfits file
+        nInt_to_load: number of integrations to load and calibrate simultaneously. Default -1 loads all integrations.
```
nkern: can this be `None` instead of -1, I'm used to `-1` being an axis specification, and also `None` follows `io.HERAData` convention when specifying `bls` `pols` etc..

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+    return abscal_delta_gains, AC
+
+
+def post_redcal_abscal_run(data_file, redcal_file, model_files, output_replace=('.omni.', '.abs.'), nInt_to_load=-1,
+                           data_solar_horizon=90, model_solar_horizon=90, min_bl_cut=1.0, max_bl_cut=None, edge_cut=0, 
+                           tol=1.0, phs_max_iter=100, phs_conv_crit=1e-6, clobber=True, add_to_history='', verbose=True):
+    '''Perform abscal on entire data files, picking relevant model_files from a list and doing partial data loading.
+    
+    Arguments:
+        data_file: string path to raw uvh5 visibility file
```
nkern: why are we limiting ourselves to only `uvh5`? why not make this flexible? One use-case is we want to compare this to IDR2.1 outputs

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        redcal_file: string path to calfits file that redundantly calibrates the data_file
```
nkern: this should be optional, in case the data are already calibrated

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        model_files: list of string paths to externally calibrated data. Strings must be sortable 
+            to produce a chronological list in LST (wrapping over 2*pi is OK)
+        output_replace: tuple of two strings to find and replace in redcal_file to produce the output calfits file
```
nkern: let's just make this more flexible and easier and be `output_fname`

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+            - gains: abscal gains for times that could be calibrated, redcal gains otherwise (but flagged)
+            - flags: redcal flags, with additional flagging if the data or model are flagged (see flag_utils.synthesize_ant_flags)
+            - quals: abscal chi^2 per antenna based on calibrated data minus model (Normalized by noise/nObs, but not with proper DoF)
+            - total_qual: abscal chi^2 based on calibrated data minus model (Normalized by noise/nObs, but not with proper DoF)
+    '''
+    # Raise error if output calfile already exists and clobber is False
+    if os.path.exists(redcal_file.replace(*output_replace)) and not clobber:
+        raise IOError("{} exists, not overwriting.".format(redcal_file.replace(*output_replace)))
+
+    # Load redcal calibration
```
nkern: if fed

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
-        gain_keys = gain_dict.keys()
-
-        # join gain_flag_dict with flag_dict
-        for k in flag_dict:
-            if k in gain_flag_dict:
-                flag_dict[k] += gain_flag_dict[k]
-
-    # make blank gains if no modelfiles
+def get_all_times_and_lsts(hd, solar_horizon=90.0, unwrap=True):
+    '''Extract all times and lsts from a HERAData object
```
nkern: this only works for a `HERAData` instantiated with a `uvh5`, see `io.get_file_lsts` for how to do this for other file formats. Let's just have one function in the codebase to do this functionality, if you want to add `times` in addition to `lsts`, great!

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        min_bl_cut: minimum baseline separation [meters] to keep in data when calibrating. None or 0 means no mininum,
+            which will include autocorrelations in the absolute calibration. Usually this is not desired, so the default is 1.0.
+        max_bl_cut: maximum baseline separation [meters] to keep in data when calibrating. None (default) means no maximum.
+        edge_cut: integer number of channels to exclude at each band edge in delay and global phase solvers
+        tol: baseline match tolerance in units of baseline vectors (e.g. meters)
+        phs_max_iter: integer maximum number of iterations of phase_slope_cal or TT_phs_cal allowed
+        phs_conv_crit: convergence criterion for updates to iterative phase calibration that compares them to all 1.0s.
+        clobber: if True, overwrites existing abscal calfits file at the output path
+        add_to_history: string to add to history of output abscal file
+
```
nkern: should provide `read_kwargs` to pass to `HERAData.read` calls on model and data. example is running only a single pol from multi-pol files

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        hc: HERACal object which was written to disk. Matches the input redcal_file with an updated history and:
+            - gains: abscal gains for times that could be calibrated, redcal gains otherwise (but flagged)
+            - flags: redcal flags, with additional flagging if the data or model are flagged (see flag_utils.synthesize_ant_flags)
+            - quals: abscal chi^2 per antenna based on calibrated data minus model (Normalized by noise/nObs, but not with proper DoF)
+            - total_qual: abscal chi^2 based on calibrated data minus model (Normalized by noise/nObs, but not with proper DoF)
+    '''
+    # Raise error if output calfile already exists and clobber is False
+    if os.path.exists(redcal_file.replace(*output_replace)) and not clobber:
+        raise IOError("{} exists, not overwriting.".format(redcal_file.replace(*output_replace)))
+
```
nkern: partial IO doesn't work for `miriad` files across time, so there should be a check on `nInt == None` if `filetype` == `miriad`

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+                [(rc_flags.keys(), data.antpos), (rc_flags.keys(),)], rc_flags,
+                gain_convention=gain_convention, max_iter=phs_max_iter, phs_conv_crit=phs_conv_crit, verbose=verbose)
+
+    return abscal_delta_gains, AC
+
+
+def post_redcal_abscal_run(data_file, redcal_file, model_files, output_replace=('.omni.', '.abs.'), nInt_to_load=-1,
+                           data_solar_horizon=90, model_solar_horizon=90, min_bl_cut=1.0, max_bl_cut=None, edge_cut=0, 
+                           tol=1.0, phs_max_iter=100, phs_conv_crit=1e-6, clobber=True, add_to_history='', verbose=True):
+    '''Perform abscal on entire data files, picking relevant model_files from a list and doing partial data loading.
```
nkern: might want to mention that this doesn't work for BDA files

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+    # Initialize full-size, totally-flagged abscal gain/flag/etc. dictionaries
+    abscal_gains = copy.deepcopy(rc_gains)
+    abscal_flags = {ant: np.ones_like(rf) for ant, rf in rc_flags.items()}
+    abscal_chisq_per_ant = {ant: np.zeros_like(rq) for ant, rq in rc_quals.items()}
+    abscal_chisq = {pol: np.zeros_like(rtq) for pol, rtq in rc_tot_qual.items()}
+
+    # match times to narrow down model_files
+    matched_model_files = sorted(set(match_times(data_file, model_files, filetype='uvh5')))
+    if len(matched_model_files) > 0:
+        hd = io.HERAData(data_file)
```
nkern: This gets a little tricky for non `uvh5` files b/c `pols` won't be loaded by default. We could do a hack and load a single integration, which would be quick and also load the relevant metadata if `filetype==miriad`

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        hdm = io.HERAData(matched_model_files)
+        pol_load_list = [pol for pol in hd.pols if split_pol(pol)[0] == split_pol(pol)[1]]
```
nkern: update given `polarizations` kwarg

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+                                          lat=hdm.telescope_location_lat_lon_alt_degrees[0], inplace=True)
+                        
+                        # update data flags w/ model flags
+                        for k in flags.keys():
+                            if k in model_flags:
+                                flags[k] += model_flags[k]
+
+                        # apply calibration
+                        data_ants = set([ant for bl in data.keys() for ant in split_bl(bl)])
+                        rc_gains_subset = {k: rc_gains[k][tinds, :] for k in data_ants}
```
nkern: if fed `input_cal`

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+                        rc_flags_subset = {k: rc_flags[k][tinds, :] for k in data_ants}
+                        calibrate_in_place(data, rc_gains_subset, data_flags=flags, 
+                                           cal_flags=rc_flags_subset, gain_convention=hc.gain_convention)
+                        
+                        # run absolute calibration, copying data because it gets modified internally
+                        delta_gains, AC = post_redcal_abscal(model, copy.deepcopy(data), flags, rc_flags_subset, edge_cut=edge_cut,
```
nkern: lets think if this deepcopy is really needed. I don't think it is. Just use `AC.data` and get rid of second `calibrate_in_place`

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+                            dt = (np.median(np.ediff1d(hd.times_by_bl[bl[:2]])) * 86400.)
+                            noise_var = predict_noise_variance_from_autos(bl, data, dt=dt)
+                            chisq_wgts[bl] = noise_var**-1 * (~flags[bl]).astype(np.float)
+                        total_qual, nObs, quals, nObs_per_ant = utils.chisq(AC.data, AC.model, chisq_wgts,
+                                                                            gain_flags=rc_flags_subset, split_by_antpol=True)
+                    
+                        # update results
+                        delta_flags = synthesize_ant_flags(flags)
+                        for ant in data_ants:
+                            abscal_gains[ant][tinds, :] = rc_gains_subset[ant] * delta_gains[ant]
```
nkern: I think you're double counting `rc_gains` here b/c `abscal_gains` was a deepcopy to begin with.

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+        max_bl_cut : float, eliminate all visibilities with baseline separation lengths
+            larger than max_bl_cut. This is assumed to be in ENU coordinates with units of meters.
+        edge_cut : integer number of channels to exclude at each band edge in delay and global phase solvers
+        tol: float distance for baseline match tolerance in units of baseline vectors (e.g. meters)
+        gain_convention: either 'divide' if raw data is calibrated by dividing it by the gains
+            otherwise, 'multiply'.
+        phs_max_iter: maximum number of iterations of phase_slope_cal or TT_phs_cal allowed
+        phs_conv_crit: convergence criterion for updates to iterative phase calibration that compares
+            the updates to all 1.0s. 
+
```
nkern: let's add a `refant` kwargs that is `None` by default which triggers `pick_reference_antenna`

hera_cal/io.py
```diff
@@ -776,6 +776,34 @@ def get_file_lst_range(filepaths, filetype='uvh5', add_int_buffer=False):
     return file_starts, file_stops, int_times
 
 
+def partial_time_io(hd, times, **kwargs):
```
nkern: specify in docstring this must be a `uvh5`-instatiated `HERAData` object

hera_cal/io.py
```diff
@@ -776,6 +776,34 @@ def get_file_lst_range(filepaths, filetype='uvh5', add_int_buffer=False):
...
+
+    Returns:
+        data: DataContainer mapping baseline keys to complex visibility waterfalls
+        flags: DataContainer mapping baseline keys to boolean flag waterfalls
+        nsamples: DataContainer mapping baseline keys to interger Nsamples waterfalls
+        '''
+    combined_hd = None
+    for f in hd.filepaths:
+        hd_here = HERAData(f)
+        times_here = [t for t in times if t in hd_here.times]
```
nkern: can we have it catch the case of files that don't have partial io and thus `hd_here.times == None` and have it just load all times?

hera_cal/abscal.py
```diff
@@ -981,510 +2636,319 @@ def TT_Phi_gain_arr(self):
...
+                        delta_flags = synthesize_ant_flags(flags)
+                        for ant in data_ants:
+                            abscal_gains[ant][tinds, :] = rc_gains_subset[ant] * delta_gains[ant]
+                            abscal_flags[ant][tinds, :] = rc_flags_subset[ant] + delta_flags[ant]
+                            abscal_chisq_per_ant[ant][tinds, :] = quals[ant] / nObs_per_ant[ant]  # Note, not normalized for DoF
+                        for antpol in total_qual.keys():
+                            abscal_chisq[antpol][tinds, :] = total_qual[antpol] / nObs[antpol]  # Note, not normalized for DoF
+                            
+        # impose a single reference antenna on the final antenna solution
+        refant = pick_reference_antenna(abscal_flags)
```
nkern: use `refant` if fed

nkern:
A really great addition to `abscal`, making it partial IO capable across time, and significantly modularized the `abscal_run` drive function. Thanks!

I've posted some minor comments, and I have one major issue, which is that I think the `post_redcal_abscal` function is a *little* bit too specific for IDR2.2 run setup. I'm fine w/ making it specific in the sense of which `abscal` functions to run, hence `post_redcal_abscal` name, but even the inputs and outputs assumes the exact handoff of data products as we will do them in IDR2.2, and I think that's a little too rigid. I think we can make this better w/o making it more complicated by:

1) allowing more than just `uvh5` inputs, which takes a little bit of massaging for some of the sub-functions, and in some cases eliminates the partial io capability, which is fine b/c we aren't doing that anyways w/ `miriad` files

2) allowing for `input_cal` optionality, instead of rigidity.

3) allowing for `output_fname` instead of assuming that the output filename is an input calfits w/ a replaced keyword

If we were able to get those 3 things in, I would feel more comfortable w/ us erasing the old `abscal_run`, which is part of this PR.

Great work so far!

hera_cal/abscal.py

hera_cal/abscal.py
jsdillon: Actually, the way I wrote it, it's very specific to phase. The non-phase steps don't suffer from phase wraps, which was the original motivation for making them iterative (unlike redcal, where the iterative steps address the fundamentally non-linear nature of redcal).

hera_cal/abscal.py
jsdillon: Good call. Done.

hera_cal/abscal.py
jsdillon: I thought the idea was to get an approximate global delay slope (assuming it doesn't change) by doing the average, then do it again for each integration. That's exactly what the old abscal run did.

jsdillon:
Thanks so much @nkern for the quick review. I agree with most of your comments and I think I can implement most of this pretty straightforwardly. 

In the interest of time, I think I'm going to punt on non-uvh5 files. That's a bunch more work and in the interest of getting IDR2.2 done, I want to get this merged in ASAP. Also, redcal is uvh5-only. Converting between data formats is so easy with pyvudata, I imagine in most cases that's a better option than building in a lot more legacy structure. However, if you want to open an issue to update the post_redcal_abscal functions to support more file formats, go right ahead.

hera_cal/abscal.py

hera_cal/abscal.py
jsdillon: I added refant.

I don't think exposing that kwarg is necessary. All I'm looking for is the least flagged antenna. If everything is flagged in synthesize_ant_flags somehow, it doesn't much matter.

hera_cal/abscal.py
jsdillon: Done. I also updated redcal to have a consistent convention.

hera_cal/abscal.py
jsdillon: I'd like to punt in more data formats. See below.

hera_cal/io.py
jsdillon: It's not clear that that's what we want it to do. I'll add an assertion instead.

hera_cal/abscal.py
jsdillon: The issue here is that I need calibrated autocorrelations to normalize chi^2... which are often deleted from AC.data. However, I think I can solve the issue in a memory-efficient way by deepcopying the autos before creating the Abscal object, rather than deepcopying the whole dataset. Thanks for making me think it through.

hera_cal/abscal.py
jsdillon: No, because I'm doing = rather than *=, so I'm overwriting.

jsdillon:
Ok. I've addressed most of your comments. However, I've given it some more thought and it's not trivial to support pre-calibrated data. This code relies on the redcal_file to do a number of things:

1. Initializing empty dictionaries for holding the abscal results that are then 
2. Holding antenna-based flags used for calibrate_in_place
3. Specifying the gain convention.
4. Providing all the correct metadata for writing the results to disk with the appropriate history.

It's certainly possible to bypass all of that, but I don't think that extra flexibility is worth complicating the control structure of the code and incurring more overhead in terms of unit testing and maintainability. The more if statements, the more unit tests we need. 

I especially don't think it's worth holding up the PR for extra flexibility that we don't think we need right now and really shouldn't be tempted to use, since we're trying to save disk space by not re-rewriting calibrated data. After all, "There should be one--and preferably only one--obvious way to do it."

hera_cal/abscal.py
jsdillon: I'd like to punt on this too. See https://github.com/HERA-Team/hera_cal/pull/430#issuecomment-457373546

hera_cal/abscal.py
jsdillon: I'd like to punt on this too. See #430 (comment)

hera_cal/abscal.py
jsdillon: Again, I think this is something to punt on and open as an issue if it ever becomes useful or necessary.

It's fairly trivial to rewrite your redcal and data files with a pyuvdata select before running this function. As it stands, if the number of polarizations to run on is less than what's in either data or redcal file, it's not obvious how it should behave. Again, more complexity than I think is likely to be useful.

hera_cal/abscal.py
jsdillon: See https://github.com/HERA-Team/hera_cal/pull/430#issuecomment-457373546

hera_cal/abscal.py
jsdillon: See https://github.com/HERA-Team/hera_cal/pull/430#discussion_r250795758

PR #433, opened 2019-01-28
initialize clean_info as a dict, not as a DataContainer

jsdillon:


jsdillon:
@adampbeardsley can you test this branch and see if it fixes your problem?

jsdillon:
All the other objects are visibilities, so conjugating them if you ask for the conjugate key is the right thing to do.

adampbeardsley:
Oh, I just saw what I was missing - Line 132 of `datacontainer.py` conjugates _and_ reverses the key. Ok, that part makes sense.

jsdillon:
For posterity: this was related to this thread: https://eoranalysis.slack.com/archives/C65UTUHPG/p1548291055052400

PR #442, opened 2019-02-19
Fix reference antenna and degeneracy issues from hera_cal validation

jsdillon:
This function fixes a couple of issues in trying to validate the combination of redcal and abscal.

First, redcal and abscal were not quite solving for the same degeneracies. Redcal was using idealized antenna positions solved for from the reds. Abscal was using the actual antenna positions. This made it impossible for the answers to agree exactly.

Second, it turns out that when redcal and abscal both operate on 'xx' and 'yy' independently, there are two overall phase degeneracies that cannot be fixed by rephasing to a single reference antenna that applies to both polarizations (as we previously thought). I have therefore added the ability to pick one reference antenna per polarization and implemented it in abscal and smoothcal. The combination of both of these should allow us to get perfect matches between the simulation and the calibration results.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -39,7 +39,7 @@ import linsolve
 
 from . import version
 from .apply_cal import calibrate_in_place
-from .smooth_cal import pick_reference_antenna
+from .smooth_cal import pick_reference_antenna, rephase_to_refant
 from .flag_utils import synthesize_ant_flags
 from .noise import predict_noise_variance_from_autos
 from . import utils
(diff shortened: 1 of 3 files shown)


hera_cal/abscal.py
```diff
@@ -2771,21 +2771,22 @@ def post_redcal_abscal(model, data, flags, rc_flags, min_bl_cut=None, max_bl_cut
...
     if refant_num is None:
-        refant_num = pick_reference_antenna(abscal_delta_gains, synthesize_ant_flags(flags), data.freqs)[0]
+        refant_num = pick_reference_antenna(abscal_delta_gains, synthesize_ant_flags(flags), data.freqs, per_pol=False)[0]
     wgts = DataContainer({k: (~flags[k]).astype(np.float) for k in flags.keys()})
-    AC = AbsCal(model, data, wgts=wgts, antpos=data.antpos, freqs=data.freqs,
+    idealized_antpos = redcal.reds_to_antpos(redcal.get_reds(data.antpos, bl_error_tol=tol))
+    AC = AbsCal(model, data, wgts=wgts, antpos=idealized_antpos, freqs=data.freqs,
                 refant=refant_num, min_bl_cut=min_bl_cut, max_bl_cut=max_bl_cut)
-    AC.antpos = data.antpos
+    AC.antpos = idealized_antpos
```
nkern: so is the solution to use the idealized antpos over the actual antpos? seems like it should be the other way, no?

jsdillon: The actual antpos aren't redundant, so the degeneracies don't really make sense with the actual antpos.

PR #444, opened 2019-03-07
Significantly improve fft_dly and fix phase offset solving

### LINKED ISSUE #141: fft_dly of abscal needs refinement at 0.5 ns level
in noise-free simulation, `fft_dly` results deviate from truth at the 0.5 ns level. the algorithm can be refined to resolve this
fft_dly

### LINKED ISSUE #445: Fix calculation of frequencies in fft_dly for correct offsets
The current algorithm for finding offsets, which averages the angle of the ratio of the data to the data rephased by the delay solved for does not use the correct set of frequencies and therefore does not find the correct offset. If we want to know the correct offset, we need need to supply the frequency of the first bin so that the following line

https://github.com/HERA-Team/hera_cal/blob/8b1537baf96878cbf919096d5fcfff6b0c89c5b7/hera_cal/utils.py#L145

becomes

`freqs = np.arange(Nfreqs, dtype=data.dtype) * df + f0`

jsdillon:
As mentioned in https://github.com/HERA-Team/hera_cal/issues/141#issuecomment-470417003, I found a much better way to estimate delays using an improved peak-finding algorithm. I've implemented that.

In the process of doing that, I found a bug in how phase offsets were calculated and fixed it (see #445).

It's not 100% obvious whether we want to include phase offsets in firstcal (we currently don't, but at least now we can), so I haven't modified redcal substantially. That's an issue I'll look into later in the context of validation. 

This closes #141 and closes #445.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -384,7 +384,7 @@ def phs_logcal(model, data, wgts=None, refant=None, verbose=True):
     return fit
 
 
-def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, solve_offsets=True, medfilt=True,
+def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, f0=0., solve_offsets=True, medfilt=True,
                  kernel=(1, 5), verbose=True, antpos=None, four_pol=False, edge_cut=0):
     """
     Solve for per-antenna delays according to the equation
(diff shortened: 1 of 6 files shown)


nkern:
Thanks for catching that bug: I think it was actually introduced in a recent PR, and wasn't part of the original fft_dly algorithm. Anyways, good to see that its fixed.

Also, see my comment on #141 : probably you just flipped the legend in your plots, but wanted to confirm that with you.

Also, while I'm sure quinn's second method does a good job in this context, could you make it possible to do either the quadratic method or quinn's second method via a kwarg: interp_method='quinn', which could by default use the new algorithm?

I don't think we need to expose this kwarg to `fft_dly` or to `dly_linsolve` etc, but it would be nice to be able to manually go in and use the quadratic method for testing purposes etc. if we find that in the future quinn's method isn't behaving as well for other contexts.

jsdillon:
Thanks @nkern!

PR #447, opened 2019-03-08
Fix flag_nchan_high/low in redcal with no partial I/O

jsdillon:
In the case where redcal is performed without partial I/O but edge channels are flagged, we get an dimension mismatch error. This fixes the error and adds a test to expose it.


Chuneeta:
@jsdillon looks fine to me.

jsdillon:
Thanks @Chuneeta!

PR #453, opened 2019-03-22
Phase offset solving in first cal and DFT-based phase slope solving in abscal

### LINKED ISSUE #451: Idealized antenna positions in abscal break min_bl_cut and max_bl_cut
https://github.com/HERA-Team/hera_cal/blob/c6e36f1752b63efae76336710292f7f4e1d253aa/hera_cal/abscal.py#L2779

This is because our algorithm arbitrarily rescales the antenna positions.

### LINKED ISSUE #452: Deprecate abscal.apply_gains
https://github.com/HERA-Team/hera_cal/blob/992ed83deec0ea2341d2744ad69de55296f958d8/hera_cal/abscal.py#L801

It is no longer used.

### LINKED ISSUE #454: Spectral Structure in Omnical Solutions due to lack of Firstcal Offset Solving
In our most recent attempt at running IDR2.2, I found considerable and worrisome new spectral structure introduced into our pipeline relative to IDR2.1.

Looking at `abscal` phases over a couple of hours, we found:
Screen Shot 2019-03-23 at 3 08 53 PM

We did not observe anything like this in IDR2.1:
Screen Shot 2019-03-23 at 3 09 00 PM

I believe this is traceable to issues in `redcal` and in particular to `firstcal`. In IDR2.1, we employed an iterative algorithm in `firstcal` to identify antennas there were 180 degrees rotated, effectively introducing a pi phase offset. This step was removed from the recent `firstcal` overhaul because we didn't think it was necessary. I now believe that it is (which is why I'm writing all this up!).

Digging into `redcal` results from this failed IDR2.2 candidate, we found evidence for complicated spectral structure:
Screen Shot 2019-03-23 at 3 14 16 PM

Now compare this to IDR2.1:
Screen Shot 2019-03-23 at 3 14 24 PM

I believe that this is due to bad `logcal` or `omnical` solutions that result from `firstcal` not getting close enough to the truth. We see something similar in simulations. For example, here is a simulated 19 element array where the gains are perfectly characterized by a single delay and offset (and all have amplitude of 1): 
Screen Shot 2019-03-23 at 3 27 02 PM

Using `redcal` on the current master branch, we get a very similar pathology:
Screen Shot 2019-03-23 at 3 23 03 PM
The error in phase and delay is expected, since both of those terms are affected by degeneracies, and can be fixed later in `abscal`. However, the spectral structure is not expected. Moreover, if we force the phase offsets (phi) to be small (rather than letting them range from 0 to 2pi), the effect goes away. It therefore appears necessary to add the ability for `firstcal` to solve for phases offsets in addition to delays.

This is what I'm doing in #453.

jsdillon:
As I argued in #454, we are observing considerable spectral structure in the redcal solutions on the current master branch which I attribute to the lack of the ability to calibrate firstcal phase offsets. This PR adds that ability.

The basic idea is that firstcal solves for gains of the form
Screen Shot 2019-03-23 at 3 27 02 PM
using a system of equations that look like
Screen Shot 2019-03-23 at 3 57 47 PM
This works because Vij and Vkl are different baselines in a group of nominally redundant baselines, so the true visibility terms cancel and the overall bandpass is taken out by the absolute value in the denominator.

This works nicely for tau, which does not suffer phase wraps, but not so nicely for phi. That particular combination of four phis is only measured for each visibility combination up to some multiple of 2pi. However, my simulations have revealed that have solving for delays and then iteratively solving for delay offsets results in phase solutions that are correct up to an overall phase and phase slope across the array (i.e. the redundant calibration degeneracies). 

This fixes the problem identified in #454. Here are the redcal gains before this fix:
Screen Shot 2019-03-23 at 3 23 03 PM

And here they are after:
Screen Shot 2019-03-23 at 4 22 41 PM

However, it turns out that this fix was not quite sufficient to get the combination of abscal + redcal to exactly reproduce the input gains (when an arbitrary per-antenna phase offset was introduced) up to an overall (irrelevant) phase. The problem lied in the step for solving the global phase slope, which could handle small amounts of phase wrapping across the array, but not a lot (#257). It's unclear whether this will be a problem in real data, but this seemed like the proper time to address the issue.

To fit global phase slopes, we looked at the DFT of the data/model ratio at various points in phase slope space (which can be thought of as a tilt/orientation of the array and therefore a Fourier dual to position). Finding the peak in this space gives us the overall phase slope. This is effectively the 2D, non-regularly spaced generalization of `utils.fft_dly`. This was implemented fairly generally in `abscal.dft_phase_slope_solver` and then `abscal.global_phase_slope_logcal` was adapted to optionally use that method instead of `linsolve.` The results of this are quite nice:
Screen Shot 2019-03-23 at 4 35 49 PM

This allows me to produce `hera_cal` gain solutions that match the input gains for arbitrary phase offsets up to machine precision, which will the subject of a forthcoming HERA Analysis Team / Validation memo (@jaguirre).

I'll note that my preliminary tests of this PR on real data have been quite promising. In #454 I showed some omnical gains in IDR2.1 and in the latest attempt at IDR2.2. The latter showed considerable spectral structure. Here's the result running a single file with this branch:
Screen Shot 2019-03-23 at 4 55 13 PM
Looks just like IDR2.1, except now our algorithm is considerably more robust. This therefore closes #454.

Finally, this PR address a couple of issues that came up along the way. Namely,
* Abscal still had a proprietary apply_gains function, but it was not being used. This PR deprecates it. This closes #452.
* The idealized antenna positions used in abscal to fix the degeneracies of redcal, which we introduced in #442, created problems for both cuts based on longest or shortest baselines and for determination of redundancies using the `tol`/`bl_error_tol` parameter for not-quite redundant baselines that are supposed to be treated as redundant. This PR handles that more carefully, introducing the idealized antenna positions derived from the reds later and not passing through the `tol` parameter once they've been introduced. This closes #451.


jsdillon:
Ok, @AaronParsons, I'm ready for your review. Everything I've been working on over the past couple of weeks is written up here and in #454 for posterity.

hera_cal/abscal.py
```diff
@@ -213,7 +214,7 @@ def TT_phs_logcal(model, data, antpos, wgts=None, refant=None, verbose=True, zer
         eqns = odict([(k, "psi_{}*a1 - psi_{}*a2 + Phi_ew_{}*{} + Phi_ns_{}*{} - Phi_ew_{}*{} - Phi_ns_{}*{}"
                        "".format(split_pol(k[2])[0], split_pol(k[2])[1], split_pol(k[2])[0],
                                  r_ew[k[0]], split_pol(k[2])[0], r_ns[k[0]], split_pol(k[2])[1],
-                                 r_ew[k[1]], k[2][1], r_ns[k[1]])) for i, k in enumerate(keys)])
+                                 r_ew[k[1]], split_pol(k[2])[1], r_ns[k[1]])) for i, k in enumerate(keys)])
```
AaronParsons: Could use "for i, (k1,k2) in enumerate(keys)" to make this splitting less obtuse.

hera_cal/abscal.py
```diff
@@ -668,7 +669,49 @@ def delay_slope_lincal(model, data, antpos, wgts=None, refant=None, df=9.765625e
     return fit
 
 
-def global_phase_slope_logcal(model, data, antpos, wgts=None, refant=None, verbose=True, tol=1.0, edge_cut=0):
+def dft_phase_slope_solver(xs, ys, data):
+    '''Solve for sptial phase slopes across an array by looking for the peak in the DFT.
```
AaronParsons: "spatial"

hera_cal/abscal.py
```diff
@@ -668,7 +669,49 @@ def delay_slope_lincal(model, data, antpos, wgts=None, refant=None, df=9.765625e
...
+            xs and ys, but subsequent dimensions will be preserved and solved independently. 
+            Any np.nan in data is interpreted as a flag.
+
+    Returns:
+        slope_x, slope_y: phase slopes in units of 1/[xs] where the best fit phase slope plane
+            is np.exp(2.0j * np.pi * (xs * slope_x + ys * slope_y)). Both have the same shape 
+            the data after collapsing along the first dimension.
+    '''
+
+    # find the range of k values in DFT space to explore and the appropriate sampling
```
AaronParsons: can't parse this comment

hera_cal/abscal.py
```diff
@@ -668,7 +669,49 @@ def delay_slope_lincal(model, data, antpos, wgts=None, refant=None, df=9.765625e
...
+        return -np.abs(np.dot(z, np.exp(-2.0j * np.pi * (x * k[0] + y * k[1]))))
+
+    # loop over data, minimizing the cost function
+    dflat = data.reshape((len(xs), -1))
+    slope_x = np.zeros_like(dflat[0, :].real)
+    slope_y = np.zeros_like(dflat[0, :].real)
+    for i in range(dflat.shape[1]):
+        if not np.all(np.isnan(dflat[:, i])):
+            dft_peak = brute(dft_abs, (search_slice, search_slice), 
+                             (xs[~np.isnan(dflat[:, i])], ys[~np.isnan(dflat[:, i])],
```
AaronParsons: suggest not recomputing isnan 3 times

hera_cal/abscal.py
```diff
@@ -668,7 +669,49 @@ def delay_slope_lincal(model, data, antpos, wgts=None, refant=None, df=9.765625e
...
+    '''Solve for sptial phase slopes across an array by looking for the peak in the DFT.
+    This is analogous to the method in utils.fft_dly(), except its in 2D and does not 
+    assume a regular grid for xs and ys.
+
+    Arguments:
+        xs: 1D array of x positions (e.g. of antennas or baselines)
+        ys: 1D array of y positions (must be same length as xs)
+        data: ndarray of complex numbers to fit with a phase slope. The first dimension must match 
+            xs and ys, but subsequent dimensions will be preserved and solved independently. 
+            Any np.nan in data is interpreted as a flag.
```
AaronParsons: Are we requiring all functions to be able to handle nans?  Or should the user be required to provide non-nan inputs?

hera_cal/abscal.py
```diff
@@ -709,7 +755,13 @@ def global_phase_slope_logcal(model, data, antpos, wgts=None, refant=None, verbo
...
           for each position component and polarization [radians / meter].
     """
-    echo("...configuring linsolve data for global_phase_slope_logcal", verbose=verbose)
+    # check solver and edgecut
+    if solver == 'linfit':
+        echo("...configuring linsolve data for global_phase_slope_logcal", verbose=verbose)
+    elif solver == 'dft':
+        echo("...finding global phase slopes using the DFT method", verbose=verbose)
+    else:
+        raise ValueError("Unrecognized solver {}. Must be either 'linfit' or 'dft'.".format(solver))
```
AaronParsons: Would prefer an "assert solver in GLOBAL_LIST_OF_SOLVERS" and an echo that is not solver-dependent

hera_cal/abscal.py
```diff
@@ -735,39 +787,60 @@ def global_phase_slope_logcal(model, data, antpos, wgts=None, refant=None, verbo
...
+        dm_ratio /= np.abs(dm_ratio)  # This gives all channels roughly equal weight, moderating the effect of RFI (as in firstcal)
         binary_flgs = np.isclose(avg_wgts[rk], 0.0)
-        delta_phi[binary_flgs] *= np.nan
-        avg_wgts[rk][np.isinf(delta_phi) + np.isnan(delta_phi)] = 0.0
-        delta_phi[np.isinf(delta_phi) + np.isnan(delta_phi)] *= np.nan
-        ls_data[eqn_str] = np.nanmedian(delta_phi[:, edge_cut:(delta_phi.shape[1] - edge_cut)], axis=1, keepdims=True)
-        ls_wgts[eqn_str] = np.sum(avg_wgts[rk][:, edge_cut:(delta_phi.shape[1] - edge_cut)], axis=1, keepdims=True)
+        dm_ratio[binary_flgs] *= np.nan
+        avg_wgts[rk][np.isinf(dm_ratio) + np.isnan(dm_ratio)] = 0.0
+        dm_ratio[np.isinf(dm_ratio) + np.isnan(dm_ratio)] *= np.nan
```
AaronParsons: put the two above lines before the computation of binary_flgs (i.e. flag once)

hera_cal/abscal.py
```diff
@@ -735,39 +787,60 @@ def global_phase_slope_logcal(model, data, antpos, wgts=None, refant=None, verbo
...
-    solver = linsolve.LinearSolver(ls_data, wgts=ls_wgts)
-    echo("...running linsolve", verbose=verbose)
-    fit = solver.solve()
-    echo("...finished linsolve", verbose=verbose)
+    if solver == 'linfit':  # build linear system for phase slopes and solve with linsolve
+        # setup linsolve and run
+        solver = linsolve.LinearSolver(ls_data, wgts=ls_wgts)
+        echo("...running linsolve", verbose=verbose)
+        fit = solver.solve()
+        echo("...finished linsolve", verbose=verbose)
```
AaronParsons: All of these "if linfit, elif dft" switches are making me wonder if these shouldn't just be two separate subclasses, one that does linfit, one that does dft

hera_cal/abscal.py
```diff
@@ -1922,6 +1921,19 @@ def taper(ratio):
             for k in self.wgts.keys():
                 self.wgts[k] *= taper(np.linalg.norm(self.bls[k]) / bl_taper_fwhm)
 
+    def _overwrite_antpos(self, antpos):
```
AaronParsons: prefer "set_antpos"

hera_cal/abscal.py
```diff
@@ -1922,6 +1921,19 @@ def taper(ratio):
...
+        '''Helper function for replacing self.antpos, self.bls, and self.antpos_arr without affecting tapering or baseline cuts.
+        Useful for replacing true antenna positions with idealized ones derived from the redundancies.'''
+        self.antpos = antpos
+        self.antpos_arr = None
+        self.bls = None
+        if self.antpos is not None:
+            # center antpos about reference antenna
+            self.antpos = odict(list(map(lambda k: (k, antpos[k] - antpos[self.refant]), self.ants)))
```
AaronParsons: make a list comprehension

hera_cal/abscal.py
```diff
@@ -1922,6 +1921,19 @@ def taper(ratio):
...
+            self.bls = odict([(x, self.antpos[x[0]] - self.antpos[x[1]]) for x in self.keys])
+            self.antpos_arr = np.array(list(map(lambda x: self.antpos[x], self.ants)))
```
AaronParsons: make a list comprehension

hera_cal/abscal.py
```diff
@@ -2775,26 +2791,31 @@ def post_redcal_abscal(model, data, flags, rc_flags, min_bl_cut=None, max_bl_cut
...
     # Global Delay Slope Calibration
     for time_avg in [True, False]:
         abscal_step(abscal_delta_gains, AC, AC.delay_slope_lincal, {'time_avg': time_avg, 'edge_cut': edge_cut, 'verbose': verbose},
                     [AC.custom_dly_slope_gain], [(rc_flags.keys(), idealized_antpos)], rc_flags,
                     gain_convention=gain_convention, verbose=verbose)
 
-    # Global Phase Slope Calibration
-    abscal_step(abscal_delta_gains, AC, AC.global_phase_slope_logcal, {'tol': tol, 'edge_cut': edge_cut, 'verbose': verbose},
+    # Global Phase Slope Calibration (first using dft, then using linfit)
+    abscal_step(abscal_delta_gains, AC, AC.global_phase_slope_logcal, {'solver': 'dft', 'tol': 1e-8,
```
AaronParsons: Magic "1e-8" appears out of nowhere

hera_cal/abscal.py
```diff
@@ -2775,26 +2791,31 @@ def post_redcal_abscal(model, data, flags, rc_flags, min_bl_cut=None, max_bl_cut
...
+                'edge_cut': edge_cut, 'verbose': verbose}, [AC.custom_phs_slope_gain], [(rc_flags.keys(), idealized_antpos)], 
+                rc_flags, gain_convention=gain_convention, verbose=verbose)
+    abscal_step(abscal_delta_gains, AC, AC.global_phase_slope_logcal, {'tol': 1e-8, 'edge_cut': edge_cut, 'verbose': verbose},
```
AaronParsons: Magic "1e-8" appears out of nowhere

hera_cal/redcal.py
```diff
@@ -569,29 +569,23 @@ def compute_ubls(self, data, gain_sols):
...
         return ubl_sols
 
-    def firstcal(self, data, df, wgts={}, sparse=False, mode='default', norm=True, medfilt=False, kernel=(1, 11)):
-        """Solves for a per-antenna delay by fitting a line to the phase difference between
-        nominally redundant measurements.  To turn these delays into gains, you need to do:
-        np.exp(2j * np.pi * delay * freqs)
-
-        Args:
-            data: visibility data in the dictionary format {(ant1,ant2,pol): np.array}
+    def _firstcal_iteration(self, data, df, f0=0.0, wgts={}, offsets_only=False,
```
AaronParsons: Why is f0=0.0 the default value?

hera_cal/redcal.py
```diff
@@ -569,29 +569,23 @@ def compute_ubls(self, data, gain_sols):
...
-            sparse: represent the A matrix (visibilities to parameters) sparsely in linsolve
-            mode: solving mode passed to the linsolve linear solver ('default', 'lsqr', 'pinv', or 'solve')
-                Suggest using 'default' unless solver is having stability (convergence) problems.
-                More documentation of modes in linsolve.LinearSolver.solve().
-            norm: calculate delays from just the phase information (not the amplitude) of the data.
-                This is a pretty effective way to get reliable delay even in the presence of RFI.
-            medfilt : boolean, median filter data before fft.  This can work for data containing
-                unflagged RFI, but tends to be less effective in practice than 'norm'.  Default False.
-            kernel : size of median filter kernel along (time, freq) axes
+            f0: frequency of the first channel in the data
```
AaronParsons: Do we still need f0 if solving for offsets?

hera_cal/redcal.py
```diff
@@ -618,8 +612,61 @@ def firstcal(self, data, df, wgts={}, sparse=False, mode='default', norm=True, m
...
-        sol = {self.unpack_sol_key(k): v[0] for k, v in sol.items()}  # ignoring offset
-        return sol
+        dly_sol = {self.unpack_sol_key(k): v[0] for k, v in sol.items()}
+        off_sol = {self.unpack_sol_key(k): v[1] for k, v in sol.items()}
+        return dly_sol, off_sol
+
+    def firstcal(self, data, freqs, wgts={}, maxiter=25, conv_crit=1e-6,
+                 sparse=False, mode='default', norm=True, medfilt=False, kernel=(1, 11)):
+        """Solve for a calibration solution parameterized by a single delay and phase offset
+        per antenna using the phase difference between nominally redudant measurements.
```
AaronParsons: "redundant"

hera_cal/redcal.py
```diff
@@ -618,8 +612,61 @@ def firstcal(self, data, df, wgts={}, sparse=False, mode='default', norm=True, m
...
+                {(index, antpol): np.exp(2j * np.pi * delay * freqs + 1j * offset)} format.
+        """
+        # Run firstcal with both delay and offset solving:
+        df = np.median(np.ediff1d(freqs))
+        dtype = np.find_common_type([d.dtype for d in data.values()], [])
+        dly_fc, off_fc = self._firstcal_iteration(data, df=df, f0=freqs[0], wgts=wgts, sparse=sparse, 
+                                                  mode=mode, norm=norm, medfilt=medfilt, kernel=kernel)
+        g_fc = {ant: np.array(np.exp(2j * np.pi * np.outer(dly, freqs) + 1.0j * off_fc[ant]), 
+                              dtype=dtype) for ant, dly in dly_fc.items()}
+        calibrate_in_place(data, g_fc, gain_convention='divide')  # applies calibration
```
AaronParsons: could "in_place" be an argument to calibrate?

hera_cal/redcal.py
```diff
@@ -618,8 +612,61 @@ def firstcal(self, data, df, wgts={}, sparse=False, mode='default', norm=True, m
...
+            kernel : size of median filter kernel along (time, freq) axes
+
+        Returns:
+            g_fc: dictionary of Ntimes x Nfreqs per-antenna gains solutions in the 
+                {(index, antpol): np.exp(2j * np.pi * delay * freqs + 1j * offset)} format.
+        """
+        # Run firstcal with both delay and offset solving:
+        df = np.median(np.ediff1d(freqs))
+        dtype = np.find_common_type([d.dtype for d in data.values()], [])
+        dly_fc, off_fc = self._firstcal_iteration(data, df=df, f0=freqs[0], wgts=wgts, sparse=sparse,
```
AaronParsons: Why have a first iteration outside the for loop?  Can this be avoided?

hera_cal/redcal.py
```diff
@@ -1178,8 +1232,10 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
             this altitude, calibration is skipped and the integrations are flagged.
         flag_nchan_low: integer number of channels at the low frequency end of the band to always flag (default 0)
         flag_nchan_high: integer number of channels at the high frequency end of the band to always flag (default 0)
-        conv_crit: maximum allowed relative change in omnical solutions for convergence
-        maxiter: maximum number of omnical iterations allowed before it gives up
+        fc_conv_crit: maximum allowed changed in firstcal phases for convergence
+        fc_maxiter: maximum number of firstcal iterations allowed for finding per-antenna phases
+        oc_conv_crit: maximum allowed relative change in omnical solutions for convergence
+        oc_maxiter: maximum number of omnical iterations allowed before it gives up
```
AaronParsons: Crazy idea: could parameter docstrings be built algorithmically to avoid having these same strings show up in a file 30 times?

hera_cal/tests/test_abscal.py
```diff
@@ -234,6 +234,28 @@ def test_cut_bl(self):
...
+        xs = np.zeros(100)
+        ys = np.zeros(100)
+        i = 0
+        for x in np.arange(0, 100, 10):
+            for y in np.arange(0, 100, 10):
+                xs[i] = x + 5 * (.5 - np.random.rand())
+                ys[i] = y + 5 * (.5 - np.random.rand())
+                i += 1
+
+        phase_slopes_x = (.2 * np.random.rand(5, 2) - .1)  # not too many phase wraps over the array
```
AaronParsons: what defines too many?

hera_cal/redcal.py
jsdillon: Sadly, no.

hera_cal/abscal.py

hera_cal/abscal.py
jsdillon: Fixed.

hera_cal/abscal.py
jsdillon: Fixed.

hera_cal/abscal.py
jsdillon: Added the ability for this function to take boolean flags.

hera_cal/abscal.py

hera_cal/abscal.py

hera_cal/abscal.py

hera_cal/abscal.py
jsdillon: It's a toss up, but a fair bit of the calculation gets used multiple times. This also makes it easier to use the Abscal.global_phase_slope_logcal and Abscal.custom_phs_slope_gain, since it enforces a universal interface.

hera_cal/abscal.py

hera_cal/abscal.py

hera_cal/abscal.py
jsdillon: Fixed and explained.

hera_cal/abscal.py
jsdillon: Fixed and explained.

hera_cal/abscal.py

hera_cal/redcal.py
jsdillon: Because it's the default value in fft_dly (which was previously missing any f0 before I fixed it) and I feel like it's confusing to have different default values for passthrough kwargs.

I can't think of a better default value. The only real alternative is to pass in the whole of `freqs` and make this function pick out f0.

hera_cal/redcal.py
jsdillon: Yes. We need it for fft_dly.

hera_cal/redcal.py
jsdillon: fixed

hera_cal/redcal.py
jsdillon: In theory, yes. But this is more memory efficient and it's only a trivial extra line (deepcopy) to do not in place.

I think I prefer that extra line than having calibrate either return `calibrated_data` or `None` depending on the kwargs.

hera_cal/redcal.py
jsdillon: Yes, it can. Done.

hera_cal/tests/test_abscal.py
jsdillon: If the phase wraps are so fast that we go through ~2 pi between grid points, the algorithm breaks down. It's basically an undersampling issue.

I clarified this in the comments.

jsdillon:
ok @AaronParsons ready for you to take a look at

jsdillon:
@AaronParsons you ok with this PR? I'd like to start a new run today or tomorrow

PR #468, opened 2019-05-07
Update extract_hh.py script

plaplant:
This PR updates the `extract_hh.py` script for the H2C data coming from the correlator. The main difference is that the `st_type` variable is no longer saved in the `extra_keywords`, and instead the correlator uses antenna numbers less than 350 to signify valid data in the files. I've been using this script behind the scenes to do the data reduction so far, so I thought I'd make it officially part of a repo.

### DIFF
--- a/scripts/extract_hh.py
+++ b/scripts/extract_hh.py
@@ -1,69 +1,47 @@
-#! /usr/bin/env python
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright 2018 the HERA Project
+# Copyright 2019 The HERA Collaboration
 # Licensed under the MIT License
 
+from __future__ import print_function, division, absolute_import
 import os
-import argparse
-import pyuvdata
-from hera_cal.version import version_info
-import pyuvdata.utils as uvutils
+import sys
 import numpy as np
+import argparse
+from pyuvdata import UVData
+from pyuvdata.uvh5 import _hera_corr_dtype
 
-parser = argparse.ArgumentParser(description='Extract HERA hex antennas from data '
-                                 'file, and save with new extension.')
-parser.add_argument('--extension', type=str, help='Extension to be appended to '
-                    'filename for output. Default="HH".', default='HH')
-parser.add_argument('--filetype', type=str, help='Input and output file type. '
-                    'Allowed values are "miriad" (default), and "uvfits".',
-                    default='miriad')
-parser.add_argument('--fixuvws', action='store_true', help='Optional flag to '
-                    'use antenna positions to replace uvws.')
-parser.add_argument('--overwrite', action='store_true', default=False,
-                    help='Optional flag to overwrite output file if it exists.')
-parser.add_argument('files', metavar='files', type=str, nargs='+',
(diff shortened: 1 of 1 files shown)


scripts/extract_hh.py
```diff
@@ -1,69 +1,47 @@
...
-parser.add_argument('--ex_ants_file', type=str, help='Text file with list of antennas'
-                    ' which are excluded downstream in RTP. Generally, these are '
-                    'antennas which are being actively commissioned, or known as bad.'
-                    ' Note these values are only stored in the history, not actually '
-                    'flagged at this step.',
-                    default=None)
-args = parser.parse_args()
+# figure out which antennas have valid data and perform select-on-read
+ant_nums = np.unique(uvd.ant_1_array)
+inds = np.where(ant_nums < 350)
```
jsdillon: The number 350 seems like it should be an optional command line argument.

scripts/extract_hh.py
```diff
@@ -1,69 +1,47 @@
...
-                    help='Files to be processed.')
-parser.add_argument('--ex_ants_file', type=str, help='Text file with list of antennas'
-                    ' which are excluded downstream in RTP. Generally, these are '
-                    'antennas which are being actively commissioned, or known as bad.'
-                    ' Note these values are only stored in the history, not actually '
-                    'flagged at this step.',
-                    default=None)
-args = parser.parse_args()
+# figure out which antennas have valid data and perform select-on-read
+ant_nums = np.unique(uvd.ant_1_array)
```
jsdillon: This actually might not work. It's possible for antenna to only show up in ant_2_array.

scripts/extract_hh.py
```diff
@@ -1,69 +1,47 @@
...
-parser.add_argument('files', metavar='files', type=str, nargs='+',
-                    help='Files to be processed.')
-parser.add_argument('--ex_ants_file', type=str, help='Text file with list of antennas'
-                    ' which are excluded downstream in RTP. Generally, these are '
-                    'antennas which are being actively commissioned, or known as bad.'
-                    ' Note these values are only stored in the history, not actually '
-                    'flagged at this step.',
-                    default=None)
-args = parser.parse_args()
+# figure out which antennas have valid data and perform select-on-read
```
jsdillon: Can you elaborate on exactly what you mean by valid data? What are you checking and why?

scripts/extract_hh.py
plaplant: I suppose I could add it, but the way that Jack is writing correlator files right now, all antennas with actual data in them have numbers less than 350.

scripts/extract_hh.py
plaplant: Good point, I'll make that change.

scripts/extract_hh.py
plaplant: The correlator is ostensibly writing down the full cross-correlation data for 192(ish) antennas, but the only ones that correspond to real data from the SNAPs (instead of just dummy placeholder antennas) have antenna numbers less than 350. I'll add more of these details to comments in the script.

PR #469, opened 2019-05-08
updates to lstbin 

nkern:
Minor cleaning-up updates to LSTBIN. Turns out I already added the ability to take calibration solutions, but I did some minor housekeeping improvements..

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -404,7 +404,7 @@ def lst_bin_arg_parser():
     a.add_argument("--lst_start", type=float, default=0, help="starting LST for binner as it sweeps across 2pi LST")
     a.add_argument("--fixed_lst_start", action='store_true', default=False, help="If True, make the start of the LST grid equal to lst_start, rather than the LST of the first data record.")
     a.add_argument("--ntimes_per_file", type=int, default=60, help="number of LST bins to write per output file")
-    a.add_argument("--file_ext", type=str, default="{type}.{time:7.5f}.uv", help="file extension for output files. See lstbin.lst_bin_files doc-string for format specs.")
+    a.add_argument("--file_ext", type=str, default="{type}.{time:7.5f}.uvh5", help="file extension for output files. See lstbin.lst_bin_files doc-string for format specs.")
     a.add_argument("--outdir", default=None, type=str, help="directory for writing output")
     a.add_argument("--overwrite", default=False, action='store_true', help="overwrite output files")
     a.add_argument("--sig_clip", default=False, action='store_true', help="perform robust sigma clipping before binning")
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -44,66 +44,45 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, lst_start=None, lst
 
     Parameters:
     -----------
-    data_list : type=list, list of DataContainer dictionaries holding complex visibility data
-
+    data_list : type=list, list of DataContainer dictionaries holding
+        complex visibility data for each night to average.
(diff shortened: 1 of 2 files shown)


hera_cal/lstbin.py
```diff
@@ -246,15 +225,11 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, lst_start=None, lst
...
     if return_no_avg:
-        # return all binned data instead of just bin average
-        data_bin = odict(list(map(lambda k: (k, np.array(list(odict(list(map(lambda k2: (k2, data[k][k2]),
-                                                                             sorted(data[k].keys())))).values()))),
-                                  sorted(data.keys()))))
-        flags_bin = odict(list(map(lambda k: (k, np.array(list(odict(list(map(lambda k2: (k2, flags[k][k2]),
-                                                                              sorted(flags[k].keys())))).values()))),
-                                   sorted(flags.keys()))))
+        # return all binned data instead of just the bin average
+        data_bins = odict([(k1, [data[k1][k2] for k2 in data[k1].keys()]) for k1 in data.keys()])
```
jsdillon: Before you cast these to numpy arrays (which I believe also makes a copy of the data). Is it OK that you're no longer copying and that you've got a list of arrays rather than a higher dimensional array?

hera_cal/lstbin.py
```diff
@@ -544,17 +519,18 @@ def lst_bin_files(data_files, filetype='uvh5', dlst=None, verbose=True, ntimes_p
...
     filetype : str, options=['uvh5', 'miriad']
     dlst : type=float, LST bin width. If None, will get this from the first file in data_files.
     lst_start : type=float, starting LST for binner as it sweeps from lst_start to lst_start + 2pi.
     fixed_lst_start : type=bool, if True, LST grid starts at lst_start, regardless of LST of first data
         record. Otherwise, LST grid starts at LST of first data record.
     ntimes_per_file : type=int, number of LST bins in a single output file
-    file_ext : type=str, extension to "zen." for output files. This must have three
-               formatting placeholders, first for polarization(s), second for type of file
-               Ex. ["LST", "STD", "NUM"] and third for starting LST bin of file.
+    file_ext : type=str, extension to "zen." for output files. This must have at least a ".{type}." field
```
jsdillon: Shouldn't the default be 4pol files? Why include {pol}?

nkern: to be backwards compatible.

hera_cal/lstbin.py
nkern: in this case yes b/c this line is really only used for debugging, or if the user wants to get higher order statistics from the data bins themselves

hera_cal/lstbin.py
jsdillon: I'd change the default kwarg, but mention in the docstring how to be backwards compatible.

hera_cal/lstbin.py
```diff
@@ -404,7 +404,7 @@ def lst_bin_arg_parser():
     a.add_argument("--lst_start", type=float, default=0, help="starting LST for binner as it sweeps across 2pi LST")
     a.add_argument("--fixed_lst_start", action='store_true', default=False, help="If True, make the start of the LST grid equal to lst_start, rather than the LST of the first data record.")
     a.add_argument("--ntimes_per_file", type=int, default=60, help="number of LST bins to write per output file")
-    a.add_argument("--file_ext", type=str, default="{}.{}.{:7.5f}.uv", help="file extension for output files. See lstbin.lst_bin_files doc-string for format specs.")
+    a.add_argument("--file_ext", type=str, default="{type}.{time:7.5f}.uv", help="file extension for output files. See lstbin.lst_bin_files doc-string for format specs.")
```
jsdillon: This should be uvh5 not uv

PR #470, opened 2019-05-09
Interp u updates

nkern:
minor updates to `ReflectionFitter.interp_u`

### DIFF
--- a/hera_cal/reflections.py
+++ b/hera_cal/reflections.py
@@ -64,7 +64,6 @@ The code here models the time and delay dependent behavior
 through a combination of SVD and fringe-rate filtering.
 """
 from __future__ import print_function, division, absolute_import
-
 import numpy as np
 import aipy
 import os
@@ -162,7 +161,7 @@ class ReflectionFitter(FRFilter):
             if key[:4] == "ref_":
                 setattr(self, key, {})
 
-    def model_auto_reflections(self, clean_resid, dly_range, clean_flags=None, clean_data=None, 
+    def model_auto_reflections(self, clean_resid, dly_range, clean_flags=None, clean_data=None,
                                keys=None, edgecut_low=0, edgecut_hi=0, window='none',
                                alpha=0.1, zeropad=0, fthin=10, Nphs=300, ref_sig_cut=2.0,
                                overwrite=False, reject_edges=True, verbose=True):
@@ -274,7 +273,7 @@ class ReflectionFitter(FRFilter):
         self.ref_gains = form_gains(self.ref_eps)
         for a in self.ants:
             for p in self.pols:
-                k = (a, uvutils.parse_jpolstr(p)) 
+                k = (a, uvutils.parse_jpolstr(p))
                 if k not in self.ref_gains:
                     self.ref_gains[k] = np.ones_like(clean_resid[keys[0]], dtype=np.complex)
 
@@ -289,7 +288,7 @@ class ReflectionFitter(FRFilter):
         Iteratively perturbs reflection parameters in ref_* dictionaries,
         and applies to input clean_data until reflection bump amplitude
         inside dly_range is minimimzed to within a tolerance.
- 
+
         Args:
             clean_data : 2D ndarray of shape (Ntimes, Nfreqs)
                 CLEANed auto-correlation visibility data
@@ -344,7 +343,7 @@ class ReflectionFitter(FRFilter):
         if clean_flags is None:
             clean_flags = DataContainer(dict([(k, np.zeros_like(clean_data[k], dtype=np.bool)) for k in clean_data]))
         if clean_model is None:
-            clean_model = DataContainer(dict([(k, np.zeros_like(clean_data[k])) for k in clean_data]))            
+            clean_model = DataContainer(dict([(k, np.zeros_like(clean_data[k])) for k in clean_data]))
 
         # get keys: only use auto correlations and auto pols to model reflections
         if keys is None:
@@ -420,7 +419,7 @@ class ReflectionFitter(FRFilter):
         Args:
             output_calfits : str, filepath to write output calfits file to
             input_calfits : str, filepath to input calfits file to multiply in with
-                reflection gains. 
+                reflection gains.
             overwrite : bool, if True, overwrite output file
             write_npz : bool, if True, write an NPZ file holding reflection
                 params with the same pathname as output_calfits
@@ -491,7 +490,7 @@ class ReflectionFitter(FRFilter):
             max_dly : float, maximum |delay| of window
             side : str, options=['pos', 'neg', 'both']
                 Specifies window as spanning only positive delays, only negative delays or both.
- 
+
         Returns:
             wgts : DataContainer, holding sv_decomp weights
         """
@@ -662,7 +661,7 @@ class ReflectionFitter(FRFilter):
             # form principal components
             pcomps = np.einsum("ij,jk->jik", umodes[k], vmodes[k])
 
-            # multiply by svals and sum 
+            # multiply by svals and sum
             pc_model = np.einsum("i,i...->...", svals[k][:Nkeep], pcomps[:Nkeep])
 
             # add to pcomp_model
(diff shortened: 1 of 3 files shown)


hera_cal/reflections.py
```diff
@@ -272,11 +271,15 @@ def model_auto_reflections(self, clean_resid, dly_range, clean_flags=None, clean
...
-                k = (a, uvutils.parse_jpolstr(p)) 
-                if k not in self.ref_gains:
-                    self.ref_gains[k] = np.ones_like(clean_resid[keys[0]], dtype=np.complex)
+
+        # if ref_gains not empty, fill in missing antenna and polarizations with unity gains
+        if len(self.ref_gains) > 0:
+            _keys = list(self.ref_gains.keys())
+            for a in self.ants:
+                for p in self.pols:
+                    k = (a, uvutils.parse_jpolstr(p))
```
jsdillon: `self.pols` is visibility polarizations, no? Shouldn't you use `hera_cal.utils.split_pol(p)[0]`?

hera_cal/reflections.py
```diff
@@ -579,17 +582,24 @@ def sv_decomp(self, dfft, wgts=None, flags=None, keys=None, Nkeep=None,
 
     def project_svd_modes(self, dfft, umodes=None, svals=None, vmodes=None):
         """
-        Given two of the 3 SVD output vectors, project the input dfft data onto
-        the last remaining SVD vector.
+        Given two of the 3 SVD output matrices, project them onto the input dfft data
+        to estimate the last remaining SVD matrix. Note U and V are unitary matrices.
+
+        If
+            D = U S V
```
jsdillon: typically, SVD is written as D = U S V^dagger

hera_cal/reflections.py
```diff
@@ -579,17 +582,24 @@ def sv_decomp(self, dfft, wgts=None, flags=None, keys=None, Nkeep=None,
...
+        then
+            U = D V_dagger S_inv
```
jsdillon: In SVD, S is usually not invertible unless D is square.

hera_cal/reflections.py
```diff
@@ -579,17 +582,24 @@ def sv_decomp(self, dfft, wgts=None, flags=None, keys=None, Nkeep=None,
...
+            S = U_dagger D V_dagger
+            V = S_inv U_dagger D
 
         Args:
             dfft : DataContainer, holds time-delay waterfall visibilities
-            umodes : DataContainer, holds SVD umodes
-            svals : DataContainer, holds SVD singular values
-            vmodes : DataContainer, holds SVD vmodes
+            umodes : DataContainer, holds SVD umodes, shape (Ntimes, Nmodes)
```
jsdillon: you might want to add to the args the assumption of unitarity.

hera_cal/reflections.py
```diff
@@ -579,17 +582,24 @@ def sv_decomp(self, dfft, wgts=None, flags=None, keys=None, Nkeep=None,
...
+            svals : DataContainer, holds SVD singular values, shape (Nmodes,)
```
jsdillon: Clarify the assumption that nFreqs > nModes, which makes this shape correct.

hera_cal/reflections.py
```diff
@@ -758,8 +768,8 @@ def subtract_model(self, data, keys=None, overwrite=False, ifft=True, ifftshift=
             self.pcomp_model_fft[k] = model_fft
             self.data_pcmodel_resid[k] = data[k] - model_fft
 
-    def interp_u(self, umodes, times, full_times=None, uflags=None, keys=None, overwrite=False,
-                 gp_frate=1.0, gp_frate_degrade=0.0, gp_nl=1e-12, kernels=None, optimizer=None, verbose=True):
+    def interp_u(self, umodes, times, full_times=None, uflags=None, keys=None, overwrite=False, Ninterp=None,
```
jsdillon: To be pedantic, I'd say this is really a "fit" rather than an interpolation. Interpolation goes exactly through the known points, but this won't in general.

hera_cal/reflections.py
```diff
@@ -847,10 +861,22 @@ def interp_u(self, umodes, times, full_times=None, uflags=None, keys=None, overw
...
+            # setup regression data
+            X = times - Xmean
+            Y = umodes[k][:, :Ninterp].copy()
+            F = uflags[k].copy()
+
+            # mirror if desired
+            if Nmirror > 0:
+                assert Nmirror < X.size, "Nmirror can't be equal or larger than Ntimes"
+                _, X = vis_clean.zeropad_array(Y, binvals=X, zeropad=Nmirror, axis=0)
+                Y = np.concatenate([Y[1:Nmirror + 1, :][::-1, :], Y, Y[-Nmirror - 1:-1, :][::-1, :]], axis=0)
```
jsdillon: I think you can do this much more cleanly with `np.pad` with the `'reflect'` option.

hera_cal/reflections.py
```diff
@@ -847,10 +861,22 @@ def interp_u(self, umodes, times, full_times=None, uflags=None, keys=None, overw
...
+                F = np.concatenate([F[1:Nmirror + 1, :][::-1, :], F, F[-Nmirror - 1:-1, :][::-1, :]], axis=0)
+
             # setup regression data: get unflagged data
-            select = ~np.max(uflags[k], axis=1)
-            X = times[select, None] - Xmean
-            Y = umodes[k][select, :].copy()
+            select = ~np.max(F, axis=1)
```
jsdillon: I really don't like using max's and mins for comparing booleans. Instead, I'd use np.any or np.all.

hera_cal/vis_clean.py
```diff
@@ -378,7 +378,7 @@ def vis_clean(self, keys=None, data=None, flags=None, wgts=None, ax='freq', hori
 
         # iterate over keys
         for k in keys:
-            if k in self.clean_model and overwrite is False:
+            if k in getattr(self, clean_model) and overwrite is False:
```
jsdillon: For clarity, I would rename these variables something like "clean_model_name"... right now it's confusing that they are strings rather than the model/flags/resid themselves.

hera_cal/reflections.py
nkern: yeah but for the purposes of the explanation in the doc string it's easier to express it this way, and in practice it makes no difference

hera_cal/reflections.py
nkern: Yeah but for this explanation this is ok. the actual output of `np.linalg` for `S` is a vector of the singular values, such that its inverse is just `1 / S`.

hera_cal/reflections.py
nkern: They don't have to be exactly unitary to use this. If they are approximately unitary then you can also use this to get an approximation of the other matrices.

hera_cal/reflections.py
nkern: okay. I'm fine w/ the nomenclature though

hera_cal/reflections.py
nkern: I don't think that is required actually

hera_cal/reflections.py
nkern: I think `np.pad` doesn't have an `axis` kwarg, so it can only pad all axes of an array, whereas I need to pad only a single axis here

hera_cal/reflections.py
nkern: ok

hera_cal/vis_clean.py
nkern: ok, I think I'll just assign them as the datacontainers, which is probably easier to read anyways

hera_cal/reflections.py
jsdillon: I actually think it's confusing. When we mean element-by-element inverse, usually it's written as (S_ij)^-1 (and then you have to do all the index math). At least clarify in the docstring that you mean an element-wise inverse.

hera_cal/reflections.py
jsdillon: You can, actually. The `pad_width` kwarg takes a list of tuples: ((before_1, after_1), … (before_N, after_N))

So just set the axis padding width on the axes you don't want to pad to 0.

nkern: ah interesting. except that I'm discovering that it doesn't do this well for boolean arrays, so I'll have to do what I have here anyways for the flag array, so might as well be consistent between data and flags..

nkern:
@jsdillon ok if you have no other comments then I'll merge this, thanks!

PR #475, opened 2019-05-11
smooth_cal updates: fix delay bug and implement thresholding

### LINKED ISSUE #474: Spectral structure and flagging issues in IDR2.2 smooth_cal
Two issues have come up in our first attempt at IDR2.2 that need attention.

The first is with spectral structure in the smoothed calibration solutions. Here's a couple of examples from different nights:

image

image

This problem I've already diagnosed as being an issue with this line: https://github.com/HERA-Team/hera_cal/blob/dcabe7f1d4717fc7fa5db92e65a89c9279f3fbf8/hera_cal/smooth_cal.py#L160
We should be taking the real part, not the abs. As a result, our rephasing doesn't work properly for antennas with negative delay. 

--------

The second is a consequence of XRFI's current thresholding scheme, which can result in integrations that are nearly entirely flagged. For example:

Screen Shot 2019-05-10 at 11 53 43 AM

I don't trust these highly flagged integrations in smooth_cal and in LST-binning, so I think we should impose a global cutoff for frequencies/times in smooth_cal. This is a band-aid that we can eventually remove in favor of something like what's proposed here: https://github.com/HERA-Team/hera_qm/issues/251

jsdillon:
This PR fixes two issues with the current pipeline. This PR closes #474.

The first, and most pressing, is the bug where delays are not properly accounted for when smoothing calibration solutions (see #474). It works... compare the following to the same antenna I showed screwed up in #474:
Screen Shot 2019-05-12 at 10 33 54 PM

----------

The second is an implementation of day long thresholding and broadcasting of flags. In this implementation, I flag iteratively with a frequency and time threshold until convergence. Then I check if there are any antennas that have half as many unflagged visibilities as the least flagged antenna and, if so, flag them entirely. This is pretty conservative, but doesn't ultimately knock out that much data. Here's an example of the new flags due to this thresholding on 2498098:
Screen Shot 2019-05-12 at 10 10 58 PM

And here's an example on 2498107 (compare this to the flags seen in #474):
Screen Shot 2019-05-12 at 10 17 06 PM

@plaplant I'd like it if you can take a look at the delay issues.
@adampbeardsley I'd like it if you could look at the flagging part.

Thanks!


hera_cal/smooth_cal.py
```diff
@@ -257,26 +301,40 @@ def rephase_to_refant(gains, refant, flags=None):
 
 class CalibrationSmoother():
 
-    def __init__(self, calfits_list, flag_file_list=[], flag_filetype='h5', pick_refant=False, antflag_thresh=0.0, verbose=False):
+    def __init__(self, calfits_list, flag_file_list=[], flag_filetype='h5', antflag_thresh=0.0,
+                 pick_refant=False, freq_threshold=1.0, time_threshold=1.0, ant_threshold=1.0, verbose=False):
```
adampbeardsley: I don't _really_ care either way, but it seems odd to have the top level wrapper have a set of `freq_threshold` and `time_threshold` defaults that are different from this, but the same as the function that actually uses these keywords.

adampbeardsley:
@jsdillon I'm confused by the plot you showed above for 2498107. Are these flags _in addition_ to what is shown in #474?

jsdillon:
@adampbeardsley Yes, that's right. Those are the new flags from the thresholding and broadcasting.

hera_cal/smooth_cal.py
jsdillon: Yeah, I thought about that a bunch. I'm a little wary of making the flagging automatic (changing the default of the constructor), in part because CalibrationSmoother is a really useful way to load a bunch of calibration solutions and make them into a big grid (it's way, way faster than loading a list with pyuvdata). But I figured if you're going to explicitly call the more general function, you'll want to use the default parameters. Also, if you're calling it from a script. 

So yeah, it's a bit odd, but I tried to make it clear in the docstrings.

adampbeardsley: I figured you had a thought process like that. I'm fine with this solution, but just to throw another idea out there: you could have the default thresholds set to the 0.35 and 0.5 like the function, but have another keyword that determines whether to actually call that function or not. That keyword would default to False here.

jsdillon: I thought about that too... I find extra keywords that control whether keywords get used to be confusing and try to avoid them. It's too easy to change the settings but not flip the switch and be surprised when nothing happens.

hera_cal/smooth_cal.py
```diff
@@ -325,17 +383,20 @@ def __init__(self, calfits_list, flag_file_list=[], flag_filetype='h5', pick_ref
...
 
         # pick a reference antenna that has the minimum number of flags (tie goes to lower antenna number) and then rephase
         if pick_refant:
             utils.echo('Now picking reference antenna(s)...', verbose=self.verbose)
             self.refant = pick_reference_antenna(self.gain_grids, self.flag_grids, self.freqs, per_pol=True)
             utils.echo('\n'.join(['Reference Antenna ' + str(self.refant[pol][0]) + ' selected for ' + pol + '.' 
                                   for pol in sorted(list(self.refant.keys()))]), verbose=self.verbose)
-            self.rephase_to_refant()
+
+        self.reset_filtering()  # initializes self.filtered_gain_grids and self.filtered_flag_grids
```
plaplant: Is there a use case for resetting the filtering, but then doing nothing if refant is not set? Put another way, can the work done inside of `reset_filtering` actually be done in `rephase_to_refant`? It would make it easier for the user to track what's going on, and potentially avoid duplicating the gains if rephasing isn't actually done.

hera_cal/tests/test_smooth_cal.py
```diff
@@ -75,28 +75,44 @@ def test_freq_filter(self):
...
-        np.testing.assert_array_almost_equal(ff, np.ones((10, 10), dtype=complex))
+        np.testing.assert_array_almost_equal(ff, np.ones((100, 100), dtype=complex))
 
         # test rephasing
-        gains = np.ones((10, 10), dtype=complex)
-        wgts = np.ones((10, 10), dtype=float)
-        gains *= np.exp(2.0j * np.pi * np.outer(150e-9 * np.ones(10), freqs))
+        gains = np.ones((100, 100), dtype=complex)
+        wgts = np.ones((100, 100), dtype=float)
+        gains *= np.exp(2.0j * np.pi * np.outer(-151e-9 * np.ones(100), freqs))
```
plaplant: Good to see that we're testing negative delays explicitly 👍

hera_cal/tests/test_smooth_cal.py
```diff
@@ -75,28 +75,44 @@ def test_freq_filter(self):
...
         ff, info = smooth_cal.time_freq_2D_filter(gains, wgts, freqs, times)
-        np.testing.assert_array_almost_equal(ff, gains)
+        np.testing.assert_array_almost_equal(ff, gains, 4)
```
plaplant: Is there a reason why you had to decrease the level of agreement here? A part in 1e4 frankly seems too high...

jsdillon: Yeah, turns out we were using a special number before that gave better agreement that we should have expected more generally. When I changed it from 150e-9 to 151e-9, it broke.

hera_cal/smooth_cal.py
jsdillon: The rephasing happens in place (and we want it to keep happening after the filtering so that the reference antennas stays the reference antenna). So it's useful that the rephasing function applies to self.filtered_gain_grids instead of self.gain_grids. I could explicitly initialize the copies in the __init__, but 

The doubling of gains is unavoidable if we want to maintain the resetting functionality. I could get rid of that functionality to save memory if we wanted...

hera_cal/smooth_cal.py
```diff
@@ -358,16 +419,17 @@ def check_consistency(self):
                 assert np.all(np.abs(self.flag_freqs[ff] - self.freqs) < 1e-4), \
                     '{} and {} have different frequencies.'.format(ff, self.cals[0])
 
+    def rephase_to_refant(self):
+        '''If the CalibrationSmoother object has a refant attribute, this function rephases the
+        filtered gains to it.'''
+        if hasattr(self, 'refant'):
```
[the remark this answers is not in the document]
jsdillon: Probably not, since it calls rephase_to_refant after every filtering step.

hera_cal/smooth_cal.py
```diff
@@ -257,26 +301,40 @@ def rephase_to_refant(gains, refant, flags=None):
...
             flag_file_list: list of string paths to files containing flags as a function of baseline, times
                 and frequency. Must have all baselines for all times. Flags on baselines are broadcast to both
                 antennas involved, unless either antenna is completely flagged for all times and frequencies.
             flag_filetype: filetype of flag_file_list to pass into io.load_flags. Either 'h5' (default) or legacy 'npz'.
-            pick_refant: if True, automatically picks one reference anteanna per polarization. The refants chosen have the
-                fewest total flags and causes the least noisy phases on other antennas when made the phase reference.
             antflag_thresh: float, fraction of flagged pixels across all visibilities (with a common antenna)
                 needed to flag that antenna gain at a particular time and frequency. antflag_thresh=0.0 is
                 aggressive flag broadcasting, antflag_thresh=1.0 is conservative flag_broadcasting.
-            verbose: print status updates
```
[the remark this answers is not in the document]
jsdillon: Oops. Good call.

hera_cal/smooth_cal.py
jsdillon: yeah, let's kill the second copy... If someone is doing interactive work and wants to keep it around, they can explicitly deepcopy it.

jsdillon: Ok, done.

hera_cal/smooth_cal.py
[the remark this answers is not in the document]
plaplant: I think it's good practice to warn the user if they try to call a function but it ends up doing nothing. One option might be to add the warning if `refant` isn't defined, and then protect the initial call to `rephase_to_refant` with a check to ensure that `refant` is defined before calling it. I'm not going to die on this hill, so we can keep it as is, just wanted to say that I think more feedback to the user is always better.

hera_cal/smooth_cal.py
plaplant: Great, thanks!

hera_cal/tests/test_smooth_cal.py
plaplant: Okay, if it's in the neighborhood of what we expect, then that's fine.

hera_cal/smooth_cal.py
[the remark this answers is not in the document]
jsdillon: I added a warning but turned it off for the rephase steps inside filtering. So if a user wants to explicitly rephase, they'll be warned if there's no refant.

jsdillon:
Thanks @plaplant and @adampbeardsley !

hera_cal/smooth_cal.py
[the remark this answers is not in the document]
plaplant: Sounds like a great compromise 😃

PR #476, opened 2019-05-16
Convert testing framework from nose to pytest

plaplant:
This PR converts the testing framework of hera_cal from nose (which is deprecated and may go away in the future) to pytest, which is more fully featured and has much broader support. The docs are [here](https://docs.pytest.org/en/latest/contents.html), but the tl;dr is:
- Tests use bare `assert` statements rather than `nt.assert_true` or `nt.assert_equal`-type statements.
- Raising of errors is testing by calling `pytest.raises`, with similar syntax to nose's `nt.assert_raises` or unittest's `self.assertRaises`.
- Specific warnings (such as floating point exceptions) can be easily filtered using function decorators (some examples of which are [here](https://github.com/HERA-Team/hera_cal/blob/940f882ca9e15ccc5d45fe93e3285db3444b8534/hera_cal/tests/test_abscal.py#L31) and [here](https://github.com/HERA-Team/hera_cal/blob/940f882ca9e15ccc5d45fe93e3285db3444b8534/hera_cal/tests/test_delay_filter.py#L22)).
- Tests can be invoked from anywhere inside of the repo, by calling `pytest` to run all tests or, e.g., `pytest test_delay_filter.py::Test_DelayFilter::test_run_filter`.

The PR updates the travis script accordingly, and also fixes a few small warnings that were being hidden behind the large number of (relatively) benign warnings triggered by the tests.

@jsdillon there isn't a huge time pressure on this, so feel free to wait until after IDR2.2 is done before reviewing/merging. Just wanted to have your eyes on it so we can make sure new tests use this framework going forward.

### DIFF
--- a/hera_cal/tests/test_redcal.py
+++ b/hera_cal/tests/test_redcal.py
@@ -108,7 +108,7 @@ class TestMethods(object):
             for bl in bls[1:]:
                 ai, aj, pol = bl
                 ans = data[bl] / (gains[(ai, 'Jxx')] * gains[(aj, 'Jxx')].conj())
-                assert np.allclose(ans0, ans, atol=1e-7)
+                np.testing.assert_almost_equal(ans0, ans, decimal=7)
 
         reds = om.get_reds(antpos, pols=['xx', 'yy', 'xy', 'yx'], pol_mode='4pol')
         gains, true_vis, data = om.sim_red_data(reds)
@@ -127,10 +127,10 @@ class TestMethods(object):
                 ans_xy = data[(ai, aj, 'xy',)] / (gains[(ai, 'Jxx')] * gains[(aj, 'Jyy')].conj())
                 ans_yx = data[(ai, aj, 'yx',)] / (gains[(ai, 'Jyy')] * gains[(aj, 'Jxx')].conj())
                 ans_yy = data[(ai, aj, 'yy',)] / (gains[(ai, 'Jyy')] * gains[(aj, 'Jyy')].conj())
-                assert np.allclose(ans0xx, ans_xx, atol=1e-7)
-                assert np.allclose(ans0xy, ans_xy, atol=1e-7)
-                assert np.allclose(ans0yx, ans_yx, atol=1e-7)
-                assert np.allclose(ans0yy, ans_yy, atol=1e-7)
+                np.testing.assert_almost_equal(ans0xx, ans_xx, decimal=7)
+                np.testing.assert_almost_equal(ans0xy, ans_xy, decimal=7)
+                np.testing.assert_almost_equal(ans0yx, ans_yx, decimal=7)
+                np.testing.assert_almost_equal(ans0yy, ans_yy, decimal=7)
 
         reds = om.get_reds(antpos, pols=['xx', 'yy', 'xy', 'yX'], pol_mode='4pol_minV')
         gains, true_vis, data = om.sim_red_data(reds)
@@ -143,17 +143,17 @@ class TestMethods(object):
             ans0xy = data[(ai, aj, 'xy',)] / (gains[(ai, 'Jxx')] * gains[(aj, 'Jyy')].conj())
             ans0yx = data[(ai, aj, 'yx',)] / (gains[(ai, 'Jyy')] * gains[(aj, 'Jxx')].conj())
             ans0yy = data[(ai, aj, 'yy',)] / (gains[(ai, 'Jyy')] * gains[(aj, 'Jyy')].conj())
-            assert np.allclose(ans0xy, ans0yx, atol=1e-7)
+            np.testing.assert_almost_equal(ans0xy, ans0yx, decimal=7)
             for bl in bls[1:]:
                 ai, aj, pol = bl
                 ans_xx = data[(ai, aj, 'xx',)] / (gains[(ai, 'Jxx')] * gains[(aj, 'Jxx')].conj())
                 ans_xy = data[(ai, aj, 'xy',)] / (gains[(ai, 'Jxx')] * gains[(aj, 'Jyy')].conj())
                 ans_yx = data[(ai, aj, 'yx',)] / (gains[(ai, 'Jyy')] * gains[(aj, 'Jxx')].conj())
                 ans_yy = data[(ai, aj, 'yy',)] / (gains[(ai, 'Jyy')] * gains[(aj, 'Jyy')].conj())
-                assert np.allclose(ans0xx, ans_xx, atol=1e-7)
-                assert np.allclose(ans0xy, ans_xy, atol=1e-7)
-                assert np.allclose(ans0yx, ans_yx, atol=1e-7)
-                assert np.allclose(ans0yy, ans_yy, atol=1e-7)
+                np.testing.assert_almost_equal(ans0xx, ans_xx, decimal=7)
+                np.testing.assert_almost_equal(ans0xy, ans_xy, decimal=7)
+                np.testing.assert_almost_equal(ans0yx, ans_yx, decimal=7)
+                np.testing.assert_almost_equal(ans0yy, ans_yy, decimal=7)
 
     def test_check_polLists_minV(self):
         polLists = [['xy']]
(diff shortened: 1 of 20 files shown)


hera_cal/tests/test_redcal.py
```diff
@@ -414,14 +414,14 @@ def test_solver(self):
         w = dict([(k, 1.) for k in d.keys()])
 
         def solver(data, wgts, **kwargs):
-            np.testing.assert_equal(data['g_0_Jxx * g_1_Jxx_ * u_0_xx'], d[0, 1, 'xx'])
-            np.testing.assert_equal(data['g_1_Jxx * g_2_Jxx_ * u_0_xx'], d[1, 2, 'xx'])
-            np.testing.assert_equal(data['g_0_Jxx * g_2_Jxx_ * u_1_xx'], d[0, 2, 'xx'])
+            assert np.allclose(data['g_0_Jxx * g_1_Jxx_ * u_0_xx'], d[0, 1, 'xx'])
```
jsdillon: Same problem.

hera_cal/tests/test_redcal.py
```diff
@@ -414,14 +414,14 @@ def test_solver(self):
...
+            assert np.allclose(data['g_1_Jxx * g_2_Jxx_ * u_0_xx'], d[1, 2, 'xx'])
+            assert np.allclose(data['g_0_Jxx * g_2_Jxx_ * u_1_xx'], d[0, 2, 'xx'])
             if len(wgts) == 0:
                 return
-            np.testing.assert_equal(wgts['g_0_Jxx * g_1_Jxx_ * u_0_xx'], w[0, 1, 'xx'])
-            np.testing.assert_equal(wgts['g_1_Jxx * g_2_Jxx_ * u_0_xx'], w[1, 2, 'xx'])
-            np.testing.assert_equal(wgts['g_0_Jxx * g_2_Jxx_ * u_1_xx'], w[0, 2, 'xx'])
+            assert np.allclose(wgts['g_0_Jxx * g_1_Jxx_ * u_0_xx'], w[0, 1, 'xx'])
```
jsdillon: And again.

hera_cal/tests/test_redcal.py
```diff
@@ -511,9 +511,9 @@ def test_logcal(self):
             sol = info.logcal(d)
         om.make_sol_finite(sol)
         for red in reds:
-            np.testing.assert_array_equal(sol[red[0]], 0.0)
+            assert np.allclose(sol[red[0]], 0.0)
```
jsdillon: And here.

hera_cal/tests/test_smooth_cal.py
```diff
@@ -55,24 +54,24 @@ def test_freq_filter(self):
...
-        np.testing.assert_array_almost_equal(ff, gains)
+        assert np.allclose(ff, gains)
 
         # test skip_wgt
         gains = np.random.randn(10, 10) + 1.0j * np.random.randn(10, 10)
         wgts = np.ones((10, 10), dtype=float)
         wgts[0, 0:8] = 0
         freqs = np.linspace(100., 200., 10, endpoint=False) * 1e6
         ff, info = smooth_cal.freq_filter(gains, wgts, freqs, skip_wgt=.5)
-        np.testing.assert_array_equal(ff[0, :], gains[0, :])
```
jsdillon: Here again.

hera_cal/tests/test_smooth_cal.py
```diff
@@ -121,111 +120,112 @@ def test_pick_reference_antenna(self):
...
-        self.assertEqual(self.cs.flag_grids[54, 'Jxx'].shape, (180, 1024))
-        np.testing.assert_array_equal(self.cs.flag_grids[54, 'Jxx'][60:120, :], True)
+        assert len(self.cs.freqs) == 1024
+        assert len(self.cs.time_grid) == 180
+        assert np.allclose(self.cs.dt, 10.737419128417969 / 24 / 60 / 60)
+        assert (54, 'Jxx') in self.cs.gain_grids
+        assert (54, 'Jxx') in self.cs.flag_grids
+        assert self.cs.gain_grids[54, 'Jxx'].shape == (180, 1024)
+        assert self.cs.flag_grids[54, 'Jxx'].shape == (180, 1024)
+        assert np.allclose(self.cs.flag_grids[54, 'Jxx'][60:120, :], True)
```
jsdillon: Here again.

hera_cal/tests/test_utils.py
```diff
@@ -323,73 +311,73 @@ def test_lst_rephase():
...
-    np.testing.assert_array_equal(nObs, 1)
-    np.testing.assert_array_equal(chisq_per_ant[0, 'Jxx'], 4.0)
-    np.testing.assert_array_equal(chisq_per_ant[1, 'Jxx'], 4.0)
-    np.testing.assert_array_equal(nObs_per_ant[0, 'Jxx'], 1)
-    np.testing.assert_array_equal(nObs_per_ant[1, 'Jxx'], 1)
+    assert chisq.shape == (5, 10)
+    assert nObs.shape == (5, 10)
+    assert chisq.dtype == float
+    assert nObs.dtype == int
+    assert np.allclose(chisq, 4.0)
```
jsdillon: Same comment

hera_cal/tests/test_utils.py
```diff
@@ -323,73 +311,73 @@ def test_lst_rephase():
...
 
     # test with reds
     data = datacontainer.DataContainer({(0, 1, 'xx'): np.ones((5, 10), dtype=complex),
                                         (1, 2, 'xx'): np.ones((5, 10), dtype=complex)})
     model = datacontainer.DataContainer({(0, 1, 'xx'): 2 * np.ones((5, 10), dtype=complex)})
     chisq, nObs, chisq_per_ant, nObs_per_ant = utils.chisq(data, model, reds=[[(0, 1, 'xx'), (1, 2, 'xx')]])
-    np.testing.assert_array_equal(chisq, 2.0)
-    np.testing.assert_array_equal(nObs, 2)
-    nt.assert_false((1, 2, 'xx') in model)
+    assert np.allclose(chisq, 2.0)
```
jsdillon: And here.

hera_cal/tests/test_utils.py
```diff
@@ -398,12 +386,12 @@ def test_chisq():
     chisq, nObs, chisq_per_ant, nObs_per_ant = utils.chisq(data, model, data_wgts)
     chisq, nObs, chisq_per_ant, nObs_per_ant = utils.chisq(data, model, data_wgts, chisq=chisq, nObs=nObs,
                                                            chisq_per_ant=chisq_per_ant, nObs_per_ant=nObs_per_ant)
-    np.testing.assert_array_equal(chisq, 2.0)
-    np.testing.assert_array_equal(nObs, 2)
-    np.testing.assert_array_equal(chisq_per_ant[0, 'Jxx'], 2.0)
-    np.testing.assert_array_equal(chisq_per_ant[1, 'Jxx'], 2.0)
-    np.testing.assert_array_equal(nObs_per_ant[0, 'Jxx'], 2)
-    np.testing.assert_array_equal(nObs_per_ant[1, 'Jxx'], 2)
+    assert np.allclose(chisq, 2.0)
```
jsdillon: And here.

hera_cal/tests/test_utils.py
```diff
@@ -412,57 +400,54 @@ def test_chisq():
...
-    nt.assert_true('Jxx' in nObs)
-    nt.assert_true(chisq['Jxx'].shape, (5, 10))
-    nt.assert_true(nObs['Jxx'].shape, (5, 10))
-    np.testing.assert_array_equal(chisq['Jxx'], 1.0)
-    np.testing.assert_array_equal(nObs['Jxx'], 1)
+    assert 'Jxx' in chisq
+    assert 'Jxx' in nObs
+    assert chisq['Jxx'].shape == (5, 10)
+    assert nObs['Jxx'].shape == (5, 10)
+    assert np.allclose(chisq['Jxx'], 1.0)
```
jsdillon: And here.

hera_cal/tests/test_vis_clean.py
```diff
@@ -113,50 +116,50 @@ def test_vis_clean(self):
...
-        nt.assert_true('skipped' in V.clean_info[(24, 25, 'xx')][0])
-        nt.assert_true('success' in V.clean_info[(24, 25, 'xx')][3])
+        assert 'skipped' in V.clean_info[(24, 25, 'xx')][0]
+        assert 'success' in V.clean_info[(24, 25, 'xx')][3]
 
         # basic 2d clean
         V.vis_clean(keys=[(24, 25, 'xx'), (24, 24, 'xx')], ax='both', max_frate=10., overwrite=True,
                     filt2d_mode='plus')
-        nt.assert_true('success' in V.clean_info[(24, 25, 'xx')])
+        'success' in V.clean_info[(24, 25, 'xx')]
```
jsdillon: Are you missing an assert here?

plaplant:
Thanks for the detailed review @jsdillon. Just to respond to the two high level points.

1. I used the `np.isclose` construction for boolean values because the "right" way to test for booleans (asserting that `var is True`) fails, because the values are the `np.bool` datatype and not the singleton `True`, and so that assert fails. The `np.isclose` function accepts all numpy datatypes, and in particular will return `False` if two boolean values are different. I agree that it's odd to see it phrased that way, but it Does The Right Thing. Though as you suggest, we could switch to using `np.all`/`np.any` to test for boolean equality.
2. I don't personally use the `np.testing` module, and didn't realize it had those benefits. It should be possible to switch them back, as pytest is generally pretty flexible and can mix and match with different frameworks. In my mind, one benefit of `allclose` is requiring the user to explicitly condone NaNs, which by default are not equal (but are considered equal in `numpy.testing`). I'd _really_ like to see us get to a point where we avoid producing them in the first place. But I can switch back if that's really a showstopper.

jsdillon:
On 1) I think we should switch them to np.all / np.any, if only just for readability. When I see `close`, I immediately think I'm looking at a floating point comparison.

On 2) I'm kinda torn. I'm fine with any changes that are not `array` comparisons. I'd prefer to leave it when they are array comparisons. I don't feel very strongly, however, so I'll leave it up to you to decide how much time you want to spend.

plaplant:
@jsdillon I've gone back and addressed the main concerns above. On 1), I replaced all the `isclose` comparisons for boolean types. On 2), I've reverted back to `np.testing` calls (where they weren't generating warnings). I'll also respond to smaller inline comments.

hera_cal/tests/test_vis_clean.py
plaplant: Good catch, thanks!

hera_cal/vis_clean.py
```diff
@@ -825,6 +825,7 @@ def zeropad_array(data, binvals=None, zeropad=0, axis=-1, undo=False):
             if undo:
                 s = [slice(None) for j in range(data.ndim)]
                 s[ax] = slice(zeropad[i], -zeropad[i])
+                s = tuple(s)
```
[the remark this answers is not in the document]
plaplant: If we don't do this, a warning is generated:
```
Using a non-tuple sequence for multidimensional indexing is
deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the
future this will be interpreted as an array index, `arr[np.array(seq)]`,
which will result either in an error or a different result.
```
I can't do a tuple comprehension, because tuples are immutable, and so the elements can't be changed after it's created (as in the line immediately prior).

PR #482, opened 2019-05-24
LSTBIN partial io and partial blgroup update

nkern:
- deprecates lstbinning for `miriad`
- enables partial IO for lstbinning for `UVH5` 
- enables reduction of instantaneous memory load via looping over baseline groups

the last update is a paradigm shift that will allow us to lstbin BDA data in the future. Doing so will require minor changes to `config_lst_bin` and `lst_bin_files` in the future, but the updated paradigm should allow for it relatively easily


hera_cal/io.py
```diff
@@ -736,6 +736,8 @@ def get_file_lst_range(filepaths, filetype='uvh5', add_int_buffer=False):
         file_starts : file starting point (bin center) in LST [radians]
         file_stops : file ending point (bin center) in LST [radians]
         int_times : integration duration in LST [radians]
+        file_lst_arrays : ndarrays of file lst_array [radians]
```
jsdillon: I think we should modify this function to *only* return these last two. Call it `get_file_times` or something similar. We'll need to modify the function that uses it in abscal, but that's fine.

hera_cal/abscal.py
```diff
@@ -1629,36 +1629,36 @@ def match_times(datafile, modelfiles, filetype='uvh5', atol=1e-5):
...
-
-    # unwrap LST
-    if data_time[1] < data_time[0]:
-        data_time[1] += 2 * np.pi
-    model_start = model_times[0][0]
-    model_times[model_times < model_start] += 2 * np.pi
-    if data_time[0] < model_start:
-        data_time += 2 * np.pi
+    # get lst arrays
+    data_dlst, data_dtime, data_lst, data_times = io.get_file_times(datafile, filetype=filetype)
```
jsdillon: Shouldn't this be `data_lsts` just for consistency?

hera_cal/io.py
```diff
@@ -716,39 +716,40 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
...
 
     Args:
         filepaths : type=list, list of filepaths
         filetype : str, options=['miriad', 'uvh5']
-        add_int_buffer : type=bool, if True, extend stop times by an
-            integration duration. This is done so that when LST matching
-            the last bin is appropriately accounted for.
 
     Returns:
         If input is a string, output are floats, otherwise outputs are lists.
```
jsdillon: This isn't quite true... the third and fourth outputs are numpy arrays or lists of numpy arrays.

hera_cal/io.py
```diff
@@ -716,39 +716,40 @@ def load_flags(flagfile, filetype='h5', return_meta=False):
...
-    Start and stop are bin center. Miriad standard is bin start,
-    so shift by int_time / 2 is done. UVH5 standard is bin center,
-    so lsts are left untouched.
+    Miriad standard is bin start, so a shift by int_time / 2 is performed.
+    UVH5 standard is bin center, so times are left untouched.
+
+    Note: this is not currently compatible with Baseline Dependent Averaged data.
 
     Args:
         filepaths : type=list, list of filepaths
```
jsdillon: You might want to note that this could be a string.

hera_cal/io.py
```diff
@@ -758,43 +759,43 @@ def get_file_lst_range(filepaths, filetype='uvh5', add_int_buffer=False):
...
+            # form time arrays
+            lst_array = (start_lst + np.arange(uv['ntimes']) * int_time_rad) % (2 * np.pi)
+            time_array = start_time + np.arange(uv['ntimes']) * int_time
+
         elif filetype == 'uvh5':
-            # get lsts directly from uhv5 file's header, much faster than using empty HERAData object
-            lst_array = np.array(h5py.File(f, mode='r')[u'Header'][u'lst_array'])
+            # get times directly from uvh5 file's header: faster than loading entire file via HERAData
+            with h5py.File(f, mode='r') as _f:
+                lst_array = np.ravel(_f[u'Header'][u'lst_array'])
```
jsdillon: Why do you need to ravel?

jsdillon: I don't think you need to ravel here... you're raveling again below.

hera_cal/io.py
jsdillon: Also, the outputs are numpy arrays, not lists.

hera_cal/io.py
```diff
@@ -758,43 +759,43 @@ def get_file_lst_range(filepaths, filetype='uvh5', add_int_buffer=False):
...
-    file_starts = np.array(file_starts)
-    file_stops = np.array(file_stops)
-    int_times = np.array(int_times)
+        dlsts.append(int_time_rad)
+        dtimes.append(int_time)
+        file_lst_arrays.append(lst_array)
+        file_time_arrays.append(time_array)
 
-    # make sure times don't wrap
-    file_starts[np.where(file_starts < 0)] += 2 * np.pi
```
jsdillon: If we're going to remove the unwrapping, perhaps we should note in the docstring that the LSTs might wrap.

hera_cal/io.py
```diff
@@ -758,43 +759,43 @@ def get_file_lst_range(filepaths, filetype='uvh5', add_int_buffer=False):
...
+    dlsts = np.asarray(dlsts)
+    dtimes = np.asarray(dtimes)
+    file_lst_arrays = np.asarray(file_lst_arrays)
+    file_time_arrays = np.asarray(file_time_arrays)
 
     if _array is False:
-        file_starts = file_starts[0]
-        file_stops = file_stops[0]
-        int_times = int_times[0]
+        dlsts = dlsts[0]
```
jsdillon: Why not just `return dlsts[0], dtimes[0], file_lst_arrays[0], file_time_arrays[0]`?

hera_cal/lstbin.py
```diff
@@ -417,10 +416,11 @@ def lst_bin_arg_parser():
     a.add_argument("--silence", default=False, action='store_true', help='stop feedback to stdout')
     a.add_argument("--output_file_select", default=None, nargs='*', help="list of output file integers ot run on. Default is all output files.")
     a.add_argument("--vis_units", default='Jy', type=str, help="visibility units of output files.")
+    a.add_argument("--ignore_flags", default=False, action='store_true', help="Ignore flags in input data files.")
```
jsdillon: Can you be a little more verbose here? What does "ignore flags" mean? All data is treated as unflagged and the flags on the lst-binned product are all False?

hera_cal/lstbin.py
```diff
@@ -433,52 +433,50 @@ def config_lst_bin_files(data_files, filetype='uvh5', dlst=None, atol=1e-10, lst
     data_files : type=list of lists: nested set of lists, with each nested list containing
                  paths to data files from a particular night. These files should be sorted
                  by ascending Julian Date. Frequency axis of each file must be identical.
```
jsdillon: You probably want to add here that they must be uvh5 files.

hera_cal/lstbin.py
```diff
@@ -433,52 +433,50 @@ def config_lst_bin_files(data_files, filetype='uvh5', dlst=None, atol=1e-10, lst
...
-
-    # get start and stop times for each list of files in data_files.
-    # add_int_buffer adds an integration to the end time of df[:-1] files,
-    # and the %(2pi) ensures everything is within a 2pi LST grid.
-    data_times = []
-    for df in data_files:
-        data_times.append(np.array(io.get_file_lst_range(df, add_int_buffer=True, filetype=filetype))[:2, :].T % (2 * np.pi))
-
-    # unwrap data_times less than lst_start, get starting and ending lst
+        dlst, dtime, lst_arr, time_arr = io.get_file_times(data_files[0][0], filetype='uvh5')
```
jsdillon: If you're only using dlst from this, it should probably be `dlst, _, _, _ =`

hera_cal/lstbin.py
```diff
@@ -433,52 +433,50 @@ def config_lst_bin_files(data_files, filetype='uvh5', dlst=None, atol=1e-10, lst
...
+    time_arrays = []
+    for dfs in data_files:
+        dlsts, dtimes, larrs, tarrs = io.get_file_times(dfs, filetype='uvh5')
+        lst_arrays.append(larrs)
+        time_arrays.append(tarrs)
+    lst_arrays = np.asarray(lst_arrays)
+    time_arrays = np.asarray(time_arrays)
+
+    # unwrap times relative to lst_start, and get global start and end lst
     start_lst = 100
```
jsdillon: I find it very confusing to have two variables, one called `start_lst` and the other `lst_start`. Can we disambiguate them somehow?

hera_cal/lstbin.py
```diff
@@ -433,52 +433,50 @@ def config_lst_bin_files(data_files, filetype='uvh5', dlst=None, atol=1e-10, lst
...
-        end_lst = np.max(np.append(end_lst, dt.ravel()))
+    for larrs in lst_arrays:
+        for la in larrs:
+            if la[0] < (lst_start - atol):
+                la += 2 * np.pi
+        start_lst = np.min(np.append(start_lst, larrs.ravel()))
+        end_lst = np.max(np.append(end_lst, larrs.ravel()))
 
     # ensure start_lst isn't beyond 2pi
     if start_lst >= (2 * np.pi):
```
jsdillon: Should this be in a while loop, or is it impossible for it to get beyond 4pi?

hera_cal/lstbin.py
```diff
@@ -500,15 +498,15 @@ def config_lst_bin_files(data_files, filetype='uvh5', dlst=None, atol=1e-10, lst
...
-    return data_times, lst_grid, dlst, file_lsts, start_lst
+    return lst_grid, dlst, file_lsts, start_lst, lst_arrays, time_arrays
 
 
-def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbose=True, ntimes_per_file=60,
+def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_per_file=60,
                   file_ext="{type}.{time:7.5f}.uvh5", outdir=None, overwrite=False, history='', lst_start=0, 
                   fixed_lst_start=False, atol=1e-6, sig_clip=True, sigma=5.0, min_N=5, rephase=False,
-                  output_file_select=None, **kwargs):
+                  output_file_select=None, Nblgroups=1, ignore_flags=False, **kwargs):
```
jsdillon: I'd recommend making the default `Nblgroups=None` load all baselines simultaneously. That would match the convention in apply_cal and delay_filter and it's the only way to make loading everything straightforwardly accessible.

hera_cal/lstbin.py
```diff
@@ -544,15 +541,19 @@ def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbo
         filepath to calfits, UVCal or HERACal objects with gain solutions to
         apply to data on-the-fly before binning via hera_cal.apply_cal.calibrate_in_place.
         If no apply cal is desired for a particular file, feed as None in input_cals.
+    Nblgroups : int, default=1, Number of baseline groups to form and run in serial out of the total list
+        of baselines in order to reduce instantaneous memory loads
+    ignore_flags : bool, if True, ignore the flags associated with input files
```
jsdillon: again, I'd say that "ignore" means more clearly.

hera_cal/lstbin.py
```diff
@@ -579,110 +577,137 @@ def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbo
     kwargs['outdir'] = outdir
     kwargs['overwrite'] = overwrite
 
-    # get metadata from the zeroth data file
-    d, fl, ap, a, f, t, l, p = io.load_vis(data_files[0][0], return_meta=True, pick_data_ants=False, filetype=filetype)
+    # get metadata from the zeroth data file in the last day
+    uv = UVData()
```
jsdillon: All of this is much more easily done with HERAData. Most of this copies what's already stored in the metadata of a HERAData object (freqs, antpos, ants, times, etc.)

hera_cal/lstbin.py
```diff
@@ -579,110 +577,137 @@ def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbo
...
-    assert np.all(np.abs(np.diff(t) - np.median(np.diff(t))) < 1e-6), 'All integrations must be of equal length (BDA not supported).'
-    del d, fl, ap, a, f, t, l, p
-    garbage_collector.collect()
+    integration_time = np.median(np.diff(times)) * 24 * 3600.
+    assert np.all(np.abs(np.diff(times) - np.median(np.diff(times))) < 1e-6), 'All integrations must be of equal length (BDA not supported).'
+
+    # get baselines from data and form baseline groups
+    bls = sorted(uv.get_antpairs())
+    Nbls = len(bls)
+    blgroups = [bls[i * (Nbls // Nblgroups + 1):(i + 1) * (Nbls // Nblgroups + 1)] for i in range(Nblgroups)]
```
jsdillon: Does this break if Nblgroups is greater than the number of baselines? If you end up using `nbl_to_load`, you can see how I handled this problem here: https://github.com/HERA-Team/hera_cal/blob/ed46b2c546831259c62bc187240562db2ec9f45c/hera_cal/delay_filter.py#L166

hera_cal/lstbin.py
```diff
@@ -579,110 +577,137 @@ def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbo
...
+                    # unwrap la relative to itself
+                    larr = lst_arrs[j][k]
+                    tarr = time_arrs[j][k]
+                    larr[larr < larr[0]] += 2 * np.pi
+
+                    # check if this file has overlap with output file
+                    if larr[-1] < fmin or larr[0] > fmax:
+                        continue
+
+                    # if no overlap, get relevant time indicies
```
jsdillon: I think you want to delete the phrase "if no overlap"... this only happens if there is overlap.

hera_cal/lstbin.py
```diff
@@ -579,110 +577,137 @@ def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbo
...
-                    nightly_data_list.append(data_status[j][k][0])  # this is data
-                    nightly_flgs_list.append(data_status[j][k][1])  # this is flgs
-                    nightly_lst_list.append(data_status[j][k][6])  # this is lsts
-
-            # skip if nothing accumulated in nightly files
-            if len(nightly_data_list) == 0:
+            # skip if data_list is empty
+            if len(data_list) == 0:
+                # erase data references
+                del data_list, flgs_list, lst_list
```
jsdillon: I believe this is unnecessary. These get overwritten with [] and therefore should get garbage collected.

hera_cal/lstbin.py
```diff
@@ -579,110 +577,137 @@ def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbo
...
+            if ignore_flags:
+                flgs_list = None
+            (bin_lst, bin_data, flag_data, std_data,
+             num_data) = lst_bin(data_list, lst_list, flags_list=flgs_list, dlst=dlst, lst_start=start_lst,
+                                 lst_low=fmin, lst_hi=fmax, truncate_empty=False, sig_clip=sig_clip,
+                                 sigma=sigma, min_N=min_N, rephase=rephase, freq_array=freq_array, antpos=antpos)
 
             # erase data references
-            del file_list, data_list, flgs_list, lst_list
+            del data_list, flgs_list, lst_list
```
jsdillon: Again, I don't think we need this anymore, do we?

hera_cal/lstbin.py
```diff
@@ -714,10 +732,11 @@ def lst_bin_files(data_files, filetype='uvh5', input_cals=None, dlst=None, verbo
             continue
 
         # write to file
-        io.write_vis(bin_file, bin_data, bin_lst, freq_array, antpos, flags=flag_data, verbose=verbose, nsamples=num_data, filetype=filetype, **kwargs)
-        io.write_vis(std_file, std_data, bin_lst, freq_array, antpos, flags=flag_data, verbose=verbose, nsamples=num_data, filetype=filetype, **kwargs)
+        io.write_vis(bin_file, bin_data, bin_lst, freq_array, antpos, flags=flag_data, verbose=verbose, nsamples=num_data, filetype='uvh5', **kwargs)
+        io.write_vis(std_file, std_data, bin_lst, freq_array, antpos, flags=flag_data, verbose=verbose, nsamples=num_data, filetype='uvh5', **kwargs)
 
         del bin_file, std_file, bin_data, std_data, num_data, bin_lst, flag_data
```
jsdillon: I don't think we need these dels now.

hera_cal/io.py
nkern: unwrapping is done above

hera_cal/lstbin.py
nkern: it means all input data are included in the binning. i'll add that to the description

hera_cal/lstbin.py
nkern: if using `fixed_lst_start` they are the same. one is just a dummy variable to keep track of the actual starting point of the files, the other is a preset for the starting output LST files.

hera_cal/lstbin.py
nkern: Actually I think its easier to conceptualize the other way around. We know how much memory an entire file will take up (if its 5 GB on disk it will expand to roughly 2x that in memory based on `pyuvdata` performance). Knowing that I can only handle, say 1 GB of memory per task, its easy to hit that benchmark by specifying `Nblgroups=5`. Otherwise, I'd have to know exactly how many baselines were in the file to begin with, and then divide that by 5

hera_cal/lstbin.py
nkern: I think these we could stand to keep. I'm not exactly sure when Python's garbage collector is invoked (automatically once a data structure hits zero references?) and b/c these are the biggest objects might as well make sure they are being removed before another batch of data is loaded in.

hera_cal/lstbin.py
jsdillon: Can we add a _ to the front of the dummy variable?

hera_cal/lstbin.py
jsdillon: I can see your point. It would be nice to be consistent, but maybe it's not worth changing things.

PR #492, opened 2019-06-18
Fix smooth_cal delay Issues by implementing a single per-night, per-antenna delay.

### LINKED ISSUE #493: Unstable delay structure in smooth_cal
Before performing smooth_cal on an antenna, we take out an overall delay (so that the gain is centered at delay=0) and then rephase. This is necessary because a high-delay antenna might not be appropriately smoothed if the range of its spectral structure is outside of +/- e.g. 200 ns.  

Currently, that delay is calculated per-integration, but then smoothed on the same timescale as the calibrations themselves. However, when the delay estimate (blue) is unstable (i.e. due to time-varying flagging), the smoothing (orange) doesn't really help.

image

This leads to smooth_cal solutions that look like this, which is obviously wrong.

image

As a side note, I believe this is a problem now but not in IDR2.1 because we did the time smoothing first, which filled in the flags with reasonable guesses for the gains before performing fft_dly, leading to much more stable delay estimates as a function of time.

jsdillon:
This closes #493 by finding a single per-night, per-antenna delay and using that to recenter the gain in delay space before smoothing.

In my initial test on the most problematic gain I found from the current smooth_cal results (the one I showed in #493 and in the analysis Slack), this seems to have fixed the problem:

image

where the axes are channel number vs. integration number.

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -272,13 +272,12 @@ def pick_reference_antenna(gains, flags, freqs, per_pol=True):
             if per_pol: dictionary mapping gain polarizations string to ant-pol tuples
             else: ant-pol tuple e.g. (0, 'Jxx')
     '''
-    # compute delay and phase for all gains to flatten them as well as possible. Average over times.
-    df = np.median(np.diff(freqs))
+    # compute delay for all gains to flatten them as well as possible. Average over times.
     rephasors = {}
     for ant in gains.keys():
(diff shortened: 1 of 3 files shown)


nkern:
@jsdillon did you say you found better results by enacting an edgecut? We should add that in if it isn't already there

jsdillon:
Instead of using edgecut, I only pass the subset of frequencies that are not flagged for all times. It's basically the same thing.

PR #496, opened 2019-06-20
Keep track of relative differences between smooth_cal and abscal

### LINKED ISSUE #495: Implement a notion of smooth_cal quality
Quantify how well the smooth_cal reproduces the abscal, and record that in the qual and total_qual arrays. This can be a metric for whether smooth_cal is working.

jsdillon:
The issues we had in #493 made clear the need to keep track of how good a fit our smooth_cal solution is to the original abscal gains. The most sensible way is a relative difference between the two, which we can easily store in the `quality_array` and `total_quality_array`s in calfits files. This PR implements that bookkeeping. Ideally, this is a useful metric going forward for the RTP to make sure smooth_cal isn't failing or doing anything too radical. 

This closes #495.


hera_cal/utils.py
```diff
@@ -845,6 +845,44 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
...
+        relative_diff: dictionary mapping keys like (0, 'Jxx') to waterfalls of relative
+            differences between old and new gains
+        avg_relative_diff: dictionary mapping antpols (e.g. 'Jxx') to waterfalls. Flagged
+            gains are excluded from the average; completely flagged times and channels are
+            replaced by 0s to match the convention of chi^2 above
+
+    """
+    if denom is None:
+        denom = old_gains
+    relative_diff = {ant: np.abs((old_gains[ant] - new_gains[ant]) / denom[ant]) for ant in new_gains}
```
nkern: there are times when denom is likely to be zero. best to do a true divide and save ourselves the hassle of having to flag infs or nans later..

jsdillon: In general, we shouldn't have 0s in the gains (flagged gains usually have 1s in the values). But in the case where we do, what does `np.true_divide` actually get us? My understanding is that it's exactly the same as `/` in python 3...

nkern: if `a = [1, 2, 3]` and `b = [1, 0, 1]` then `np.true_divide(a, b, where=~np.isclose(b, 0))` returns `array([1., 0., 3.])` whereas `a/b` returns `array([ 1., inf,  3.])` and raises a dividebyzero warning.

nkern: Sometimes the data can be zero, whereas the model is not zero but the data isn't flagged (typically at band edges), so the gains go to zero

Issue #487, opened 2019-06-05
Calibrate omnical visibility solutions further


adampbeardsley:

Chuneeta:
The absolute calibration gains are identical for all baselines in redundant group, however the smoothcal gains does not line up for all the redundant baseline. The plot below show the waterfall plot of the smooth cal gains only for all the 14 m East-West redundant antenna pairs.
![smooth_gains_only](https://user-images.githubusercontent.com/8384144/59954990-87df0780-943c-11e9-878b-b5430d31147f.png)

Slices across one timestamp is shown below.
![smooth_gains_only_tslice](https://user-images.githubusercontent.com/8384144/59955018-af35d480-943c-11e9-8146-a785d7fc1acd.png)

We notice a huge discrepancies for baseline (69, 70) -- the red line and (68, 69) --the green line. This discrepancy is due to missing data since a hug chunk of data across time is flagged for antenna 69 and in the future the antenna will be considered as bad and flagged as per @jsdillon. So this should pose great issue for now. However we need to find a way to either account for the solutions for these weird baselines during the smoothing process.
  
Also the small fluctuation that we see in the remaining baselines are mainly due to missing data. Averaging the solutions across all the redundant pairs might be possible solution.

PR #499, opened 2019-06-25
Average smooth gains

Chuneeta:
This PR addresses issue #487. It includes a module that allows to undo old calibration gains (omnical gains) and apply new gains (with abscal and smoothing) averaged over all redundant baselines in a redundant group. We chose to use the averaged gains because of the variations of the smooth cal gains.


Chuneeta:
@jsdillon looks fine to me. The doc string was missing one parameter which I have added.

jsdillon:
Here's some proof that it's working: here I'm comparing absolute value and phase of a single short baseline (top) and the redundant baseline group post-calibrated (bottom).

image

image

PR #500, opened 2019-07-03
Delay Filter CLEAN model trimming

nkern:
This adds the ability to trim the CLEAN model below an amplitude level, which is useful in the case of over-CLEANing down to the noise of a visibility

### DIFF
--- a/hera_cal/reflections.py
+++ b/hera_cal/reflections.py
@@ -885,6 +885,7 @@ class ReflectionFitter(FRFilter):
                 # shift by median and normalize by MAD
                 ymed = np.median(y, axis=0, keepdims=True)
                 ymad = np.median(np.abs(y - ymed), axis=0, keepdims=True) * 1.4826
+                # in rare case that ymad == 0, set it to 1.0 to prevent divbyzero error
                 ymad[np.isclose(ymad, 0.0)] = 1.0
                 y = (y - ymed) / ymad**2
 
--- a/hera_cal/reflections.py
+++ b/hera_cal/reflections.py
@@ -885,6 +885,7 @@ class ReflectionFitter(FRFilter):
                 # shift by median and normalize by MAD
                 ymed = np.median(y, axis=0, keepdims=True)
                 ymad = np.median(np.abs(y - ymed), axis=0, keepdims=True) * 1.4826
+                ymad[np.isclose(ymad, 0.0)] = 1.0
                 y = (y - ymed) / ymad**2
 
                 # fit gp and predict
--- a/hera_cal/tests/test_vis_clean.py
+++ b/hera_cal/tests/test_vis_clean.py
@@ -10,8 +10,10 @@ from copy import deepcopy
 import os
 import sys
 import shutil
-from scipy import constants
+from scipy import constants, interpolate
 from pyuvdata import UVCal, UVData
+from hera_sim import noise
+from uvtools import dspec
 
 from hera_cal import io, datacontainer
 from hera_cal import vis_clean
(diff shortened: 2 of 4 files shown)


jsdillon:
Looks like you're failing tests @nkern

nkern:
@jsdillon okay seems like failing tests are IERS table, so feel free to review whenever you get a chance, thanks!

nkern:
@jsdillon ready for review whenever you get a minute, thanks

hera_cal/reflections.py
```diff
@@ -885,6 +885,7 @@ def interp_u(self, umodes, times, full_times=None, uflags=None, keys=None, overw
                 # shift by median and normalize by MAD
                 ymed = np.median(y, axis=0, keepdims=True)
                 ymad = np.median(np.abs(y - ymed), axis=0, keepdims=True) * 1.4826
+                ymad[np.isclose(ymad, 0.0)] = 1.0
```
jsdillon: This seems a bit dangerous. Perhaps add a comment or documentation? Are you accounting for it in some way?

hera_cal/tests/test_vis_clean.py
```diff
@@ -173,6 +175,57 @@ def test_fft_data(self):
...
+
+        # trim model
+        mdl, n = vis_clean.trim_model(V.clean_model, V.clean_resid, V.dnu, noise_thresh=3.0, delay_cut=500,
+                                      kernel_size=21)
+        clean_data2 = deepcopy(V.clean_data)
+        clean_data2[k][V.flags[k]] = mdl[k][V.flags[k]]
+        V.fft_data(clean_data2, window='bh', overwrite=True, assign='dfft3')
+
+        # get averaged spectra
+        n1 = vis_clean.noise_eq_bandwidth(dspec.gen_window('bh', V.Nfreqs))
```
jsdillon: This function should probably have its own unit test.

hera_cal/vis_clean.py
```diff
@@ -786,6 +787,76 @@ def fft_data(data, delta_bin, wgts=None, axis=-1, window='none', alpha=0.2, edge
...
+    # estimate noise in Fourier space
+    model = DataContainer({})
+    noise = DataContainer({})    
+    for k in keys:
+        # get rfft
+        rfft, delays = fft_data(clean_resid[k], dnu, axis=1, window='none', edgecut_low=edgecut_low, edgecut_hi=edgecut_hi, ifft=False, ifftshift=False, fftshift=False)
+        delays *= 1e9
+
+        # get clean_resid NEB
+        neb = noise_eq_bandwidth(~np.isclose(clean_resid[k], 0.0))[:, None]
```
jsdillon: I don't understand this step. We're talking about a noisy quantity and we're just checking to see if it's not close to 0? But shouldn't it be averaged on zero? So sometimes it will be... does that matter? Should there be an np.abs() in here? Wouldn't it make more sense to look at the flags?

At the very least, explain what you're doing in the comments.

hera_cal/vis_clean.py
```diff
@@ -786,6 +787,76 @@ def fft_data(data, delta_bin, wgts=None, axis=-1, window='none', alpha=0.2, edge
     return data, fourier_axes
 
 
+def trim_model(clean_model, clean_resid, dnu, keys=None, noise_thresh=5.0, delay_cut=3000,
```
jsdillon: Why 5.0?

hera_cal/vis_clean.py
```diff
@@ -786,6 +787,76 @@ def fft_data(data, delta_bin, wgts=None, axis=-1, window='none', alpha=0.2, edge
...
+    for k in keys:
+        # get rfft
+        rfft, delays = fft_data(clean_resid[k], dnu, axis=1, window='none', edgecut_low=edgecut_low, edgecut_hi=edgecut_hi, ifft=False, ifftshift=False, fftshift=False)
+        delays *= 1e9
+
+        # get clean_resid NEB
+        neb = noise_eq_bandwidth(~np.isclose(clean_resid[k], 0.0))[:, None]
+
+        # get noise estimate
+        noise[k] = np.median(np.abs((rfft * neb)[:, np.abs(delays) > delay_cut]), axis=1)
```
jsdillon: We should document what we're assuming about the noise here.

hera_cal/vis_clean.py
```diff
@@ -786,6 +787,76 @@ def fft_data(data, delta_bin, wgts=None, axis=-1, window='none', alpha=0.2, edge
...
+
+        # median filter it
+        if kernel_size is not None:
+            n = noise[k]
+            nlen = len(n)
+            n = np.pad(n, nlen, 'reflect', reflect_type='odd')
+            noise[k] = signal.medfilt(n, kernel_size=kernel_size)[nlen:-nlen]
+
+        # get mfft
+        mfft, _ = fft_data(clean_model[k], dnu, axis=1, window='none', edgecut_low=edgecut_low, edgecut_hi=edgecut_hi, ifft=False, ifftshift=False, fftshift=False)
```
jsdillon: Why assume that the window is none? Shouldn't it be the same window you CLEANed with?

nkern: No, it should be `none` b/c the clean model is ensured to have power only at delay modes of the data (i.e. no leakage) b/c its been CLEANed

nkern:
@jsdillon yes its doing the latter of what you say: filtering above the residual, but I have assumed this to be the same as the noise in the context of deep or over-cleaninng, which is the scenario one would use this function for

hera_cal/reflections.py
nkern: its was done to fix a rare but catastrophic failure that the ymad == 0, in which case you get a division by zero. its actually quite safe to do, b/c in this case the data itself is just a constant

hera_cal/vis_clean.py
nkern: yeah sorry, this is just getting the NEB of a tophat window with flags. so 1 unflagged and 0 flagged. But I'll make this more clear

hera_cal/vis_clean.py
nkern: just b/c its safer than 1.0, no particular reason, we could do 2.0 as well.

jsdillon:
Sorry I missed you this morning @nkern. Free to chat at 3pm?

acliu:
@jsdillon I love how there's now going to be permanent record on Github that you guys may have chatted at 3pm.

nkern:
@acliu for posterity, I can confirm that this indeed happened

PR #502, opened 2019-07-09
fix basic installation order issues

steven-murray:
This PR tries to fix the installation-order issue, similar to the way it was implemented in `pyuvdata` recently.

Along with this, I've amended the README installation instructions, and added a `conda` `environment.yml` file for simple development setup. There's a bit of debate whether this file should have absolutely pinned versions or not, but I've left them as free as the `install_requires` section in the `setup.py`. 

I've tested on a fresh environment, and everything seems to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,8 @@ install:
   # Useful for debugging any issues with conda
   - conda info -a
   # create environment and install dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coverage pytest-cov coveralls
-  - conda activate test-environment
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pytest-cov coveralls
+  - source activate test-environment
   - conda env update -n test-environment -f environment.yml
   - pip install .
   - conda list
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,17 +26,11 @@ install:
   # Useful for debugging any issues with conda
   - conda info -a
   # create environment and install dependencies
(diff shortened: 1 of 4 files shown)


steven-murray:
Note that I don't expect the PR to pass tests until `linsolve` and `uvtools` merge their respective PRs.

.travis.yml
```diff
@@ -26,17 +26,11 @@ install:
...
-  - source activate test-environment
-  - conda install -c conda-forge healpy aipy pycodestyle attrs h5py healpy pyyaml mpi4py
-  - pip install git+https://github.com/HERA-Team/pyuvdata.git
-  - pip install git+https://github.com/HERA-Team/pyuvsim.git
-  - pip install git+https://github.com/HERA-Team/linsolve.git
-  - pip install git+https://github.com/HERA-Team/hera_qm.git
-  - pip install git+https://github.com/HERA-Team/uvtools.git
-  - pip install git+https://github.com/HERA-Team/hera_sim.git
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coverage pytest-cov coveralls
+  - conda activate test-environment
```
plaplant: It looks like this line is making the Travis builds die. I think it should be `source activate test-environment`.

steven-murray: Yup, I've fixed that, thanks. It's now dying because of `pyuvsim`, which is one that I had forgotten to make these changes for. It should now have a PR, so we'll wait on that.

steven-murray:
Awesome, this looks all good to go then!

PR #508, opened 2019-07-24
Adds TempCal

nkern:
Adds a module `TempCal` for temperature calibration as a function of JD. Also refactors the gaussian process interpolation function into `utils.py` for broader use throughout `hera_cal`.

### DIFF
--- a/hera_cal/tempcal.py
+++ b/hera_cal/tempcal.py
@@ -9,6 +9,7 @@ import os
 import copy
 from six.moves import map, range
 from pyuvdata import UVData, UVCal, utils as uvutils
+from functools import wraps
 
 from .vis_clean import VisClean
 from . import utils
@@ -16,11 +17,133 @@ from .datacontainer import DataContainer
 from . import version, io
 
 
+def gains_from_autos(data, times, flags=None, smooth_frate=1.0, nl=1e-10,
+                     Nmirror=0, keys=None, edgeflag=0, verbose=False):
+    """
+    Model temperature fluctuations in auto-correlations
+    by dividing auto-correlation by its time-smoothed counterpart.
+
+    This is only valid for timescales significantly shorter than
+    the beam crossing timescale. Resultant gains have a "divide" gain_convention.
+
+    Rcommended to factorize the flags beforehand using utils.factorize_flags.
+
+    Args:
+        data : 2D ndarray or DataContainer
+            Auto-correlation waterfall of shape (Ntimes, Nfreqs) or a
+            DataContainer of them, in which case keys are ant-pair-pol tuples.
+        times : ndarray
+            Holds time array in Julian Date of shape (Ntimes,)
+        flags : 2D ndarray of DataContainer
+            Data flags or DataContainer of same shape as data.
+        smooth_frate : float
+            Maximum fringe-rate cutoff [mHz] for "low-pass" smoothing
+        nl : float
+            Noise level in GP interpolation.
+            Recommended to keep this near 1e-10 for non-expert users.
+        Nmirror : int
+            Number of time bins to mirror about edges before interpolation.
+            This can minimimze impact of boundary effects.
+        keys : list
+            List of ant-pair-pol tuples to operate on if fed DataContainers
+        edgeflag : int or len-2 tuple
+            Number of channels to flag on low and high end of frequency band.
+            Low and high can be specified independently via a len-2 tuple.
+        verbose : bool
+            If True, report feedback to stdout.
+
+    Returns:
+        ndarray or dictionary
+            Gain ndarray (or dict holding ant-pol keys and antenna gain values)
+        ndarray or dictionary
+            Gain flags (or dict holding ant-pol keys and gain flag values)
+        ndarray or DataContainer
+            Smoothed data (or DataContainer of such with same shape as input)
+    """
+    # parse input
+    if not isinstance(data, np.ndarray):
+        # assume its a DataContainer
+        gains, gflags, smooth = {}, {}, DataContainer({})
+        if keys is None:
+            keys = data.keys()
+        # only use autocorrs
+        keys = [k for k in keys if k[0] == k[1]]
+        assert len(keys) > 1, "Can only operate on auto-correlations!"
+        if flags is None:
+            flags = DataContainer(dict([(k, None) for k in keys]))
+        for key in keys:
+            utils.echo("starting {}".format(key), verbose=verbose)
+            gkey = utils.split_bl(key)[0]
+            g, gf, s = gains_from_autos(data[key], times, flags=flags[key], smooth_frate=smooth_frate,
+                                        nl=nl, Nmirror=Nmirror, edgeflag=edgeflag, verbose=False)
+            gains[gkey], gflags[gkey], smooth[key] = g, gf, s
+        return gains, gflags, smooth
+
+    # from here onward, assume ndarrays
+    # edgeflag
+    if edgeflag is not None:
+        if flags is None:
+            flags = np.zeros_like(data, np.bool)
+        else:
+            flags = copy.deepcopy(flags)
+        if isinstance(edgeflag, (int, np.int, float, np.float)):
+            edgeflag = (edgeflag, edgeflag)
+        assert len(edgeflag) == 2
+        if edgeflag[0] > 0:
+            flags[:, :edgeflag[0]] = True
+        if edgeflag[1] > 0:
+            flags[:, -edgeflag[1]:] = True
+
+    # get length scale in JD
+    length_scale = 1.0 / (smooth_frate * 1e-3) / (24.0 * 3600.0)
+
+    # smooth
+    data_shape = data.shape 
+    smooth = utils.gp_interp1d(times, data, flags=flags, length_scale=length_scale, nl=nl,
+                               Nmirror=Nmirror)
+
+    # take ratio and compute gain term
+    gflags = np.isclose(smooth, 0.0)
+    gains = np.sqrt(np.true_divide(data, smooth, where=~gflags))
+
+    # only allow frequency-averaged gains for now via nanmedian
+    # use median over mean to help w/ unflagged RFI
+    gains[gflags] = np.nan
+    gains = np.nanmedian(gains, axis=1, keepdims=True)
+    gains[np.isnan(gains)] = 1.0  # catch for fully flagged integrations
+    gains = np.repeat(gains, data_shape[1], axis=1)
+    gflags = np.repeat(np.all(gflags, axis=1, keepdims=True), data_shape[1], axis=1)
+
+    return gains, gflags, smooth
+
+
+def gains_from_tempdata(tempdata, times, temp_coeff):
+    """
+    Convert measurements of temperature data to gains based on temperature
+    coefficient and the time chosen initially for absolute calibration.
+
+    Args:
+        tempdata : ndarray or DataContainer
+            Holds temperature data [Kelvin] for each auto-correlation
+            as a function of time.
+        times : 1d ndarray
+            Times of data in JD
+        temp_coeff : float
+            Temperature coefficient converting dTemp to dGain
+            in units [dGain / dTemp] 
+
+    Returns:
+        ndarray or dictionary
+            Per-antenna gains given temperature data.
+    """
+    raise NotImplementedError("gains_from_tempdata not yet implemented")
+
+
 class TempCal(VisClean):
     
     def __init__(self, *args, **kwargs):
         """
-        Initialize TempCal, see vis_clean.VisClean for kwargs.
+        Initialize TempCal with vis_clean.VisClean args and kwargs.
         """ 
         super(TempCal, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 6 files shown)


nkern:
Here are some performance figures for posterity. This is on IDR2.1 data [2458098]

auto_amp_cal
^ Autocorrelation amplitude across time

auto_gains
^ Derived antenna gains (colored) and antenna-average (black)

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # smooth data
+        if not hasattr(self, 'smooth'):
+            self.smooth = DataContainer({})
+        if not hasattr(self, 'ratio'):
+            self.ratio = DataContainer({})
+        for k in keys:
+            # smooth data
+            utils.echo("working on {}".format(k), verbose=verbose)
+            gkey = utils.split_bl(k)[0]
```
jsdillon: This makes an assumption that `utils.split_bl(k)[0] == utils.split_bl(k)[1]`, i.e. that it's an autocorrelation. You might want to enforce that assumption with an assert.

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+            self.gains[gkey] = np.sqrt(self.ratio[k])
+
+            # only allow frequency-averaged gains for now via nanmedian
+            # use median over mean to help w/ unflagged RFI
+            g = self.gains[gkey].copy()
+            g[self.gflags[gkey]] = np.nan
+            self.gains[gkey] = np.nanmedian(g, axis=1, keepdims=True)
+            self.gains[gkey][np.isnan(self.gains[gkey])] = 1.0  # catch for fully flagged integrations
+            self.gains[gkey] = np.repeat(self.gains[gkey], data_shape[1], axis=1)
+            self.gflags[gkey] = np.repeat(np.min(self.gflags[gkey], axis=1, keepdims=True), data_shape[1], axis=1)
```
jsdillon: I'd use np.all() rather than np.min(). It's more much readable since it's clear that you're working with bools.

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+from six.moves import map, range
+from pyuvdata import UVData, UVCal, utils as uvutils
+
+from .vis_clean import VisClean
+from . import utils
+from .datacontainer import DataContainer
+from . import version, io
+
+
+class TempCal(VisClean):
```
jsdillon: I don't see the point of making this a subclass of VisClean... you're not really using any of the build in functions in VisClean.

I think this code would be a lot more modular and reusable if so much stuff wasn't stored under the hood rather than transparently returned. I'd strongly recommend starting with functions at the top level that take DataContainers rather than having to init an object to use these functions. Given the number of inputs and outputs you have right now, it seems like you should just be returning things rather than opaquely storing them in the TempCal object. 

If you need to write standard ways of chaining things together, write a run function.

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+            # only allow frequency-averaged gains for now via nanmedian
+            # use median over mean to help w/ unflagged RFI
+            g = self.gains[gkey].copy()
+            g[self.gflags[gkey]] = np.nan
+            self.gains[gkey] = np.nanmedian(g, axis=1, keepdims=True)
+            self.gains[gkey][np.isnan(self.gains[gkey])] = 1.0  # catch for fully flagged integrations
+            self.gains[gkey] = np.repeat(self.gains[gkey], data_shape[1], axis=1)
+            self.gflags[gkey] = np.repeat(np.min(self.gflags[gkey], axis=1, keepdims=True), data_shape[1], axis=1)
+
+    def avg_ants(self, avg_ants):
```
jsdillon: I think supporting both lists and lists of lists is a mistake... it makes the function harder to read and maintain and more opaque to the user. If the user wants to average together a variety of subsets of antennas, let them call this in a for loop.

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def set_abscal_time(self, times, cal_time):
+        """
+        Normalize gains by Julian Date of abscal.
+
+        Args:
+            times : ndarray
+                Times of data in Julian Date.
+            cal_time : float
+                Julian Date of when absolute calibration was performed.
```
jsdillon: This is confusing. This really has nothing to do with abscal. This function is way more general than that. It's just pick a time to call |g| = 1. Could be anything! I'd rename the arguments and function to clarify that it's just `normalize_gains_at_given_time` or something like that.

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+            self.gains[gkey] /= abs(self.gains[gkey][tind])
+
+    def gains_from_tempdata(self, tempdata, times, cal_time, temp_coeff):
+        """
+        Convert measurements of temperature data to gains based on temperature
+        coefficient and the time chosen initially for absolute calibration.
+
+        Args:
+            tempdata : DataContainer
+                Holds temperature data [Kelvin] for each auto-correlation
```
jsdillon: This probably shouldn't be a datacontainer. It should just be a numpy array.

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+                Temperature coefficient converting dTemp to dGain
+                in units [dGain / dTemp] 
+
+        Result:
+            self.gains : dictionary
+                Holds per-antenna gains given temperature data.
+        """
+        raise NotImplementedError("gains_from_tempdata not yet implemented")
+
+    def write_gains(self, fname, add_to_history='', overwrite=False, verbose=True, **kwargs):
```
jsdillon: This can probably be eliminated.

hera_cal/tempcal.py
```diff
@@ -0,0 +1,203 @@
...
+                If True overwrite output if it exists
+            verbose : bool
+                Report feedback to stdout
+        
+        Returns:
+            UVCal instance
+                Output calibration object
+        """
+        utils.echo("...writing {}".format(fname), verbose=verbose)
+        uvc = io.write_cal(fname, self.gains, self.freqs, self.times, flags=self.gflags,
```
jsdillon: Unless you specifically invoke `return_uvc`, this function doesn't return anything.

hera_cal/tests/test_tempcal.py
```diff
@@ -0,0 +1,68 @@
...
+from .. import utils, io, apply_cal
+from ..tempcal import TempCal
+from ..data import DATA_PATH
+
+
+def test_tempcal():
+    # load data
+    dfiles = sorted(glob.glob(os.path.join(DATA_PATH, "zen.2458043.*.xx.HH.XRAA.uvh5")))
+    T = TempCal(dfiles)
+    autos = [bl for bl in T.bls if bl[0] == bl[1]]
```
jsdillon: This will also include cross-polarizations, which I don't think you want. Dunno if that file has any, but best to be explicit.

hera_cal/tests/test_tempcal.py
```diff
@@ -0,0 +1,68 @@
...
+import six
+import copy
+from pyuvdata import UVData
+
+from .. import utils, io, apply_cal
+from ..tempcal import TempCal
+from ..data import DATA_PATH
+
+
+def test_tempcal():
```
jsdillon: You should break your tests into multiple discrete tests, ideally one per function.

nkern: the load is not terribly fast, so I can probably chunk it into two or three separate functions, but don't want to do more than that.

nkern:
Not sure I follow: what are the potentially useless variables in self?

hera_cal/utils.py
```diff
@@ -845,6 +846,110 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
     return chisq, nObs, chisq_per_ant, nObs_per_ant
 
 
+def gp_interp1d(x, y, x_eval=None, flags=None, length_scale=1.0, nl=1e-10,
```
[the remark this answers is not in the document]
nkern: nope i can't: the GP module doesn't allow for heteroskedastic noise

hera_cal/utils.py
```diff
@@ -845,6 +846,110 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
...
+                kernel=None, Nmirror=0, optimizer=None):
+    """
+    Gaussian Process interpolation
+
+    Interpolate or smooth a series of datavectors y with a Gaussian Process.
```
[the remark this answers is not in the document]
nkern: yeah its supposed to mirror the `scipy.interpolate.interp1d` function. I can add a word on dependent and independent variable, but it seems pretty obvious given the doc string right above it

hera_cal/utils.py
```diff
@@ -845,6 +846,110 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
...
+        flags : ndarray
+            A boolean array of y flags of shape (Nvalues, Nvectors)
+        length_scale : float
+            Length scale for RBF kernel if input kernel is None.
+        nl : float
+            Noise level for WhiteNoise kernel if input kernel is None.
+        kernel : sklearn Kernel object
+            Custom kernel to use. This supercedes length_scale and nl choices.
+        Nmirror : int
+            Number of x values to mirror about either end before interpolation.
```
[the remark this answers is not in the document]
nkern: yup

hera_cal/utils.py
```diff
@@ -845,6 +846,110 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
...
+        optimizer : str
+            Hyperparameter optimization method. Default is no optimization.
+    """
+    # type checks
+    assert isinstance(x, np.ndarray)
+    assert isinstance(y, np.ndarray)
+    if y.ndim == 1:
+        y = y.reshape(-1, 1)
```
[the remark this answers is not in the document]
nkern: is it doing that? i don't think it is

hera_cal/utils.py
```diff
@@ -845,6 +846,110 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
...
+    if Nmirror > 0:
+        assert Nmirror < x.size, "Nmirror can't be equal or larger than x"
+        x = np.pad(x, Nmirror, mode='reflect', reflect_type='odd') 
+        y = np.concatenate([y[1:Nmirror + 1, :][::-1, :], y, y[-Nmirror - 1:-1, :][::-1, :]], axis=0)
+        flags = np.concatenate([flags[1:Nmirror + 1, :][::-1, :], flags, flags[-Nmirror - 1:-1, :][::-1, :]], axis=0)
+
+    # setup training values
+    X = x.reshape(-1, 1)
+
+    # get unique flagging patterns
```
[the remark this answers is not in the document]
nkern: but i don't just need a set of unique values: I need their indices in the array as well.

hera_cal/utils.py
```diff
@@ -845,6 +846,110 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
...
+    ypredict = []
+    for _y in [y.real, y.imag]:
+        # shift by median and normalize by MAD
+        ymed = np.median(_y, axis=0, keepdims=True)
+        ymad = np.median(np.abs(_y - ymed), axis=0, keepdims=True) * 1.4826
+        # in rare case that ymad == 0, set it to 1.0 to prevent divbyzero error
+        ymad[np.isclose(ymad, 0.0)] = 1.0
+        _y = (_y - ymed) / ymad
+
+        # iterate over flagging patterns
```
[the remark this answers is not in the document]
nkern: each frequency is done independently, but the `GP` module is vectorized so if the flagging pattern for each y-vector (column in y) is the same it goes much quicker, so this just does the minimum number of calls to `GP.fit` and `GP.predict` necessary given flagging patterns. In the limit that there is only one flagging pattern then all of `y` is interpolated at once; in the limit that each freq-bin has an individual flagging pattern than GP is called Nvectors times.

hera_cal/utils.py
```diff
@@ -845,6 +846,110 @@ def chisq(data, model, data_wgts=None, gains=None, gain_flags=None, split_by_ant
...
+        ypred = np.zeros((len(x_eval), _y.shape[1]), np.float)
+        for i, fh in enumerate(flag_hashes):
+            inds = flag_indices[fh]
+            select = ~flag_patterns[i]
+            if np.any(select):
+                GP.fit(X[select, :], _y[select, :][:, inds])
+                ypred[:, inds] = GP.predict(x_eval) * ymad[:, inds] + ymed[:, inds]
```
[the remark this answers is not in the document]
nkern: sure thing

hera_cal/tempcal.py
nkern: Well you can see that it inits from VisClean, so it inherits all of VisClean partial IO, metadata storing, on-the-fly input calibration, and data writing... which is kinda a lot.

But I agree that its useful to have top-level functions, although I think I'd prefer them to operate on numpy arrays. I'll add something and then make the `TempCal` object wrap around that. The `TempCal` object is a little thin currently b/c there are some features we haven't incorporated that might be important in the future (like calibration given actual temperature measurements)

hera_cal/tempcal.py
nkern: yeah a good point

hera_cal/tempcal.py
nkern: I mean, in practice it has everything to do with abscal: that's exactly why you'd use this function. It can be a general function, but it may be useful to keep its use specific to why we'd use it _in the context of temperature calibration_. If we had a function say in `utils` that did this, it might be named as you suggest.

hera_cal/tempcal.py
nkern: let's not fret over code that hasn't been written yet :) but i do think it should be a datacontainer eventually.

hera_cal/tempcal.py
nkern: yup sure thing

hera_cal/tempcal.py
jsdillon: I don't think it needs to move to utils until enough other code wants to use it, but giving it a very specific name when it's a more general function means that we're more likely to rewrite the code than reuse it if we ever have to do it again in another context. I'm totally fine with the docstring highlighting the specific use case.

jsdillon:
The extra variables in self are everything that comes with VisClean, though honestly I don't see the point of storing gains, gflags, and smooth in the object. Why not just write functions that return them?

As it stands, having just three things stored in the object is of course not a big deal. But I think this leads us down a bad path--namely, toward how abscal has dozens of attributes and properties and everything is stored internally to the object. This makes it opaque and hard to maintain. It also makes it harder to debug: when you're only using functions that have clear interfaces, you know exactly what's going on when you build up a new function from the existing ones.

Unless there's a good reason why the user should have to instantiate an object to do these operations, it's not helpful to future users who want to work with the code interactively to force them into using objects.

hera_cal/tempcal.py
nkern: but I think we should be specific if its in `TempCal`: if we want to use this functionality we shouldn't be calling it from `tempcal.TempCal`. I feel that if you want to rename it to something general, let's move it to utils, otherwise let's keep it as is

nkern:
but that's what the top-level functions are for. the class object is a convenience function for streamlined analysis. anyways, let's just break this into top-level functions for now b/c there isn't a whole lot we are doing with `tempcal` anyways right now.

Issue #322, opened 2018-06-18
add tests to redcal for non-unity weights


jsdillon:

jsdillon:
Also update documentation in redcal to reflect this change.

plaplant:
This was deemed "not critical" for IDR2.2/3.1.

PR #514, opened 2019-07-29
option to skip step in redundantly_calibrate

nkern:
Minor edits to `redcal.redundantly_calibrate` and `redcal_run` for option to skip one or more of the three steps in redcal (firstcal, logcal or omnical). Also allows for user to feed custom weights in `redundantly_calibrate` which can lead to significantly different results than the default behavior of computing from autos.

Default API is not changed.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -25,7 +25,6 @@ the i-th and j-th antenna respectively.
 from __future__ import print_function, division, absolute_import
 
 import os
-import sys
 from collections import OrderedDict as odict
 import copy
 import argparse
@@ -855,19 +854,22 @@ def global_phase_slope_logcal(model, data, antpos, solver='linfit', wgts=None,
 
 def merge_gains(gains, merge_shared=True):
     """
-    Merge a list of gain dictionaries.
+    Merge a list of gain (or flag) dictionaries.
+
+    If gains has boolean ndarray keys, interpret as flags
+    and merge with a logical OR.
 
     Parameters:
     -----------
     gains : type=list or tuple, series of gain dictionaries with (ant, pol) keys
-            and complex ndarrays as values.
+            and complex ndarrays as values (or boolean ndarrays if flags)
     merge_shared : type=bool, If True merge only shared keys, eliminating the others.
         Otherwise, merge all keys.
 
     Output:
     -------
-    merged_gains : type=dictionary, merged gain dictionary with same key-value
-                   structure as input gain dictionaries.
+    merged_gains : type=dictionary, merged gain (or flag) dictionary with same key-value
+                   structure as input dict.
     """
     # get shared keys
     if merge_shared:
@@ -878,9 +880,17 @@ def merge_gains(gains, merge_shared=True):
     # form merged_gains dict
     merged_gains = odict()
 
+    # determine if gains or flags from first entry in gains
+    fedflags = False
+    if gains[0][list(gains[0].keys())[0]].dtype == np.bool_:
+        fedflags = True
+
     # iterate over keys
     for i, k in enumerate(keys):
-        merged_gains[k] = reduce(operator.mul, [g.get(k, 1.0) for g in gains])
+        if fedflags:
+            merged_gains[k] = reduce(operator.add, [g.get(k, True) for g in gains])
+        else:
+            merged_gains[k] = reduce(operator.mul, [g.get(k, 1.0) for g in gains])
 
     return merged_gains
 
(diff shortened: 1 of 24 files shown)


scripts/redcal_run.py
```diff
@@ -17,5 +17,5 @@
            outdir=a.outdir, ant_metrics_file=a.ant_metrics_file, clobber=a.clobber, nInt_to_load=a.nInt_to_load, pol_mode=a.pol_mode,
            ex_ants=a.ex_ants, ant_z_thresh=a.ant_z_thresh, max_rerun=a.max_rerun, solar_horizon=a.solar_horizon, flag_nchan_low=a.flag_nchan_low,
            flag_nchan_high=a.flag_nchan_high, bl_error_tol=a.bl_error_tol, min_bl_cut=a.min_bl_cut, max_bl_cut=a.max_bl_cut, 
-           fc_conv_crit=a.fc_conv_crit, fc_maxiter=a.fc_maxiter, oc_conv_crit=a.oc_conv_crit, oc_maxiter=a.oc_maxiter, 
+           fc_conv_crit=a.fc_conv_crit, fc_maxiter=a.fc_maxiter, run_logcal=not a.no_logcal, oc_conv_crit=a.oc_conv_crit, oc_maxiter=a.oc_maxiter,
```
jsdillon: For clarity, I'd put this in parens: `run_logcal=(not a.no_logcal)`

hera_cal/redcal.py
```diff
@@ -985,18 +989,41 @@ def redundantly_calibrate(data, reds, freqs=None, times_by_bl=None, fc_conv_crit
...
+    data_dtype = data[list(data.keys())[0]].dtype
+    gain_keys = sorted(set([utils.split_bl(k)[0] for k in data] + [utils.split_bl(k)[1] for k in data]))
+    vis_keys = [r[0] for r in reds]
 
     # perform firstcal
-    rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit)
-    rv['gf_firstcal'] = {ant: np.zeros_like(g, dtype=bool) for ant, g in rv['g_firstcal'].items()}
+    if fc_maxiter > 0:
+        rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit, wgts=wgts)
+        rv['gf_firstcal'] = {ant: np.zeros_like(g, dtype=np.bool) for ant, g in rv['g_firstcal'].items()}
```
jsdillon: I think you can pull this line out of the conditional. It's equivalent to the version for fc_maxiter == 0.

hera_cal/redcal.py
```diff
@@ -985,17 +997,31 @@ def redundantly_calibrate(data, reds, freqs=None, times_by_bl=None, fc_conv_crit
...
-    rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit)
+    ants = sorted(set([ant for red in reds for bl in red for ant in utils.split_bl(bl)]))
+
+    # perform firstcal if required
+    if prior_firstcal is False:
+        # perform firstcal normally on raw data
+        rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit, wgts=wgts)
+    elif prior_firstcal is True:
+        # assume the data is pre-calibrated, so all firstcal gains are 1.0 + 0.0j
+        rv['g_firstcal'] = {ant: np.ones_like(list(data.values())[0]) for ant in ant}
```
nkern: `for ant in ants`

jsdillon: fixed

nkern:
I think this accomplishes effectively what I'd like to add to `redundantly_calibrate`, so it looks good to me overall! Not sure if you are done with editing the `test_` file, but we should update some of the tests which I think were from the old code. if the tests still pass that's fine, but maybe good to add a few more given what you've done (test assertionerror given maxiter=0) etc

jsdillon:
I have not edited the tests at all. They need to be updated, ideally in a way that closes #322. We're exposing weights in a way that hasn't been thoroughly vetted, so I'm slightly worried about that. It *should* work...

hera_cal/abscal.py
```diff
@@ -2845,6 +2846,9 @@ def post_redcal_abscal(model, data, flags, rc_flags, min_bl_cut=None, max_bl_cut
         phs_conv_crit: convergence criterion for updates to iterative phase calibration that compares
             the updates to all 1.0s.
         refant_num: integer antenna number defined to have 0 phase. If None, refant will be automatically chosen.
+        use_idealized_antpos : bool, Instead of actual antenna positions, use redcal to generate idealized
```
jsdillon: If this is false, then you're not correcting exactly the same degeneracies that redcal is removing. This was actually an issue that came up in validation step 2.0. I explain the issue in #451 and #453. I don't think we want this option in an explicitly post-redcal abscal. 

However, I can imagine wanting an option to turn the solved phase/delay slopes into physical units... we can chat when you get back about how best to do that.

jsdillon:
I'm closing this PR in favor of #823 which I think accomplishes the desired functionality

Issue #515, opened 2019-07-31
XX / YY mixup bug in abscal custom_ functions


nkern:
https://github.com/HERA-Team/hera_cal/blob/0953cd50d57ef095e6ed41d1adea49739e0b9e95/hera_cal/abscal.py#L2440

This is a bug in 2-pol calibration where xx solutions are used to fill yy solutions, because `self._gain_keys` is e.g. `[[(1, 'Jxx'), ...], [(1, 'Jyy'), ...]]`. 

I discovered this using `hera_cal.abscal.post_abscal_run` on IDR2 data, which uses `abscal.abscal_step` and feeds `abscal.custom_abs_amp_gain` function

PR #516, opened 2019-07-31
fixed XX / YY 2pol mixup in AbsCal.custom_*

nkern:
Resolves issue #515 where in abscal 2pol calibration, the custom_* funcs assigned XX solutions to requested YY keys. This is relevant for `abscal.post_redcal_abscal_run` which uses these functions. Tests have been added to catch this.

If IDR2.2 ran XX and YY pol abscal simultaneously, then it will need to be re-run :(. If XX and YY pol were ran independently (different tasks) then this bug won't affect the results.

[Update:] I just checked and it looks like the partial abs component of the gain for IDR2.2 is different for XX and YY, meaning that this bug did not affect the most recent run of IDR2.2 (dated 7/30/2019), which is probably due to the fact that the `post_redcal_abscal_run` function performs partial IO analysis across polarization groups. So seems like no need to re-run the pipeline.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -2438,7 +2438,7 @@ class AbsCal(object):
         if hasattr(self, '_dly_slope'):
             # form dict of delay slopes for each polarization in self._gain_keys
             # b/c they are identical for all antennas of the same polarization
-            dly_slope_dict = {ants[0][1]: self.dly_slope[k[0]] for ants in self._gain_keys}
+            dly_slope_dict = {ants[0][1]: self.dly_slope[ants[0]] for ants in self._gain_keys}
 
             # turn delay slope into per-antenna complex gains, while iterating over input gain_keys
             dly_slope_gain = odict()
@@ -2503,7 +2503,7 @@ class AbsCal(object):
         if hasattr(self, '_phs_slope'):
             # form dict of phs slopes for each polarization in self._gain_keys
             # b/c they are identical for all antennas of the same polarization
-            phs_slope_dict = {ants[0][1]: self.phs_slope[k[0]] for ants in self._gain_keys}
+            phs_slope_dict = {ants[0][1]: self.phs_slope[ants[0]] for ants in self._gain_keys}
 
             # turn phs slope into per-antenna complex gains, while iterating over input gain_keys
             phs_slope_gain = odict()
@@ -2565,7 +2565,7 @@ class AbsCal(object):
         if hasattr(self, '_abs_eta'):
             # form dict of abs eta for each polarization in self._gain_keys
             # b/c they are identical for all antennas of the same polarization
-            abs_eta_dict = {ants[0][1]: self.abs_eta[k[0]] for ants in self._gain_keys}
+            abs_eta_dict = {ants[0][1]: self.abs_eta[ants[0]] for ants in self._gain_keys}
 
             # turn abs eta into per-antenna complex gains, while iterating over input gain_keys
             abs_eta_gain = odict()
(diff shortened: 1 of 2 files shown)


hera_cal/abscal.py
```diff
@@ -2436,11 +2436,15 @@ def custom_dly_slope_gain(self, gain_keys, antpos):
...
         """
         if hasattr(self, '_dly_slope'):
-            # get dly slope dictionary
-            dly_slope = self.dly_slope[self._gain_keys[0][0]]
-            # turn delay slope into per-antenna complex gains, while iterating over gain_keys
-            return odict(list(map(lambda k: (k, np.exp(2j * np.pi * self.freqs.reshape(1, -1) * np.einsum("i...,i->...", dly_slope, antpos[k[0]][:2]))),
-                                  gain_keys)))
+            # form dict of delay slopes for each polarization in self._gain_keys
+            # b/c they are identical for all antennas of the same polarization
+            dly_slope_dict = {k[0][1]: self.dly_slope[k[0]] for k in self._gain_keys}
```
jsdillon: Given that this is a nested list, it might be more readable if you use something more evocative than `k` for the iterator. Perhaps `for ants in self._gain_keys`?

hera_cal/abscal.py
```diff
@@ -2436,11 +2436,15 @@ def custom_dly_slope_gain(self, gain_keys, antpos):
...
+
+            # turn delay slope into per-antenna complex gains, while iterating over input gain_keys
+            dly_slope_gain = odict()
+            for gk in gain_keys:
+                dly_slope_gain[gk] = np.exp(2j * np.pi * self.freqs.reshape(1, -1) * np.einsum("i...,i->...", dly_slope_dict[gk[1]], antpos[gk[0]][:2]))
```
jsdillon: Can this be done with np.dot? Einsums are always kinda hard to read and usually not worth it unless there's a substantial speedup involved.

hera_cal/abscal.py
```diff
@@ -2497,11 +2501,16 @@ def custom_phs_slope_gain(self, gain_keys, antpos):
         antpos : type=dictionary, contains antenna position vectors. keys are ant integer, values are ant position vectors
         """
         if hasattr(self, '_phs_slope'):
-            # get phs slope dictionary
-            phs_slope = self.phs_slope[self._gain_keys[0][0]]
-            # turn phs slope into per-antenna complex gains, while iterating over gain_keys
-            return odict(list(map(lambda k: (k, np.exp(1.0j * np.ones_like(self.freqs).reshape(1, -1) * np.einsum("i...,i->...", phs_slope, antpos[k[0]][:2]))),
-                                  gain_keys)))
+            # form dict of phs slopes for each polarization in self._gain_keys
```
jsdillon: same two comments as above

hera_cal/abscal.py
```diff
@@ -2554,8 +2563,16 @@ def custom_abs_eta_gain(self, gain_keys):
         gain_keys : type=list, list of unique (ant, pol). Ex. [(0, 'x'), (1, 'x'), (0, 'y'), (1, 'y')]
         """
         if hasattr(self, '_abs_eta'):
-            abs_eta = self.abs_eta[self._gain_keys[0][0]]
-            return odict(list(map(lambda k: (k, np.exp(abs_eta).astype(np.complex)), gain_keys)))
+            # form dict of abs eta for each polarization in self._gain_keys
```
jsdillon: same two comments as above

hera_cal/abscal.py
```diff
@@ -2600,8 +2617,15 @@ def custom_abs_psi_gain(self, gain_keys):
         gain_keys : type=list, list of unique (ant, pol). Ex. [(0, 'x'), (1, 'x'), (0, 'y'), (1, 'y')]
         """
         if hasattr(self, '_abs_psi'):
-            abs_psi = self.abs_psi[self._gain_keys[0][0]]
-            return odict(list(map(lambda k: (k, np.exp(1j * abs_psi)), gain_keys)))
+            # form dict of abs psi for each polarization in self._gain_keys
```
jsdillon: same two comments as above

hera_cal/abscal.py
nkern: I think einsum is the way to go here: there is a common theme across all of the functions using einsum, some of which are a little more complex (by one dimension) then this line you've picked out, so I'm gonna ask to defer that to a different issue if you feel you don't understand whats going on here

jsdillon: That's fine. Maybe add a few comments to help people read off which axes are being summed over.

PR #524, opened 2019-08-16
Dead antenna resurrection

### LINKED ISSUE #489: Allow Redcal to solve for flagged antennas without changing any other solutions
This can be an extension of this function:

https://github.com/HERA-Team/hera_cal/blob/8c05031856626b8b4d17fe1c5bb8759d896cd02f/hera_cal/redcal.py#L936

We also probably want to do something very similar for absolute calibration.

This is sort of related to #487

jsdillon:
This PR implements a way to solve for the gains of antennas despite flagging them in redundant calibration. After omnical, this solves for all unique visibilities and antenna gains in a way that leaves them flagged and does not affect any other antennas. This also lets us calculate chi^2 per antenna, which might help us assess whether to readmit a dubious antenna.

This follows-up on the work done by @Vasqalex this summer.

This closes #489.

### DIFF
--- a/hera_cal/tests/test_redcal.py
+++ b/hera_cal/tests/test_redcal.py
@@ -1112,6 +1112,11 @@ class TestRunMethods(object):
             # test redundant baseline counting
             np.testing.assert_array_equal(rv['vns_omnical'][(1, 12, pol)][~rv['vf_omnical'][(1, 12, pol)]], 4.0)
             np.testing.assert_array_equal(rv['vns_omnical'][(23, 27, pol)], 0.0)
+            np.testing.assert_array_equal(rv['vns_omnical'][(1, 27, pol)], 0.0)
+        for ant in [(1, 'Jxx'), (1, 'Jyy'), (27, 'Jxx'), (27, 'Jyy')]:
+            assert not np.all(rv['g_omnical'][ant] == 1.0)
+            assert not np.all(rv['chisq_per_ant'][ant] == 0.0)
+            np.testing.assert_array_equal(rv['gf_omnical'][ant], True)
 
     def test_redcal_run(self):
         input_data = os.path.join(DATA_PATH, 'zen.2458098.43124.downsample.uvh5')
--- a/hera_cal/tests/test_redcal.py
+++ b/hera_cal/tests/test_redcal.py
@@ -1098,6 +1098,7 @@ class TestRunMethods(object):
         for nsamples in rv['vns_omnical'].values():
             np.testing.assert_array_equal(nsamples, 0)
 
+        # this tests redcal.expand_omni_sol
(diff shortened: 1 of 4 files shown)


jsdillon:
@Chuneeta I'm having some trouble with the unit tests on travis, but they are passing locally so I think I'm OK getting your review now while I sort out the remaining details.

jsdillon:
@Chuneeta ok, finally figured out the travis issues. Ready for your review.

jsdillon:
For the record, this seems to be working as expected: 
Screen Shot 2019-08-17 at 11 04 16 AM

hera_cal/tests/test_redcal.py
```diff
@@ -1060,6 +1060,41 @@ def test_redundantly_calibrate(self):
...
+
+        # test that all chisqs are 0
+        for red in reds:
+            for bl in red:
+                ant0, ant1 = split_bl(bl)
+                np.testing.assert_array_almost_equal(d[bl], cal['g_omnical'][ant0] * np.conj(cal['g_omnical'][ant1]) * cal['v_omnical'][red[0]])
+        assert len(pols) * len(antpos) == len(cal['g_omnical'])
+        for ant in cal['chisq_per_ant']:
+            np.testing.assert_array_less(cal['chisq_per_ant'][ant], 1e-10)
+
```
Chuneeta: maybe add unit test to test if it works for unsolved unique baselines whose gains are known; this test will also test for the flags.

Chuneeta: I think it is a good to also add a unit test to check that the function breaks when more one antennas are excluded.

Chuneeta: Overall it looks good; just need the mentioned testing to ensure the outputs are as expected.

jsdillon: > maybe add unit test to test if it works for unsolved unique baselines whose gains are known; this test will also test for the flags.

I'm basically doing this by testing chi^2 per ant and testing `np.testing.assert_array_almost_equal(d[bl], cal['g_omnical'][ant0] * np.conj(cal['g_omnical'][ant1]) * cal['v_omnical'][red[0]])`. Because of the degeneracies, I don't know exactly what values for gains and visbilities I expect, but I expect them to satisfy that equation.

jsdillon: > I think it is a good to also add a unit test to check that the function breaks when more one antennas are excluded.

In this case I've excluded 4 (`ex_ants = [0, 13, 2, 18]`) of 37 antennas.

jsdillon:
@Chuneeta I think I've addressed your concerns.

jsdillon:
No problem. Thanks for your quick review @Chuneeta!

jsdillon:
@Vasqalex if you're curious, the final PR on your project has been merged in. Thanks for your work on this problem!

PR #528, opened 2019-08-21
do not require AIPY for most functions

### LINKED ISSUE #527: Bury aipy imports

steven-murray:
Closes #527.

Made sure to change the `setup.py` and `README` as well. 

We should consider adding a set of tests that test user-based installs (without necessarily running the test-suite). I think this could be achieved reasonably easily with `tox`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,14 +48,14 @@ setup_args = {
         'astropy',
         'pyuvdata',
         'pyephem',
-        'uvtools @ git+git://github.com/HERA-Team/uvtools',
         'linsolve @ git+git://github.com/HERA-Team/linsolve',
         'hera_qm @ git+git://github.com/HERA-Team/hera_qm',
         'scikit-learn'
     ],
     'extras_require': {
         "all": [
-            'aipy>=3.0rc2',
+            'aipy>=3.0',
+            'uvtools @ git+git://github.com/HERA-Team/uvtools',
         ]
     },
     'zip_safe': False,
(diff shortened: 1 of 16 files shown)


setup.py
```diff
@@ -47,13 +47,18 @@ def package_files(package_dir, subdirectory):
...
         'uvtools @ git+git://github.com/HERA-Team/uvtools',
         'linsolve @ git+git://github.com/HERA-Team/linsolve',
         'hera_qm @ git+git://github.com/HERA-Team/hera_qm',
         'scikit-learn'
     ],
+    'extras_require': {
+        "all": [
+            'aipy>=3.0rc2',
+        ]
+    }
```
plaplant: I think we need a trailing comma on this line to fix the installation issues.

steven-murray: yeah... :man_facepalming:

plaplant:
Also, if we wanted to set up the `tox` tests you mentioned as part of this, I'm all for that.

steven-murray:
I've tried to fix the uvtools imports as well, but I'm not sure doing `tox` in this PR is best. I think it might take a little thinking, and I know people are wanting this PR to be merged. I'll create an issue for it.

plaplant:
Hmm, looks like tests are failing because importing `hera_sim` is broken... did something change recently on hera_sim?

steven-murray:
Yup, and it's hopefully now fixed. I've restarted the build.

steven-murray:
@plaplant do you need to approve? I can't tell.

PR #531, opened 2019-08-22
Added polyfit_deg to vis_clean.trim_model

nkern:
Adds ability to fit a polynomial to derived delay-space noise curves across time in `vis_clean.trim_model` function.

### DIFF
--- a/hera_cal/vis_clean.py
+++ b/hera_cal/vis_clean.py
@@ -788,7 +788,7 @@ def fft_data(data, delta_bin, wgts=None, axis=-1, window='none', alpha=0.2, edge
 
 
 def trim_model(clean_model, clean_resid, dnu, keys=None, noise_thresh=2.0, delay_cut=3000,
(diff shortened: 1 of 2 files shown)


hera_cal/vis_clean.py
```diff
@@ -849,6 +852,13 @@ def trim_model(clean_model, clean_resid, dnu, keys=None, noise_thresh=2.0, delay
             n = np.pad(n, nlen, 'reflect', reflect_type='odd')
             noise[k] = signal.medfilt(n, kernel_size=kernel_size)[nlen:-nlen]
 
+        # fit a polynomial if desired
+        if polyfit_deg is not None:
+            x = np.arange(noise[k].size, dtype=np.float)
+            f = ~np.isnan(noise[k])
```
jsdillon: Might be a good idea to use `np.isfinite` to catch other bad stuff like `inf`s.

PR #543, opened 2019-10-29
Add astropy-healpix to dependencies

jsdillon:


mkolopanis:
From a little inspection it doesn't look like the main channel has the latest astropy updates for python 2.7, but it is pulling version 2.0.9 from main. This does not account for the recent IERS change. You may want to set conda-forge as a higher priority for the time being to get the latest version of astropy. 

I would also recommend taking a look at RadioAstronomySoftwareGroup/pyuvdata#708 to see how we are handling trying both the main and mirror downloads (it is very simple). Not 100% tested but it is looking good from my initial tests.

jsdillon:
I'm closing this PR in favor of #548

PR #548, opened 2019-11-07
Dump python 2, update documentation, and increase version number

### LINKED ISSUE #507: Update the README
The README is wildly out of date with respect to the list of modules in the repo as well as the scripts. (Firstcal and omnical still feature prominently...) These should be updated to reflect the current contents of the repo.

### LINKED ISSUE #511: Make new hera_cal release
To coincide with the release if IDR2.2, we should bump up the version number of hera_cal.

### LINKED ISSUE #547: Time to dump python 2?
The recent troubles with the IERS table and astropy have made it pretty clear that it's getting to be more trouble supporting python 2 than it's worth. IDR2.2 was run entirely with python 3, as will all future IDRs. Are we ready to dump support for python 2 and remove it from travis?

Any objections from @AaronParsons, @nkern, @plaplant, @adampbeardsley, or @bhazelton?

Also, we should handle this simultaneously with #507 and #511.

jsdillon:
This PR supercedes #543.

This closes #547. This closes #507. This closes #511.


jsdillon:
That's a bit more drastic than what I originally had in mind for this update... maybe make it an issue and resasses after Jan 1? That way if there's anyone left using python 2, we don't break things any earlier than the rest of the pack.

plaplant:
Okay, I'm fine with that compromise for now. I know not everyone is a fan of strict semantic versioning, but I think if we're not actually changing the compatibility of the package, then this should not be a major version change. I'd advocate making this `v2.1.0` (incrementing the minor version and adding a micro version, which is fairly standard versioning practice). When we tear out python2 compatibility, that would be the appropriate time to increment to `v3.0.0`.

jsdillon:
On second thought, let's just strip out python 2.

hera_cal/VERSION
```diff
@@ -1 +1 @@
-2.0
+3.0
```
plaplant: I think something like `3.0.0` is better, but I'll leave it up to you.

hera_cal/__init__.py
```diff
@@ -1,9 +1,7 @@
 # -*- coding: utf-8 -*-
-# Copyright 2018 the HERA Project
```
plaplant: My understanding  is that typically you only have to put the oldest year for the copyright and then it's implicitly carried forward, but now that they've been changed we can keep them as is.

hera_cal/VERSION
jsdillon: Given our history, I highly doubt we'll ever do that level of versioning.

hera_cal/__init__.py
jsdillon: I think you're right, but for some reason we did it last year...

PR #553, opened 2019-11-23
Add support for ee/nn and Jee/Jnn polarization strings

### LINKED ISSUE #462: Update polarization handling to ee/nn and Jnn/Jee
Once https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/580 is merged in.

### LINKED ISSUE #544: This can't handle 'nn' style polarizations
https://github.com/HERA-Team/hera_cal/blob/e85b503a296c1ef3f96e571c4270d5c99503006c/hera_cal/utils.py#L38

Also for function above. Need to pass x_orientation.

jsdillon:
This PR closes #462. This PR closes #544.

Basically, this PR does the following:
- All test visibility files now have `x_orientation`=`east`. This is consistent with H1C, even though the actual x_orientations were not recorded there.
- All unit tests use `ee` and `nn` strings when data is involved
- `hera_cal.io` now uses x_orientation to figure out polarization key strings
- `hera_cal.utils` polarization manipulation functions identify if the strings are in 'ee'/'nn' format by looking for an `e` or an `n`. If so, they keep the strings in that format.

Initially, I pursued a somewhat different strategy where DataContainers were initialized with x_orientations and then you could ask a DataContainer for a 'xx' baseline and get the same answer as if you asked it for an 'ee' baseline (assuming x_orientation = east). Also, in this version, if you try doing utils.comply_pol with xx and x_orientation=east, it'll give you back ee. I basically finished that version of this PR in 0367274cdf195c3ed2d26569953bdbe5959100c4 but then I decided that it was clunky and hard to maintain--passing around all those x_orientations was annoying and often it's useful to initialize a DataContainer on an empty dictionary, which was no longer really possible.

Also, I updated the environment because we need  https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/724


jsdillon:
@nkern I asked @adampbeardsley to review this PR, but since it touches a lot of your code I think it would be good for you to take a look at the changes to the part of the code you wrote. (If you want to review the whole thing, that's cool...I doubt adam will mind).

mkolopanis:
I would like to chime in and add that RTP on site currently Runs on code from this branch back at commit https://github.com/HERA-Team/hera_cal/pull/553/commits/1ba17344e6d4ac47623374e4fa4466de1e54302f. I think if people can find time, it would make me feel better if RTP did not run on random hot_fixes in a branch.

jsdillon:
@mkolopanis I've made some changes to the interfaces of some of the functions since that commit. Can you check to see whether the qm branch is compatible with the most recent commits here? If not, I wouldn't update it until this review is finished, but I would recommend flagging it as an issue over there.

mkolopanis:
Checked out this branch locally and all `hera_qm` tests pass for its x_orientation branch. Not a thorough test but at a cursory test seems like there is no incompatibility.

hera_cal/io.py
```diff
@@ -1292,7 +1294,7 @@ def load_cal(input_cal, return_meta=False):
 
 def write_cal(fname, gains, freqs, times, flags=None, quality=None, total_qual=None, write_file=True,
               return_uvc=True, outdir='./', overwrite=False, gain_convention='divide',
-              history=' ', x_orientation="east", telescope_name='HERA', cal_style='redundant',
+              history=' ', x_orientation="north", telescope_name='HERA', cal_style='redundant',
```
adampbeardsley: I don't remember where we landed on this, so I just want to clarify - should writing error if we don't actually know the `x_orientation`?

hera_cal/utils.py
```diff
@@ -23,19 +23,40 @@
     AIPY = False
 
 
+# Defines characters to look for to see if the polarization string is in east/north format. Nominally ['e', 'n'].
+_KEY_CARDINAL_CHARS = [c.lower() for c in _x_orientation_rep_dict('north').values()]
+
+
+def _is_cardinal(polstr):
+    '''Returns true if any characters in the polarization match those in _KEY_CARDINAL_CHARS.'''
+    return np.any([c.lower() in _KEY_CARDINAL_CHARS for c in polstr])
```
adampbeardsley: this worries me a bit. I can think of a million silly ways this can break, but a couple important ones are, "unknown" or "none". Is it not possible to just list all the possible (lower case) cardinal versions?

hera_cal/io.py
jsdillon: So, I've changed the default to match H3C. However, I took out a change I had originally made, which was to turn `None` into `"north"`. So if the `x_orientation` is `None`, usually because it was inherited from some data file, this will throw an error because x_orientation is a required UVCal parameter.

hera_cal/utils.py
jsdillon: That's a fair point. I'm trying to explicitly define everything in terms of the hard-coded definitions in pyuvdata, so that if those change again we might still be OK. Let me think about this for a bit and maybe I can come up with something more robust.

jsdillon: Does 41b5162 satisfy @adampbeardsley ?

hera_cal/io.py
adampbeardsley: so it sounds like there's an assumption that `x_orientation` will always be set when running `write_cal` -- either a known orientation, or `None`?

hera_cal/utils.py
adampbeardsley: Yeah, I think that works. Thanks!

hera_cal/io.py
jsdillon: Write_cal is meant to use all default parameters unless you set them, so it's easy to make a brand-new pyvudata compatible file. 

However, everytime write_cal is called in the pipeline, I bring in the x_orientation from the associated data file. If the x_orientation is not defined (i.e. `None`) and you use a routine like redcal, this will raise an error, which I think is the desired behavior.

adampbeardsley: Ok, I think that's reasonable enough.

PR #560, opened 2019-12-05
Update read to support pyuvdata's handling of lists of files

### LINKED ISSUE #549: upcoming uvdata read change
UVData is changing the read, and the specialized reads (like `read_uvh5`) to only use `read` when looking through multiple files at ones RadioAstronomySoftwareGroup/pyuvdata#705 
see the new lines in read_uvh5 here https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/e24279c7b1b1af296012e213fa2d25e1286d7bd3/pyuvdata/uvdata.py#L2968-L3079

jsdillon:
We now use super to call `UVData.read()` rather than calling `UVData.read_uvh5()`, etc. inside `io.HERAData` objects. Also added support for fast concatenation via the `axis` paramter.

This closes #549.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -42,7 +42,7 @@ class HERACal(UVCal):
         Arguments:
             input_cal: string calfits file path or list of paths
         '''
-        super(HERACal, self).__init__()
+        super().__init__()
 
         # parse input_data as filepath(s)
         if isinstance(input_cal, str):
@@ -211,7 +211,7 @@ class HERAData(UVData):
                 See UVData.read for more details.
         '''
         # initialize as empty UVData object
-        super(HERAData, self).__init__()
+        super().__init__()
 
         # parse input_data as filepath(s)
         if isinstance(input_data, str):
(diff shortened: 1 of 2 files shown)


jsdillon:
@mkolopanis Can you verify that this resolves the issue you were having?

mkolopanis:
That particular pyuvdata PR is having some trouble of it's own right now unfortunately and I can't tell if that's what is giving me issues locally.
If you want to try to build off it you can always make the necessary line in the envirnoment.yaml file look like 
`   - git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata@refs/pull/705/merge`

Otherwise may have to wait until after pyuvdata merges that one in.

jsdillon:
Well, if this works on the current master and no longer ever calls read_uvh5, read_miriad, etc. then this should work when that's merged in, right?

mkolopanis:
I'm not entirely convinced that will be true.

jsdillon:
Well, instead of letting this linger for months and maybe causing merge conflicts, my inclination is get this PR in and but not close #549 until pyuvdata is ready... especially since pyuvdata runs our tests, so a possible fix will make their lives easier. Even if it's only a lateral change for hera_cal, it doesn't hurt anything to get it in early.

jsdillon:
@mkolopanis hera_cal tests are passing locally with pyvudata on that branch

PR #561, opened 2019-12-13
Force antpos calculation to ignore antennas not in data

jsdillon:
This avoids an error if the antennas are at the center of the earth. This should probably be reversed at some point.


jsdillon:
Nevermind. We're going to change pyuvdata to issue a warning instead of a an error.

Issue #565, opened 2019-12-19
pyuvdata update seems to be breaking a test in test_reflections.py


jsdillon:
@bhazelton brought this to my attention: https://dev.azure.com/radioastronomysoftwaregroup/pyuvdata/_build/results?buildId=931&view=logs&j=1e2ce84c-b415-50fd-8deb-1c729abb5796&t=ab4dfa19-415c-5c99-e1f4-438d9e2c435c

Unfortunately, I can't reproduce it locally. I'll try running travis on master once https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/705 is merged in. 

@nkern what do you make of this?

bhazelton:
I see the same error on another branch of pyuvdata that doesn't contain the changes in the branch that the above error was seen on (and which has since been merged to master):
https://dev.azure.com/radioastronomysoftwaregroup/pyuvdata/_build/results?buildId=929&view=logs&jobId=1e2ce84c-b415-50fd-8deb-1c729abb5796&j=1e2ce84c-b415-50fd-8deb-1c729abb5796&t=ab4dfa19-415c-5c99-e1f4-438d9e2c435c

Note that I'm about to rebase this branch on master so future builds will not be independent of the prior PR.

jsdillon:
And now master is broken: https://travis-ci.org/HERA-Team/hera_cal/builds/626550585

@nkern any ideas???

jsdillon:
This was fixed in #569

PR #566, opened 2019-12-21
Remove unused check_histories

jsdillon:
These are deprecated in pyuvdata 1.5, but we don't use them anyway.


jsdillon:
@bhazelton hasn't fixed #565, but I'm ok merging it in since master is broken anyway.

PR #572, opened 2020-01-21
Add apply_cal x_orientation check

### LINKED ISSUE #570: apply_cal calibrate_in_place silent failure when x_orientation differs
When x_orientation differs from data and calibration, calibrate_in_place will silently flag all data. We should check that polarization convention (dipole vs. cardinal) agree and raise an error if not

jsdillon:
Added a check in `apply_cal.calibrate_in_place` and `apply_cal.calibrate_redundant_solution` to make sure if that all the gain keys are cardinal (e.g. north/east rather than x/y) but all the data keys are not (or vice versa), an error is raised.

This closes #570.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -25,5 +25,6 @@ dependencies:
     - git+https://github.com/HERA-Team/hera_qm
     - git+https://github.com/HERA-Team/linsolve
     - git+https://github.com/RadioAstronomySoftwareGroup/pyuvsim
+    - git+https://github.com/RadioAstronomySoftwareGroup/pyradiosky
     - git+https://github.com/HERA-Team/uvtools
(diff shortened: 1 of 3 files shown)


anguta:
> I don't see an explicit check that the calfits.x_orientation == uvdata.x_orientation. I think this should be done in addition to checking that both use cardinal keys since it's possible when both the calfits and data do not have cardinal keys, for them to be combined with inconsistent x-orientation.
> 
> In addition, checking for consistent x_orientation would allow for cal solutions without cardinal keys to be properly applied to data sets with cardinal keys and vice versa.

This can be pushed to a future pull request. I've noted an issue at https://github.com/HERA-Team/hera_cal/issues/574

PR #579, opened 2020-02-04
Reference Antenna Flag Propagation

jsdillon:
On a single integration of a single file in IDR3 (2458185.23736) I'm finding that the antenna chosen automatically as the reference in abscal (in part because it has the minimal number of flags) has flags where another antenna does not. This raises an error, but it seems to me like the safest thing to do is just also flag those antennas there. This implements that as an option for rephase_to_reference.

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -324,7 +324,7 @@ def pick_reference_antenna(gains, flags, freqs, per_pol=True):
     return refant
 
 
-def rephase_to_refant(gains, refant, flags=None):
+def rephase_to_refant(gains, refant, flags=None, propagate_refant_flags=False):
     '''Rephase all gains in place so that the phase of the reference antenna is 0 for all times and frequencies.
 
     Arguments:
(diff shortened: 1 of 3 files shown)


hera_cal/smooth_cal.py
```diff
@@ -324,24 +324,27 @@ def narrow_refant_candidates(candidates):
...
-def rephase_to_refant(gains, refant, flags=None):
+def rephase_to_refant(gains, refant, flags=None, propagate_refant_flags=False):
     '''Rephase all gains in place so that the phase of the reference antenna is 0 for all times and frequencies.
 
     Arguments:
         gains: Dictionary mapping antenna keys to gain waterfalls. Modified in place.
         refant: Antenna key of antenna to make the reference antenna (or dictionary mapping pols to keys)
         flags: Optional dictionary mapping antenna keys to flag waterfall.
             Used only to verify that all gains are flagged where the refant is flagged.
+        propagate_refant_flags: If True and flags is not None, update flags so that all antennas are flagged
```
anguta: expand this a tiny bit; "...all antennas are flagged at the time/frequencies where the reference antenna was also flagged."

anguta:
Looks good (minor comment on docstring). Assuming this passes its unit tests, I approve.

hera_cal/smooth_cal.py
jsdillon: Done.

PR #580, opened 2020-02-11
Find Polarity-Flipped Antennas and Report Them in a New Redcal Meta File

jsdillon:
This PR does two things which will help us with H3C RTP and redcal notebooks.

First, it tries to find polarity-flipped antennas using an algorithm spelled out here: https://github.com/HERA-Team/hera_cal/blob/6d8ec8e7b70d6433d7faa1bbdcc3c80a26d97133/hera_cal/redcal.py#L394-L410. If these are found successfully, then those angles (either 0 or pi) are used as the first iteration of firstcal's iterative phase solver. In my tests on H3C, this speeds up firstcal by roughly a factor of 2. 

Second, this piece of metadata, along with delays and everything previous `omni_meta` and some other useful metadata (history, lsts, times, antpos, freqs) are stored in a new .hdf5 file for which reading and writing capability is now in `hera_cal.io.`


jsdillon:
@Chuneeta Can you review the changes to redcal, test_redcal, and redcal_run, please?

@plaplant Do you mind reviewing the changes to io and test_io, since you wrote uvh5 and are the local hdf5 expert?

hera_cal/redcal.py
```diff
@@ -288,6 +288,85 @@ def _build_polarity_baseline_groups(dly_cal_data, reds, edge_cut=0, max_rel_angl
...
+
+    # Find all antennas in the polarity groups
+    ants = set([ant for grp_pair in polarity_groups.values() 
+                for grp in grp_pair for bl in grp for ant in utils.split_bl(bl)])
+
+    # Count how many times each antenna is involved in an assumed even group minus odd group
+    ant_even_counts = {ant: 0 for ant in ants}
+    for key, (grp1, grp2) in polarity_groups.items():
+        if key in even_vs_odd_IDs:
+            even, odd = {'even/odd': (grp1, grp2), 'odd/even': (grp2, grp1)}[even_vs_odd_IDs[key]]
```
Chuneeta: Any specific reason why the keys are "even/odd" and "odd/even" and not simply "even" and "odd"?

hera_cal/redcal.py
```diff
@@ -1454,11 +1455,11 @@ def redundantly_calibrate(data, reds, freqs=None, times_by_bl=None, fc_conv_crit
         times_by_bl = data.times_by_bl
 
     # perform firstcal
-    rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit)
+    rv['fc_meta'], rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit)
     rv['gf_firstcal'] = {ant: np.zeros_like(g, dtype=bool) for ant, g in rv['g_firstcal'].items()}
 
     # perform logcal and omnical
-    log_sol = rc.logcal(data, sol0=rv['g_firstcal'])
+    log_sol = _, rc.logcal(data, sol0=rv['g_firstcal'])
```
Chuneeta: I'm not sure I understand what is "_" for?

hera_cal/io.py
```diff
@@ -852,6 +852,89 @@ def partial_time_io(hd, times, **kwargs):
...
+    with h5py.File(meta_filename, "w") as outfile:
+        # save the metadata of the metadata
+        header = outfile.create_group('header')
+        header['freqs'] = freqs
+        header['times'] = times
+        header['lsts'] = lsts
+        antnums = np.array(sorted(list(antpos.keys())))
+        header['antpos'] = np.array([antpos[antnum] for antnum in antnums])
+        header['antpos'].attrs['antnums'] = antnums
+        header['history'] = history.encode('utf8')
```
plaplant: It's generally safer to cast `str` types to numpy strings using the `np.string_` method. So something like:
```
header['history'] = np.string_(history)
```
Without special handling, h5py behind the scenes will make an "object"-type dataset for `str` types (not sure what happens if you encode it first, as is done here), rather than a "string"-type one, which makes them more broadly compatible outside of python.

hera_cal/io.py
```diff
@@ -852,6 +852,89 @@ def partial_time_io(hd, times, **kwargs):
...
+        history: string describing the creation of this file
+    '''
+    with h5py.File(meta_filename, "r") as infile:
+        # decode metadata of metadata
+        freqs = infile['header']['freqs'][:]
+        times = infile['header']['times'][:]
+        lsts = infile['header']['lsts'][:]
+        antpos = {ant: pos for ant, pos in zip(infile['header']['antpos'].attrs['antnums'], 
+                                               infile['header']['antpos'][:, :])}
+        history = infile['header']['history'][()].decode('utf8')
```
plaplant: On the flip side of the `np.string_` comment above, you can unpack this using:
```
history = infile['header']['history'][()].tostring().decode('utf8')
```

hera_cal/io.py
```diff
@@ -852,6 +852,89 @@ def partial_time_io(hd, times, **kwargs):
...
+        header['times'] = times
+        header['lsts'] = lsts
+        antnums = np.array(sorted(list(antpos.keys())))
+        header['antpos'] = np.array([antpos[antnum] for antnum in antnums])
+        header['antpos'].attrs['antnums'] = antnums
+        header['history'] = history.encode('utf8')
+
+        # save firstcal metadata, saving dictionary keys as attrs
+        fc_grp = outfile.create_group('fc_meta')
+        ant_keys = sorted(list(fc_meta['dlys'].keys()))
```
plaplant: What are `ant_keys`? Are they tuples of strings (what it seems like) or something else? If they're strings, they'll need special handling like for the history key.

hera_cal/io.py
```diff
@@ -852,6 +852,89 @@ def partial_time_io(hd, times, **kwargs):
...
+        fc_grp['dlys'] = np.array([fc_meta['dlys'][ant] for ant in ant_keys])
+        fc_grp['dlys'].attrs['ants'] = ant_keys
+        fc_grp['polarity_flips'] = np.array([fc_meta['polarity_flips'][ant] for ant in ant_keys])
+        fc_grp['polarity_flips'].attrs['ants'] = ant_keys
+
+        # save the omnical metadata, saving dictionary keys as attrs
+        omni_grp = outfile.create_group('omni_meta')
+        pols_keys = sorted(list(omni_meta['chisq'].keys()))
```
plaplant: Same question as above re: `ant_keys`.

hera_cal/redcal.py
```diff
@@ -257,6 +257,37 @@ def reds_to_antpos(reds, tol=1e-10):
...
 
 
+def _build_polarity_baseline_groups(dly_cal_data, reds, edge_cut=0, max_rel_angle=np.pi/4):
+    '''This function looks at all redundant baselines and sees whether they mostly agree with the median
+    baseline or whether they look closer to being off by pi radians. The ones close to the median are the
+    "majority". The ones close to pi phase are the "minority." The rest are ambiguious and ignored in our
+    analysis. This function returns a dictionary that maps unique baselines to two groups of baselines,
+    both of which have the same internal polarity, through we don't know if it's "odd" (one flipped antenna)
+    or "even" (two or zero flipped antennas). See find_polarity_flipped_ants() for parameter descriptions.
+    '''
```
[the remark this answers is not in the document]
jsdillon: They aren't assigned even/odd in this function, just group 1 or group 2. Later functions will try to determine which is which.

hera_cal/redcal.py
```diff
@@ -257,6 +257,37 @@ def reds_to_antpos(reds, tol=1e-10):
...
+    # make sure edge_cut and max_rel_angle are sensible
+    assert 0 < max_rel_angle < np.pi / 2, "max_rel_angle must be between 0 and np.pi/2."
+    Nfreqs = list(dly_cal_data.values())[0].shape[1]
+    assert 2 * edge_cut < Nfreqs - 1, "edge_cut cannot be >= Nfreqs/2 - 1"
```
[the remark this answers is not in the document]
jsdillon: Why?

hera_cal/redcal.py
jsdillon: Yes. We're trying to figure out whether group 1 is odd and group 2 is even or group 2 is odd and group 1 is even.

hera_cal/redcal.py
jsdillon: It's just a convention for a variable that is returned but not used later.

hera_cal/redcal.py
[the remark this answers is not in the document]
jsdillon: fixed.

hera_cal/io.py

hera_cal/io.py

hera_cal/io.py

hera_cal/io.py

PR #588, opened 2020-04-14
added antnums2antnames dict in io.write_cal

nkern:
Allow `io.write_cal` to set antenna_names by user input, rather than by-default setting them to `'ant{ant_num}'`.


nkern:
@jsdillon if you'd like to do what you suggested (to have a `HERAData` passed as a kwarg) feel free to take a stab and push to this branch. otherwise I think this minimal change accomplishes what we want.

PR #591, opened 2020-04-16
Add the ability to filter reds for max number of tip/tilt degeneracies

### LINKED ISSUE #590: Enabling filtering of antennas to only calibrated "redundantly calibratable" antennas.
Some antenna configurations are not classically "redundantly calibratatable" in the sense that they introduce more than 2 tip/tilt degeneracies. Right now abscal can only solve for a single EW tilp/tilt and a single NS tip/tilt per polarization. The simplest solution is to have redcal flag the smallest number of antennas possible to reduce the dimensionality of the problem. A better solution would be give more tip/tilt DoF to abscal. I'll make that a separate issue...

EDIT: see #592 

This turns out to have been an issue for H3C in a way that depends sensitively on antenna flagging. For example, if we flag only the black antennas, the array is redcalable but if we also flag the red ones, it is no longer.

image

This turns out to have been the source of the phase discontinuities in abscal that I've been chasing down for a month or so.

Here's a phase after redcal and  abscal using RIMEz on 2458792 when we only flag the black antennas:

image

And here's what it looks like when we flag to the red ones too:

image

jsdillon:
This PR adds the ability to filter redundancies to keep the maximum number of tip/tilt degeneracies below some number. This is achieved by removing antennas, which leaves them flagged after calibration. Since a "classically redundantly calibratable" array has 2 tip/tilt degeneracies, this is the new default for the redcal run algorithms, though the parameter is exposed to the user.

This resolves #590

### DIFF
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -134,9 +134,9 @@ def filter_reds(reds, bls=None, ex_bls=None, ants=None, ex_ants=None, ubls=None,
                 pols=None, ex_pols=None, antpos=None, min_bl_cut=None, max_bl_cut=None, max_dims=None):
     '''
     Filter redundancies to include/exclude the specified bls, antennas, unique bl groups and polarizations.
-    Also allows filtering reds by removing antennas so that the number of tip/tilt degeneracies is no more
-    than max_dims. Arguments are evaluated, in order of increasing precedence: (pols, ex_pols, ubls, ex_ubls, 
-    bls, ex_bls, ants, ex_ants, min_bl_cut, max_bl_cut, max_dims).
+    Also allows filtering reds by removing antennas so that the number of generalized tip/tilt degeneracies
+    is no more than max_dims. Arguments are evaluated, in order of increasing precedence: (pols, ex_pols, 
+    ubls, ex_ubls, bls, ex_bls, ants, ex_ants, min_bl_cut, max_bl_cut, max_dims).
 
     Args:
         reds: list of lists of redundant (i,j,pol) baseline tuples, e.g. the output of get_reds().
@@ -161,8 +161,9 @@ def filter_reds(reds, bls=None, ex_bls=None, ants=None, ex_ants=None, ubls=None,
         max_bl_cut: cut redundant groups with average baselines lengths longer than this. Same units as antpos
             which must be specified.
         max_dims: maximum number of dimensions required to specify antenna positions (up to some arbitary shear).
-            This is equivalent to the number of tip/tilt phase degeneracies of redcal. 2 is a classically 
-            "redundantly calibratable" planar array. None means no filtering.
+            This is equivalent to the number of generalized tip/tilt phase degeneracies of redcal that are fixed
+            with remove_degen() and must be later abscaled. 2 is a classically "redundantly calibratable" planar
+            array. More than 2 usually arises with subarrays of redundant baselines. None means no filtering. 
 
     Return:
         reds: list of lists of redundant baselines in the same form as input reds.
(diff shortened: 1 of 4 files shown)


hera_cal/redcal.py
```diff
@@ -157,6 +160,9 @@ def filter_reds(reds, bls=None, ex_bls=None, ants=None, ex_ants=None, ubls=None,
             which must be specified.
         max_bl_cut: cut redundant groups with average baselines lengths longer than this. Same units as antpos
             which must be specified.
+        max_dims: maximum number of dimensions required to specify antenna positions (up to some arbitary shear).
```
anguta, line 163: Explain what dimensions mean in a comment. Are these unspecified degrees of freedom that are left undetermined after redcal that need to be fixed by an abscal model (or may not even be abscalable?)

hera_cal/redcal.py
```diff
@@ -130,14 +131,16 @@ def get_reds(antpos, pols=['nn'], pol_mode='1pol', bl_error_tol=1.0):
 
 
 def filter_reds(reds, bls=None, ex_bls=None, ants=None, ex_ants=None, ubls=None, ex_ubls=None, 
-                pols=None, ex_pols=None, antpos=None, min_bl_cut=None, max_bl_cut=None):
+                pols=None, ex_pols=None, antpos=None, min_bl_cut=None, max_bl_cut=None, max_dims=None):
     '''
     Filter redundancies to include/exclude the specified bls, antennas, unique bl groups and polarizations.
-    Arguments are evaluated, in order of increasing precedence: (pols, ex_pols, ubls, ex_ubls, bls, ex_bls,
-    ants, ex_ants).
+    Also allows filtering reds by removing antennas so that the number of tip/tilt degeneracies is no more
```
anguta, line 137: I think the fact that arbitrary additional numbers of redcal degeneracies are tip/tilt is somewhat confusing. Can we interpret them as a physical tip/tilt or, tip-tilts of different sub-arrays, are they even more general? If they aren't simply tip/tilt, I think you should phrase it as "additional degeneracies beyond tip/tilt" and maybe add more explanation if they are, for example, tip/tilts of sub-arrays or something similar. If they are tip/tilt explain what it means to describe tip/tilt with more than two degenerate parameters.

hera_cal/redcal.py
```diff
@@ -218,6 +224,28 @@ def split_bls(bls):
         lengths = [np.mean([np.linalg.norm(antpos[bl[1]] - antpos[bl[0]]) for bl in gp]) for gp in reds]
         reds = [gp for gp, l in zip(reds, lengths) if ((min_bl_cut is None or l > min_bl_cut)
                                                        and (max_bl_cut is None or l < max_bl_cut))]
+    
+    if max_dims is not None:
+        while True:
+            # Compute idealized antenna positions from redundancies
```
anguta, line 230: I think it would be helpful to put a reference here to where idealized antenna positions are described and maybe mention here that they are antenna positions with some arbitrary linear transformation (that's what I think they are based on reading other comments/code but if not say something else ;) ).

hera_cal/redcal.py
```diff
@@ -218,6 +224,28 @@ def split_bls(bls):
         lengths = [np.mean([np.linalg.norm(antpos[bl[1]] - antpos[bl[0]]) for bl in gp]) for gp in reds]
         reds = [gp for gp, l in zip(reds, lengths) if ((min_bl_cut is None or l > min_bl_cut)
                                                        and (max_bl_cut is None or l < max_bl_cut))]
+
```
anguta, line 227: I think the block of code below is hard too follow. Would it be possible to add a comment paragraph paragraph or two describing how it checks degrees of freedom and makes sure that the degeneracies do not end up exceeding  max_dims and then in each line add a short comment tying to the english narrative that is layed out int he paragraph?

jsdillon:
Thanks @anguta! I've updated the documentation a fair bit. Let me know if it's still confusing and we can iterate further.

hera_cal/redcal.py
anguta: Thanks for clarifying here about subarrays.

hera_cal/redcal.py
```diff
@@ -218,6 +225,36 @@ def split_bls(bls):
...
+                unique, counts = np.unique(coords, return_counts=True)
+                if np.max(counts) > mode_count:
+                    mode_count = np.max(counts)
+                    mode_value = unique[counts == mode_count]
+                    mode_dim = dim
+                        
+            # Cut all antennas not part of that mode to reduce the dimensionality of idealized_antpos 
+            new_ex_ants = [ant for ant in idealized_antpos if 
+                           np.abs(idealized_antpos[ant][mode_dim] - mode_value) > IDEALIZED_BL_TOL]
+            reds = filter_reds(reds, ex_ants=new_ex_ants)
```
anguta, line 256: The description really helps.

PR #593, opened 2020-04-22
Adds **kwargs to HERAData.read

nkern:
This adds **kwargs to HERAData.read signature to allow for passing extra kwargs upon read. This is motivated for reading in UVH5 files with multidimensional slicing (recent UVData addition). We could explicitly pass this kwarg, but it most general cases this kwarg should be False, and it makes sense to just add this ability for many of the other `UVData.read` kwargs. 

This does the same for `VisClean.write_data`, however, it had an existing `**kwargs` argument for updating additional `UVData` attributes before the write. This moves that to `extra_attrs={}` and makes `**kwargs` for passing additional kwargs to `UVData.write_`. This is a small API change--@jsdillon let me know if you think this is fine, or if we should wait on this and issue deprecation warnings. I don't think there are that many users of `VisClean` currently other than myself..

### DIFF
--- a/hera_cal/delay_filter.py
+++ b/hera_cal/delay_filter.py
@@ -91,7 +91,7 @@ class DelayFilter(VisClean):
         return filled_data, filled_flags
 
     def write_filtered_data(self, res_outfilename=None, CLEAN_outfilename=None, filled_outfilename=None, filetype='uvh5',
-                            partial_write=False, clobber=False, add_to_history='', **kwargs):
+                            partial_write=False, clobber=False, add_to_history='', extra_attrs={}, **kwargs):
         '''
         Method for writing data products.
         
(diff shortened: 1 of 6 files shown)


hera_cal/delay_filter.py
```diff
@@ -91,7 +91,7 @@ def get_filled_data(self):
         return filled_data, filled_flags
 
     def write_filtered_data(self, res_outfilename=None, CLEAN_outfilename=None, filled_outfilename=None, filetype='uvh5',
-                            partial_write=False, clobber=False, add_to_history='', **kwargs):
+                            partial_write=False, clobber=False, add_to_history='', extra_attrs={}, **kwargs):
```
jsdillon, line 94: I don't understand why we need kwargs for write_data. Which pyuvdata kwargs would you envision using that you don't think we should pass explicitly?

hera_cal/vis_clean.py
```diff
@@ -224,7 +224,8 @@ def write_data(self, data, filename, overwrite=False, flags=None, nsamples=None,
             partial_write : bool, if True, begin (or continue) a partial write to
             the output filename and store file descriptor in self.hd._writers.
             add_to_history : string, string to append to hd history.
-            kwargs : additional attributes to update before write to disk.
+            extra_attrs : additional attributes to update to HERAData before write
```
jsdillon, line 227: See above on wording

hera_cal/delay_filter.py
nkern: One example is the newly added `chunks` parameter, but also the HDF5 compression parameters, etc. Things that the average user would use the defaults in pyuvdata, but for us we might want access to it. It also gives us automatic access to any future kwarg in pyuvdata that we don't have explicitly here

PR #595, opened 2020-04-24
Update abscal to allow multiple data times to map to a single model time when appropriate

### LINKED ISSUE #594: Make abscal.get_d2m_time_map handle non-unique mappings
Currently `abscal.get_d2m_time_map` does not correctly handle non-unique mappings (i.e. where two times in the data map to a single time in the model). These cases result in one of the times in the data being mapped to `None`, when it seems like the correct behavior is for the times to map to the same time in the model. This is bound to happen when the integration times in the data and model do not exactly agree and there are enough integrations to allow for a double-up to occur.

jsdillon:
This is accomplished in two steps.

First `get_d2m_time_map` now allows multiple data times to map to a single model time. The current algorithm is that all data times map to the nearest model time in LST, but will map to `None` if they are more than half an LST step outside the range of LSTs covered by the model. This preserves the original functionality of flagging data outside the range of model files.

However, if multiple identical model times to load are passed to `io.partial_time_io`, the partial read will ignore duplicated times. To account for this, I've added a check to see if the model times match the ones we wanted to load. If they don't, we can use the new `select_or_expand_times` functionality of DataContainers to update themselves using fancy indexing, duplicating data where necessary.

This closes #594


hera_cal/abscal.py
```diff
@@ -3239,7 +3245,11 @@ def post_redcal_abscal_run(data_file, redcal_file, model_files, raw_auto_file=No
                         if not np.all(list(flags.values())):
                             # load model and rephase
                             model_times_to_load = [d2m_time_map[time] for time in hd.times[tinds]]
-                            model, model_flags, _ = io.partial_time_io(hdm, model_times_to_load, bls=model_bl_to_load)
+                            model, model_flags, _ = io.partial_time_io(hdm, np.unique(model_times_to_load), bls=model_bl_to_load)
```
r-pascua, line 3248: It may be worth the effort to not use `np.unique` here, since `np.unique` sorts the array it acts on. This has certainly caused problems with the validation work I'm doing, where the sim runs from 0-24 hrs LST linearly and the data has a phase wrap--the sorting done by `np.unique` means that the times are loaded in the wrong order, at least in my case (using the `select` method of the `HERAData` object).

hera_cal/datacontainer.py
```diff
@@ -433,3 +433,48 @@ def has_pol(self, pol):
...
+            in_place: if True, this DataContainer is modified. Otherwise, a modified copy is returned.
+        '''
+        if in_place:
+            dc = self
+        else:
+            dc = copy.deepcopy(self)
+
+        # make sure this is a sensible object for performing this operation
+        assert dc.times is not None
+        assert np.all([nt in dc.times for nt in new_times]), 'All new_times must be in self.times.'
```
r-pascua, line 454: Same comment as above about bare `assert` statements.

hera_cal/abscal.py
r-pascua: I suspect the easiest way to avoid this problem is to just do `model.select_or_expand_times(model_times_to_load)` regardless of whether times are repeated in `model_times_to_load`.

hera_cal/abscal.py
```diff
@@ -2790,41 +2790,46 @@ def get_all_times_and_lsts(hd, solar_horizon=90.0, unwrap=True):
...
-            each model time maps to at most one data time. Data times without corresponding
-            model times map to None.
+            that are closest in LST. Data times map to None when the nearest model LST is too far, 
+            as defined by the extrap_limit. 
     '''
-    if unwrap:  # avoid phase wraps
-        data_lsts[data_lsts < data_lsts[0]] += 2 * np.pi
-        model_lsts[model_lsts < model_lsts[0]] += 2 * np.pi
+    # check that the input is sensible
+    assert len(data_times) == len(data_lsts), 'data_times and data_lsts must have the same length.'
```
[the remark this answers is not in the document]

hera_cal/datacontainer.py
jsdillon: I've done this for the one where the input is bad. The other circumstances where the object is a very non-standard data container I think are rare enough that we're talking about a power user anyway.

hera_cal/abscal.py
jsdillon: The way this works under the hood, the order of the times that we're doing a select on doesn't matter, nor does the existence of duplicate times. I only put this here to be explicit that there might be duplicate times. Besides, this is acting on JDs not LSTs, so sorting would be OK anyway.

jsdillon:
Thanks @r-pascua!

PR #596, opened 2020-04-28
Smooth_cal blacklists in time and frequency

### LINKED ISSUE #537: Allow smoothcal to blacklist LSTs
Once we start absolute calibration against a simulation, we know that certain LSTs will have bad abscal solutions. We want to use smooth_cal to interpolate over these LSTs. Effectively, these will be treated as flags in the CLEAN algorithm, but not flags on the final calibration solutions.

jsdillon:
This PR implements a notion of blacklisted times/freqs in smooth_cal. These are not flagged, but they get 0 weight in the CLEAN algorithm, effectively forcing the calibration solution to interpolate or extrapolate over them. This is particularly useful for abscal vs. simulations with missing sources at known LSTs.

Here's some screenshots of it in action.

Screen Shot 2020-04-29 at 9 47 11 PM
Screen Shot 2020-04-29 at 9 47 20 PM

In particular, I want to highlight the two regions of strongest disagreement, the fornax transit and the galactic anticenter transit. In this plot of time-averaged calibration amplitude, it's clear that our blacklisted lsts, in this case 2.5-4.3h and 6.5-9.5h, leave us with much more stable gains.

Screen Shot 2020-04-29 at 9 47 46 PM

This closes #537 .

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -367,7 +367,7 @@ def build_time_blacklist(time_grid, time_blacklists=[], lst_blacklists=[], lat_l
     Returns:
         time_blacklist_array: boolean array with the same shape as time_grid in which blacklisted integrations are True'''
     
-    time_blacklist_array = np.zeros(length(time_grid), dtype=bool)
+    time_blacklist_array = np.zeros(len(time_grid), dtype=bool)
 
     # Calculate blacklisted times
     if len(time_blacklists) > 0:
@@ -562,8 +562,8 @@ class CalibrationSmoother():
                                      time_threshold=time_threshold, ant_threshold=time_threshold)
 
         # build blacklists
-        self.time_blacklist = build_time_grid_blacklist(self.time_grid, time_blacklists=time_blacklists, lst_blacklists=lst_blacklists,
-                                                        lat_lon_alt_degrees=lat_lon_alt_degrees, telescope_name=hc.telescope_name)
+        self.time_blacklist = build_time_blacklist(self.time_grid, time_blacklists=time_blacklists, lst_blacklists=lst_blacklists,
+                                                   lat_lon_alt_degrees=lat_lon_alt_degrees, telescope_name=hc.telescope_name)
         self.freq_blacklist = build_freq_blacklist(self.freqs, freq_blacklists=freq_blacklists, chan_blacklists=chan_blacklists)
 
         # pick a reference antenna that has the minimum number of flags (tie goes to lower antenna number) and then rephase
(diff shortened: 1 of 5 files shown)


hera_cal/smooth_cal.py
```diff
@@ -350,9 +351,102 @@ def rephase_to_refant(gains, refant, flags=None, propagate_refant_flags=False):
...
+    Returns:
+        time_blacklist_array: boolean array with the same shape as time_grid in which blacklisted integrations are True'''
+    
+    time_blacklist_array = np.zeros(len(time_grid), dtype=bool)
+
+    # Calculate blacklisted times
+    if len(time_blacklists) > 0:
+        for bounds in time_blacklists:
+            assert isinstance(bounds, Iterable) and len(bounds) == 2, 'time_blacklists must be list of pairs of bounds'
+            assert bounds[0] <= bounds[1], 'time_blacklist bounds must be in chronological order'
```
Chuneeta, line 377: We could lift the restriction from the user to input the time in chronological by doing this evaluation in the code itself.

hera_cal/smooth_cal.py
```diff
@@ -350,9 +351,102 @@ def rephase_to_refant(gains, refant, flags=None, propagate_refant_flags=False):
...
+            assert bounds[0] <= bounds[1], 'freq_blacklists bounds must be in ascending order'
+            freq_blacklist_array[(freqs >= bounds[0]) & (freqs <= bounds[1])] = True
+
+    # Calculate blacklisted channels
+    if len(chan_blacklists) > 0:
+        for bounds in chan_blacklists:
+            assert isinstance(bounds, Iterable) and len(bounds) == 2, 'chan_blacklists must be list of pairs of bounds'
+            assert bounds[0] <= bounds[1], 'chan_blacklists bounds must be in ascending order'
+            freq_blacklist_array[(np.arange(len(freqs)) >= bounds[0]) & (np.arange(len(freqs)) <= bounds[1])] = True
+
```
Chuneeta, line 432: Again the chronological order can be determined inside the code itself. This will alleviate user effort and it is straightforward to include.

hera_cal/smooth_cal.py
jsdillon: I think it'd be quite odd for the user to put the constraints in the opposite order; I think it's more likely that the user put the times out of order by mistake and I'd rather the code not fail silently. 

Also, if we don't enforce chronological order, then specifying 23-1 for LST blacklists becomes ambiguous whether you want to blacklist a total of 2 hours or a total of 22 hours.

So I'd prefer not to make this change, but I'm open to making the error message more clear if you have suggestions.

Chuneeta: Thanks. Okay, I think I agree with you here. Then for consistency better to enforce the time, frequency and channel inputs to be in chronological order except the lst since the lst ranges from 24-0 hours.

jsdillon:
Thanks @Chuneeta!

PR #597, opened 2020-05-05
Flexible interpolation

anguta:
Adding support for new linear interpolation methods in uvtools. 

Adding this support in vis_clean, delay_filter and smoothcal (not FRFilter). 

X-talk filtering will be added in the next PR

Also fixes compatibility with updates to uvtools. 

Addresses #598 #604


jsdillon:
Wanna let me know when tests are passing? Or make a new PR that passes tests given the uvtools changes?

anguta:
> Wanna let me know when tests are passing? Or make a new PR that passes tests given the uvtools changes?

All tests should be passing. For some reason, test_version.py is failing. It does not fail when I run it on my local machine. Is there an issue with travis?

ci/hera_cal_tests.yml
```diff
@@ -29,5 +28,5 @@ dependencies:
     - git+https://github.com/HERA-Team/linsolve
     - git+https://github.com/RadioAstronomySoftwareGroup/pyuvsim
     - git+https://github.com/RadioAstronomySoftwareGroup/pyradiosky
-    - git+https://github.com/HERA-Team/uvtools
+    - git+https://github.com/HERA-Team/uvtools.git@skip_wgt_clean_passing_fix
```
jsdillon, line 31: I'd undo this.

hera_cal/delay_filter.py
```diff
@@ -69,6 +74,35 @@ def run_filter(self, to_filter=None, weight_dict=None, standoff=15., horizon=1.,
                        window=window, gain=gain, skip_wgt=skip_wgt, edgecut_low=edgecut_low,
                        edgecut_hi=edgecut_hi, alpha=alpha, overwrite=True, verbose=verbose)
 
+    def run_dayenu_foreground_filter(self, to_filter=None, weight_dict=None, standoff=100., horizon=1., min_dly=0.0,
```
jsdillon, line 77: What's the point of this function?

hera_cal/delay_filter.py
```diff
@@ -180,6 +282,9 @@ def delay_filter_argparser():
     a.add_argument("--CLEAN_outfilename", default=None, type=str, help="path for writing the CLEAN model visibilities (with the same flags)")
     a.add_argument("--filled_outfilename", default=None, type=str, help="path for writing the original data but with flags unflagged and replaced with CLEAN models wherever possible")
     a.add_argument("--clobber", default=False, action="store_true", help='overwrites existing file at outfile')
+    a.add_argument("--spw_range", type=int, nargs=2, help="spectral window of data to foreground filter.")
```
jsdillon, line 285: If you're going to have multiple run scripts, you should have multiple argparsers or do something like: https://github.com/HERA-Team/hera_qm/blob/8ed3a97f9e37326295bf97df143ec803255b3d2c/hera_qm/utils.py#L24

As it stands, if you run the scripts with `-h` you'll get documentation for parameters that aren't used and aren't relevant, which is confusing.

scripts/dayenu_delay_filter_run.py
```diff
@@ -0,0 +1,58 @@
...
+
+# Run Delay Filter
+if a.partial_load_Nbls is not None:  # partial loading
+    delay_filter.partial_load_dayenu_delay_filter_and_write(a.infilename, calfile=a.calfile,
+                                                            Nbls=a.partial_load_Nbls, spw_range=a.spw_range,
+                                                            cache_dir=a.cache_dir, res_outfilename=a.res_outfilename,
+                                                            clobber=a.clobber, update_cache=a.update_cache,
+                                                            add_to_history=' '.join(sys.argv), **filter_kwargs)
+else:
+    cache_dir = a.cache_dir
```
jsdillon, line 30: This is way too much code to sit outside unit testing. All of this cache handling should happen in a run function inside `delay_filter.py`

hera_cal/delay_filter.py
```diff
@@ -166,7 +200,75 @@ def partial_load_delay_filter_and_write(infilename, calfile=None, Nbls=1,
...
+        clobber: if True, overwrites existing file at the outfilename
+        add_to_history: string appended to the history of the output file
+        write_cache: bool. If true, create new cache file with precomputed matrices
+                           that were not in previously loaded cache files.
+        read_cache: bool, If true, read existing cache files in cache_dir before running.
+        filter_kwargs: additional keyword arguments to be passed to DelayFilter.run_dayenu_foreground_filter()
+    '''
+    # Load up the cache file with the most keys (precomputed filter matrices).
+    cache = {}
+    if cache_dir is not None and read_cache:
```
jsdillon, line 229: you probably want to pull out this cache io and put it in io or even in uvtools. Looks like you use it a lot.

hera_cal/delay_filter.py
```diff
@@ -166,7 +200,75 @@ def partial_load_delay_filter_and_write(infilename, calfile=None, Nbls=1,
...
+                               clobber=clobber, add_to_history=add_to_history,
+                               freq_array=np.asarray([df.freqs]), Nfreqs=df.Nfreqs)
+        df.hd.data_array = None  # this forces a reload in the next loop
+    # Write a new cache file that only contains the newly computed filter matrices.
+    if write_cache:
+        # if the keys_before instantiation wasn't a list, then
+        # keys_before would just be the current keys of cache and we
+        # wouldn't have any new keys.
+        new_filters = {k: cache[k] for k in cache if k not in keys_before}
+        if len(new_filters) > 0:
```
jsdillon, line 264: Likewise

hera_cal/io.py
```diff
@@ -583,6 +585,28 @@ def partial_write(self, output_path, data=None, flags=None, nsamples=None, clobb
         else:  # make a copy of this object and then update the relevant arrays using DataContainers
             this = copy.deepcopy(self)
         this.update(data=data, flags=flags, nsamples=nsamples)
+        '''
```
jsdillon, line 588: delete

hera_cal/smooth_cal.py
```diff
@@ -79,34 +79,32 @@ def freq_filter(gains, wgts, freqs, filter_scale=10.0, tol=1e-09, window='tukey'
...
-        window: window function for filtering applied to the frequency axis.
-            See aipy.dsp.gen_window for options.
         skip_wgt: skips filtering rows with very low total weight (unflagged fraction ~< skip_wgt).
             filtered is left unchanged and info is {'skipped': True} for that time.
             Only works properly when all weights are all between 0 and 1.
-        maxiter: Maximum number of iterations for aipy.deconv.clean to converge.
-        win_kwargs : any keyword arguments for the window function selection in aipy.dsp.gen_window.
+        mode: deconvolution method to use. See uvtools.dspec.fourier_filter for full list of supported modes.
+              examples include 'dpss_leastsq', 'clean'.
+        filter_kwargs : any keyword arguments for the window function selection in aipy.dsp.gen_window.
```
jsdillon, line 87: This isn't true anymore, right? It's much more general now...

hera_cal/smooth_cal.py
```diff
@@ -79,34 +79,32 @@ def freq_filter(gains, wgts, freqs, filter_scale=10.0, tol=1e-09, window='tukey'
...
 
-    df = np.median(np.diff(freqs))  # in Hz
     filter_size = (filter_scale * 1e6)**-1  # Puts it in s
     dly = single_iterative_fft_dly(gains, wgts, freqs)  # dly in s
     rephasor = np.exp(-2.0j * np.pi * dly * freqs)
-    filtered, res, info = uvtools.dspec.high_pass_fourier_filter(gains * rephasor, wgts, filter_size, df, tol=tol, window=window,
-                                                                 skip_wgt=skip_wgt, maxiter=maxiter, **win_kwargs)
-    filtered /= rephasor
+
+    filtered, res, info = uvtools.dspec.fourier_filter(x=freqs, data=gains * rephasor, wgts=wgts, mode=mode, filter_centers=[0.],
```
jsdillon, line 100: Since this now supports a filter center other than zero, do you want to try removing the rephasor in favor of a using the `dly` as the filter center?

hera_cal/smooth_cal.py
```diff
@@ -664,8 +662,8 @@ def freq_filter(self, filter_scale=10.0, tol=1e-09, window='tukey', skip_wgt=0.1
                                                      filter_scale=filter_scale, tol=tol, window=window,
                                                      skip_wgt=skip_wgt, maxiter=maxiter, **win_kwargs)
             # flag all channels for any time that triggers the skip_wgt
-            for i, info_dict in enumerate(info):
-                if info_dict.get('skipped', False):
+            for i in info['status']['axis_1']:
```
jsdillon, line 665: The keys of this dictionary are integer indices? You're positive it shouldn't be for i in range(len(...?

hera_cal/tests/test_vis_clean.py
```diff
@@ -103,6 +115,80 @@ def test_read_write(self):
         assert os.path.exists("ex.uvh5")
         os.remove('ex.uvh5')
 
+    @pytest.mark.filterwarnings("ignore:.*dspec.vis_filter will soon be deprecated")
```
jsdillon, line 118: We probably don't want to filter deprecation warnings... just makes it easier to accumulate technical debt.

hera_cal/tests/test_vis_clean.py
```diff
@@ -173,6 +262,12 @@ def test_fft_data(self):
         pytest.raises(ValueError, V.fft_data, keys=[])
         pytest.raises(ValueError, V.fft_data, keys=[('foo')])
 
+    # THIS UNIT TEST IS BROKEN!!!
+    # It was cleaning with one dnu and trimming with another.
+    # I am preserving this bad behavior but when I remove the bad
+    # behavior it fails. Someone should fix this unit test.
+    # I don't think it should be done in this PR though -- whose goal is to
```
jsdillon, line 269: These kind of comments belong in an issue, not in the code.

scripts/dayenu_delay_filter_run.py
jsdillon: (or possibly in io.py)

hera_cal/vis_clean.py
```diff
@@ -43,13 +45,21 @@ def __init__(self, input_data, filetype='uvh5', input_cal=None, link_data=True,
                 as they are built.
             link_data : bool, if True, attempt to link DataContainers
                 from HERAData object, otherwise only link metadata if possible.
+            spw_range : range of frequencies to delay-filter / write out.
```
jsdillon, line 48: clarify the format and datatype expected here. Also clarify that it's channels and not frequencies. Also, why are we doing channels? hera_cal usually frequencies wherever possible

jsdillon: Also, what's the spw_range for? When is it used? Doesn't it make more spense to attach to a specific filter function or read function (it's already basically supported in the read functions). 

I'd remove this...

hera_cal/vis_clean.py
```diff
@@ -281,22 +307,419 @@ def write_data(self, data, filename, overwrite=False, flags=None, nsamples=None,
...
+                    filter_centers = filter_centers_time
+                    filter_half_widths = filter_half_widths_time
+            self.fourier_filter(keys=[k], filter_centers=filter_centers, filter_half_widths=filter_half_widths,
+                                suppression_factors=suppression_factors, mode='dayenu_dpss_leastsq',
+                                x=x, data=data, flags=flags, wgts=wgts, output_prefix=output_prefix,
+                                cache=cache, ax=ax,
+                                skip_wgt=skip_wgt, max_contiguous_edge_flags=max_contiguous_edge_flags,
+                                verbose=verbose, overwrite=overwrite)
+    ###
+    # TODO: zeropad here will error given its default option if 2d filtering is being used.
```
jsdillon, line 401: This should be an issue, not a TODO in the code. Otherwise we'll lose track of it.

hera_cal/vis_clean.py
```diff
@@ -281,22 +307,419 @@ def write_data(self, data, filename, overwrite=False, flags=None, nsamples=None,
...
+                    'skip_wgt' : skips filtering rows with very low total weight (unflagged fraction ~< skip_wgt).
+                        Model is left as 0s, residual is left as data, and info is {'skipped': True} for that
+                        time. Only works properly when all weights are all between 0 and 1.
+                    'gain': The fraction of a residual used in each iteration. If this is too low, clean takes
+                        unnecessarily long. If it is too high, clean does a poor job of deconvolving.
+                    'alpha': float, if window is 'tukey', this is its alpha parameter.
+        x : array-like, x-values of axes to be filtered. Numpy array if 1d filter.
+            2-list/tuple of numpy arrays if 2d filter.
+     """
+        if not HAVE_UVTOOLS:
```
jsdillon, line 573: uvtools is a hera_cal prereq. This is unnecessary.

ci/hera_cal_tests.yml
anguta: Will do this now that the uvtools fixes have been merged.

hera_cal/smooth_cal.py
anguta: That's true. I've updated the docstring to reflect this properly.

hera_cal/smooth_cal.py
anguta: I think that would be more elegant.

anguta: I've changed to filter_centers=[dly] and gotten rid of the rephasoring.

hera_cal/smooth_cal.py
anguta: The keys are integer indices so I'm happy to iterate through them. I am open to suggestions if you think this system is opaque.

hera_cal/tests/test_vis_clean.py
anguta: I've removed this.

hera_cal/tests/test_vis_clean.py
anguta: I also created an issue at the same time. https://github.com/HERA-Team/hera_cal/issues/603

I'll reference the issue here as well.

hera_cal/vis_clean.py
anguta: This is a stale TODO. Zeropadding is handled correctly now.

hera_cal/vis_clean.py
anguta: I'll remove this (and other instances from previous commits).

hera_cal/vis_clean.py
anguta: `spw_range` allows for the user to create a `vis_clean` object that only filters across a subband and saves data files that only include said subband. 

it is also used to select proper subbands for any attached `HERACal` objects. 

The alternative, that I think you are suggesting, is that one could initialize a vis_clean object with `link_data=False` and then run `read(freqs = frequencies_of_interest)` 

If we did things this way, there would be several extra steps for the user (rather then just initializing with an spw_range). They would have to call `read(freqs=frequencies_of_interest)` afterwards and we would have to keep the extra code in `apply_calibration` but now change from `spw_range` to the attached data frequencies. 

So I guess the tradeoff would be less code for different (depending on who you are less clear) usability. Since the rest of `io` uses `read` as a common way of downselecting data, I'm inclined to go with the `read` method instead of the `spw_range` method.

hera_cal/delay_filter.py
anguta: I think we can get rid of it.

hera_cal/smooth_cal.py
anguta: This actually passes unit tests for `dpss` fitting but not for `clean`. Potentially because of the discretized clean area. Rather then dive into this rabbit hole, I'm going to keep things as is.

hera_cal/delay_filter.py
anguta: I've split this parser up into several that are appropriate for different scripts.

scripts/dayenu_delay_filter_run.py
anguta: I've moved this to io.py

hera_cal/delay_filter.py
anguta: i've put it in io.

hera_cal/io.py
```diff
@@ -678,6 +658,61 @@ def iterate_over_times(self, Nints=1, times=None):
             yield self.read(times=times[i:i + Nints])
 
 
+def read_filter_cache(cache_dir):
+    """
+    Load files from a cache specified by cache_dir.
```
jsdillon, line 663: I'd give a little more details about when this is used and what these caches contain are. Also, if you talk to @plaplant he'll explain why pickles are a bad format (I forget the reasons precisely, I think they are insecure or platform-dependent or something). Anyway, I'm fine with this for now but maybe make an issue to establish some better (probably h5-based) data format?

hera_cal/io.py
```diff
@@ -678,6 +658,61 @@ def iterate_over_times(self, Nints=1, times=None):
...
+        cache_t = pickle.load(cfile)
+        for key in cache_t:
+            if key not in cache:
+                cache[key] = cache_t[key]
+    return cache
+
+
+def write_filter_cache(filter_cache, cache_dir=None, skip_keys=None):
+    """
+    write cached cache to a new cache file.
```
jsdillon, line 688: more documentation about what chaches are and what they're used for would be helpful

hera_cal/delay_filter.py
```diff
@@ -169,16 +148,24 @@ def write_filtered_data(self, res_outfilename=None, CLEAN_outfilename=None, fill
...
     '''
     Uses partial data loading and writing to perform delay filtering.
 
     Arguments:
         infilename: string path to data to uvh5 file to load
         cal: optional string path to calibration file to apply to data before delay filtering
-        Nbls: the number of baselines to load at once.
+        Nbls_per_load: the number of baselines to load at once.
+                       If None, load all baselines at once. default : None.
+        spw_range: spw_range of data to delay-filter.
```
jsdillon, line 163: You should specify the type

hera_cal/io.py
anguta: The current purpose of the cache files is to serve as a temporary disk scratch for filtering matrices to speed up calculations that are being conducted across a night (rather then long term cross-platform storage). The cache keys tend to be fairly complicated too (they often involve all x-values of a filtered data array and the values of weights) so I think pickling might be OK for this but I'm open to what @plaplant  thinks.

hera_cal/vis_clean.py
```diff
@@ -192,20 +172,19 @@ def apply_calibration(self, input_cal, unapply=False):
...
         cal_gains, cal_flags, cal_quals, cal_tquals = hc.read()
-        if self.spw_range is not None:
-            for ant in cal_gains:
-                cal_gains[ant] = cal_gains[ant][:, self.spw_range[0]:self.spw_range[1]]
-                cal_flags[ant] = cal_flags[ant][:, self.spw_range[0]:self.spw_range[1]]
-                cal_quals[ant] = cal_quals[ant][:, self.spw_range[0]:self.spw_range[1]]
-            if cal_tquals is not None:
-                for pol in cal_tquals:
-                    cal_tquals[pol] = cal_tquals[pol][:, self.spw_range[0]:self.spw_range[1]]
+        data_freqs = np.logical_and(hc.freqs >= self.data.freqs.min(), hc.freqs <= self.data.freqs.max())
```
jsdillon, line 180: I'd rename this variable something to clarify that it's a bool like `cal_freqs_in_data`

hera_cal/io.py
anguta: That said, a more robust format that stored the matrices on disk for the long term across nights would probably have major benefits for the RTP.

scripts/delay_filter_run.py
```diff
@@ -4,6 +4,7 @@
 # Licensed under the MIT License
 
 """Command-line drive script for hera_cal.delay_filter"""
+"""Only performs CLEAN Filtering"""
```
jsdillon, line 7: don't you mean to use `delay_clean_argparser`?

hera_cal/io.py
anguta: I've added a description for both descriptions.

hera_cal/delay_filter.py
anguta: specified.

scripts/delay_filter_run.py
anguta: Yep! Thanks for catching that.

hera_cal/vis_clean.py
anguta: I agree that this suggested variable name is much more readible. I've changed this.

PR #606, opened 2020-06-02
Cross talk filtering

anguta:
Added opm-able interfaces for clean and linear cross-talk filtering.


jsdillon:
@nkern since I reviewed @anguta's last big hera_cal PR and since this primarily touches your code, I'm going to bow out of this review if that's OK

nkern:
no worries, i'll take a stab at it :)

hera_cal/filter_parser.py
```diff
@@ -0,0 +1,132 @@
+from copy import deepcopy
```
nkern, line 1: seems like this could all go under `hera_cal.utils` or `hera_cal.io` rather than being its own submodule. Lmk what you think: I generally tend to have a high activation energy for a new module.

hera_cal/delay_filter.py
```diff
@@ -27,8 +25,8 @@ class DelayFilter(VisClean):
     See vis_clean.VisClean for CLEAN functions.
     """
 
-    def run_filter(self, to_filter=None, weight_dict=None, horizon=1., standoff=0.15, min_dly=0.0, mode='clean',
-                   skip_wgt=0.1, tol=1e-9, verbose=False, cache_dir=None, read_cache=False, write_cache=False, **filter_kwargs):
+    def run_delay_filter(self, to_filter=None, weight_dict=None, horizon=1., standoff=0.15, min_dly=0.0, mode='clean',
```
nkern, line 28: although this is technically a more informative name, i don't know if it is enough to warrant the API break. lmk your thoughts

hera_cal/delay_filter.py
```diff
@@ -82,71 +80,6 @@ def run_filter(self, to_filter=None, weight_dict=None, horizon=1., standoff=0.15
             if write_cache:
                 filter_cache = io.write_filter_cache_scratch(filter_cache, cache_dir, skip_keys=keys_before)
 
-    def get_filled_data(self):
```
nkern, line 85: also, is there a good reason to move this elsewhere? this will break a lot of existing pipeline. if you want to move the core code elsewhere that's fine, but I think we should leave shallow wrappers here with their docstrings for consistency.

hera_cal/xtalk_filter.py
```diff
@@ -0,0 +1,137 @@
...
+from .vis_clean import VisClean
+
+import pickle
+import random
+import glob
+import os
+import warnings
+
+
+class XTalkFilter(VisClean):
```
nkern, line 20: I think this should be a subclass of `delay_filter` and you shoudl leave the `get_filled_data` and `write_filled` data in the DelayFilter class

hera_cal/xtalk_filter.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    DelayFilter object.
```
nkern, line 22: outdated doc

scripts/xtalk_filter_run.py
```diff
@@ -0,0 +1,28 @@
...
+a = parser.parse_args()
+
+# set kwargs
+filter_kwargs = {'tol': a.tol, 'window': a.window, 'max_frate_coeffs': a.max_frate_coeffs, 
+                 'skip_wgt': a.skip_wgt, 'maxiter': a.maxiter, 'edgecut_hi': a.edgecut_hi,
+                 'edgecut_low': a.edgecut_low, 'gain': a.gain}
+if a.window == 'tukey':
+    filter_kwargs['alpha'] = a.alpha
+spw_range = a.spw_range
+# Run Delay Filter
```
nkern, line 24: update comment

hera_cal/filter_parser.py
anguta: It seems that most of the time, the argparsers stick with their code: abscal for example. In this case, we have two different pieces of code that do xtalk filtering or delay filtering and they both inherit vis_clean.

Maybe the right thing to do here is to put the internal argparsers in vis_clean and then we put the exposed argparsers in xtalk and delay filtering. What do you think?

hera_cal/delay_filter.py
anguta: I don't think the function is exposed so I'm not too worried about breaking the API since I think the name is more specific and readable in the context of us now supporting both delay and fringe-rate domain high pass filtering. If this does break the pipeline then I do agree we should reconsider.

hera_cal/delay_filter.py
anguta: I decided to move this here because it has functionality that is useful for all kinds of filtering (delay/fringe-rate domain) and since VisClean is the base class for this functionality I think it makes sense for it to be available for all code inheriting VisClean.

anguta: I'm not sure how this breaks the pipeline since `DelayFilter` inherits `VisClean` so anything that was calling this for a `DelayFilter` object should still be able to do so.

hera_cal/xtalk_filter.py
anguta: `DelayFilter` by it's very name does delay filtering so I much prefer a separate `xtalk_filter` module. `get_filled` and `write_filled` can just be moved to `VisClean` where any other filtering modules inherit them (they are useful in many contexts).

hera_cal/xtalk_filter.py
anguta: Thanks for catching this! Fixed it :).

hera_cal/delay_filter.py
anguta: OK. Here's an actual helpful response. 

Add a `run_filter` wrapper for `run_delay_filter` and a deprecation warning?

scripts/xtalk_filter_run.py
anguta: Done.

hera_cal/filter_parser.py
nkern: that sounds reasonable

hera_cal/delay_filter.py
nkern: ah you are right. yeah I thought some of the stuff in `get_filled_data` was specific to DelayFilter, but it is all consistent with functionality in `visclean` so this is probably fine!

hera_cal/xtalk_filter.py
nkern: ah yes, I agree--I think this is probably fine

hera_cal/delay_filter.py
nkern: Yeah I think that's a good idea, just have a shallow wrapper only in `DelayFilter.run_filter(**kwargs)` that then calls `self.run_delay_filter(**kwargs)` with a deprecation warning

anguta:
Thanks for the review @nkern! I've gone ahead and made the requested changes. I think this is ready for another look whenever you get a chance!

hera_cal/delay_filter.py
anguta: I've made this change.

hera_cal/delay_filter.py
anguta: Cool :)

hera_cal/delay_filter.py
```diff
@@ -193,58 +131,36 @@ def load_delay_filter_and_write(infilename, calfile=None, Nbls_per_load=None, sp
...
-    a.add_argument("--calfile", default=None, type=str, help="optional string path to calibration file to apply to data before delay filtering")
-    a.add_argument("--partial_load_Nbls", default=None, type=int, help="the number of baselines to load at once (default None means load full data")
-    a.add_argument("--res_outfilename", default=None, type=str, help="path for writing the filtered visibilities with flags")
-    a.add_argument("--clobber", default=False, action="store_true", help='overwrites existing file at outfile')
-    a.add_argument("--spw_range", type=int, default=None, nargs=2, help="spectral window of data to foreground filter.")
 
+def delay_filter_argparser(mode='clean'):
+    '''
+    Arg parser for commandline operation of delay filters.
+    Parameters
```
nkern, line 149: minor doc string formatting issue here: needs proper line breaks and indentation

hera_cal/vis_clean.py
```diff
@@ -6,6 +6,9 @@
 from collections import OrderedDict as odict
 import datetime
 from uvtools import dspec
+from copy import deepcopy
```
nkern, line 9: you've got `copy` module below, lets just use `copy.deepcopy` instead of re-importing

hera_cal/vis_clean.py
```diff
@@ -882,6 +930,28 @@ def _get_delta_bin(self, dtime=None, dnu=None):
 
         return dtime, dnu
 
+    def get_filled_data(self, prefix='clean'):
+        """Get data with flagged pixels filled with clean_model.
```
nkern, line 934: doc string line breaks

hera_cal/vis_clean.py
```diff
@@ -1196,3 +1266,81 @@ def noise_eq_bandwidth(window, axis=-1):
...
+# Here is an argparser with core arguments
+# needed for all types of xtalk and delay
+# filtering.
+# ------------------------------------------
+
+
+def _filter_argparser():
+    """
+    Core Arg parser for commandline operation of hera_cal.delay_filter and hera_cal.xtalk_filter
+    Returns
```
nkern, line 1280: line break

hera_cal/xtalk_filter.py
```diff
@@ -0,0 +1,162 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 the HERA Project
+# Licensed under the MIT License
+
+"""Module for delay filtering data and related operations."""
```
nkern, line 5: update this doc string

scripts/xtalk_filter_run.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python2.7
+# -*- coding: utf-8 -*-
+# Copyright 2019 the HERA Project
+# Licensed under the MIT License
+
+"""Command-line drive script for hera_cal.delay_filter"""
```
nkern, line 6: make this one string

scripts/xtalk_dayenu_filter_run.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python2.7
+# -*- coding: utf-8 -*-
+# Copyright 2019 the HERA Project
+# Licensed under the MIT License
+
+"""Command-line drive script for hera_cal.delay_filter"""
```
nkern, line 6: one string

nkern:
@anguta thanks for the update, a few more minor things.

hera_cal/delay_filter.py
anguta: I've added line break.

hera_cal/vis_clean.py
anguta: I've cleaned this up.

hera_cal/vis_clean.py
anguta: added break.

hera_cal/xtalk_filter.py
anguta: updated.

scripts/xtalk_filter_run.py
anguta: one string now.

scripts/xtalk_dayenu_filter_run.py
anguta: I've fixed this.

anguta:
Thanks for the followup @nkern. This should be good to go now!

PR #617, opened 2020-07-15
Update to use current pyuvdata handling of iers in conftest.py

jsdillon:
This might help issues over at https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/880

### DIFF
--- a/hera_cal/tests/conftest.py
+++ b/hera_cal/tests/conftest.py
@@ -14,15 +14,18 @@ from astropy.time import Time
 
 @pytest.fixture(autouse=True, scope="session")
 def setup_and_teardown_package():
-    # Try to download the lateest IERS table. If the download succeeds, run a
-    # computation that requires the values, so they are cached for all future
-    # tests. If it fails, turn off auto downloading for the tests and turn it
-    # back on in teardown_package (done by extending auto_max_age).
+    """Make data/test directory to put test output files in."""
+    # Do a calculation that requires a current IERS table. This will trigger
+    # automatic downloading of the IERS table if needed, including trying the
+    # mirror site in python 3 (but won't redownload if a current one exists).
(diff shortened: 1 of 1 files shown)


hera_cal/tests/conftest.py
```diff
@@ -14,15 +14,18 @@
 
 @pytest.fixture(autouse=True, scope="session")
 def setup_and_teardown_package():
-    # Try to download the lateest IERS table. If the download succeeds, run a
-    # computation that requires the values, so they are cached for all future
-    # tests. If it fails, turn off auto downloading for the tests and turn it
-    # back on in teardown_package (done by extending auto_max_age).
+    """Make data/test directory to put test output files in."""
```
bhazelton, line 17: Is this accurate? It doesn't look like you're making a test directory. I think it's copied from `pyuvdata`, which isn't doing that anymore either, so it should be fixed there too.

hera_cal/tests/conftest.py
```diff
@@ -14,15 +14,18 @@
...
+    # If there's not a current IERS table and it can't be downloaded, turn off
+    # auto downloading for the tests and turn it back on once all tests are
+    # completed (done by extending auto_max_age).
+    # Also, the checkWarnings function will ignore IERS-related warnings.
     try:
-        iers_a = iers.IERS_A.open(iers.IERS_A_URL)
         t1 = Time.now()
         t1.ut1
-    except(urllib.error.URLError):
+    except (Exception):
```
bhazelton, line 28: Ah, I think this might be the critical difference!

hera_cal/tests/conftest.py
jsdillon: No, it's not. I removed it.

PR #619, opened 2020-08-02
Set filter residuals and models in skipped integrations / channels equal to zero.

### LINKED ISSUE #618: make sure clean residuals are equal to zero in skipped integrations.
Otherwise, flagged residual channels contain artifacts. This can trip someone up if they aren't making sure to propagate clean_flags. There is an argument against this in that it is a "loud" signifier if someone does not carry clean flags forward. However, I think this "loud" signifier might just be confusing.

anguta:
Closes #618 Depends on #625

### DIFF
--- a/hera_cal/vis_clean.py
+++ b/hera_cal/vis_clean.py
@@ -695,9 +695,9 @@ class VisClean(object):
                             elif dim == 1:
                                 flgs[i] = True
 
-            filtered_model[k] = mdl
-            filtered_resid[k] = res
-            filtered_data[k] = mdl + res * fw
+            filtered_model[k] = mdl * ~flgs
+            filtered_resid[k] = res * (~flgs | ~f)
+            filtered_data[k] = filtered_model[k] + filtered_resid[k]
             filtered_flags[k] = flgs
             filtered_info[k] = info
 
--- a/hera_cal/vis_clean.py
+++ b/hera_cal/vis_clean.py
@@ -695,8 +695,10 @@ class VisClean(object):
                             elif dim == 1:
                                 flgs[i] = True
 
-            filtered_model[k] = mdl * ~flgs
-            filtered_resid[k] = res * (~flgs | ~f)
+            filtered_model[k] = mdl
+            filtered_model[k][flgs] = 0.
+            filtered_resid[k] = res
+            filtered_resid[k][flgs | f] = 0.
             filtered_data[k] = filtered_model[k] + filtered_resid[k]
             filtered_flags[k] = flgs
             filtered_info[k] = info
--- a/hera_cal/vis_clean.py
+++ b/hera_cal/vis_clean.py
@@ -685,22 +685,22 @@ class VisClean(object):
                         res, _ = zeropad_array(res, zeropad=zeropad[i], axis=i, undo=True)
                     _trim_status(info, i, zeropad[i - 1])
 
-            flgs = np.zeros_like(mdl, dtype=np.bool)
+            skipped = np.zeros_like(mdl, dtype=np.bool)
             for dim in range(2):
                 if len(info['status']['axis_%d' % dim]) > 0:
                     for i in range(len(info['status']['axis_%d' % dim])):
                         if info['status']['axis_%d' % dim][i] == 'skipped':
                             if dim == 0:
-                                flgs[:, i] = True
+                                skipped[:, i] = True
                             elif dim == 1:
-                                flgs[i] = True
+                                skipped[i] = True
 
             filtered_model[k] = mdl
-            filtered_model[k][flgs] = 0.
-            filtered_resid[k] = res
-            filtered_resid[k][flgs | f] = 0.
+            filtered_model[k][skipped] = 0.
+            filtered_resid[k] = res * fw
+            filtered_resid[k][skipped] = 0.
             filtered_data[k] = filtered_model[k] + filtered_resid[k]
-            filtered_flags[k] = flgs
+            filtered_flags[k] = skipped
             filtered_info[k] = info
 
         if hasattr(data, 'times'):
--- a/hera_cal/tests/test_vis_clean.py
+++ b/hera_cal/tests/test_vis_clean.py
@@ -128,8 +128,8 @@ class Test_VisClean(object):
         assert V.clean_info[k]['status']['axis_0'][0] == 'skipped'
         assert V.clean_info[k]['status']['axis_0'][3] == 'success'
         # check that clean resid is equal to zero in flagged channels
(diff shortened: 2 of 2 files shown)


nkern:
@anguta to be clear, this does not affect channels that weren't skipped?

anguta:
> @anguta to be clear, this does not affect channels that weren't skipped?
> One of the tests you add contains this line
> `# check that clean resid is equal to zero in flagged channels`
> which seems different than zeroing skipped integrations

That's correct. Filtering sets residuals to zeros in flagged channels and skipped channels. 

Before, it was only setting channels that were previously flagged (which is done at the `uvtools` level). This PR adds the skipped channels.

nkern:
@anguta okay, but it is actually important that we keep the model at the flagged integrations. can you explain what the motivation is for this? I see your issue, but that is only for skipped integrations, we shouldn't change the treatment of flagged channels

anguta:
> @anguta okay, but it is actually important that we keep the model at the flagged channels. can you explain what the motivation is for this? I see your issue, but that is only for skipped integrations, we shouldn't change the treatment of flagged channels

I think there may be some confusion

After this PR

The residual now has preflagged  (vis_clean.flags) and skipped (vis_clean.clean_flags) integrations / channels set to zero. 

The model still only has skipped (vis_clean.clean_flags) integrations / channels set to zero. pre-flagged model channels (vis_clean.flags) that were not skipped have the interpolated model values in them.

hera_cal/vis_clean.py
```diff
@@ -688,8 +688,10 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
                                 flgs[i] = True
 
             filtered_model[k] = mdl
+            filtered_model[k][flgs] = 0.
```
nkern, line 691: so `flgs` here is skipped integrations? Maybe we should change this to `skipped`?

anguta: yeah, `flgs` are the skipped integrations. I agree with changing the name ;)

hera_cal/vis_clean.py
```diff
@@ -688,8 +688,10 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
...
             filtered_resid[k] = res
-            filtered_data[k] = mdl + res * fw
+            filtered_resid[k][flgs | f] = 0.
```
nkern, line 693: perhaps to preserve continuity from how we've been doing it, you can use
```
filtered_resid[k] = res * fw
filtered_resid[k][flgs] = 0.0
```
which sets the residual at flagged channels to zero

nkern:
@anguta one thing to possibly add is a unit-test that ensures that the `filtered_data` is only altered in flagged (but not skipped) channels. I think this unit-test already exists, but it if it doesn't then it would be a good test. I'm fairly sure (but not certain) that your edits conform to this, so it would be good to have. Thanks!

hera_cal/vis_clean.py
anguta: I've adopted this convention.

anguta:
> @anguta one thing to possibly add is a unit-test that ensures that the `filtered_data` is only altered in flagged (but not skipped) channels. This unit-test may already exist, but if it doesn't then it would be a good test. I'm fairly sure (but not certain) that your edits conform to this, so it would be good to have. Thanks!

@nkern , good suggestion. I added this unittest and funnily enough it fails. Strangely enough, the problem is fixed when I change one line in `uvtools.dspec`. The line is where residuals in `dayenu_filter` are set to zero in flagged channels and integrations. i have no idea why the change actually fixes the problem or why it was failing in the first place since it seems to me that they should be doing the same thing. At any rate, there is now a two line PR. Once this is merged, then the unittests checking that the data and filtered data are the same outside of flagged / skipped integrations / channels should pass. 

here is the PR

https://github.com/HERA-Team/uvtools/pull/102

cheers,

-Aaron

anguta:
> > @anguta one thing to possibly add is a unit-test that ensures that the `filtered_data` is only altered in flagged (but not skipped) channels. This unit-test may already exist, but if it doesn't then it would be a good test. I'm fairly sure (but not certain) that your edits conform to this, so it would be good to have. Thanks!
> 
> @nkern , good suggestion. I added this to the unittests and funnily enough it fails. The problem is fixed when I change one line in `uvtools.dspec`. The line is where residuals in `dayenu_filter` are set to zero in flagged channels and integrations. i have no idea why the change actually fixes the problem or why it was failing in the first place since it seems to me that they should be doing the same thing and there are unittests in place in `uvtools.dspec` to ensure that this zero setting worked (which it was passing).
> 
> the change that makes the unittests pass is in line 831 / 832 of `uvtools.dspec`. The change is
> 
> `output[wgts == 0.] = 0.` to `output = output * (np.abs(wgts) > 0.).astype(float)`
> 
> At any rate, there is now a two line PR. Once this is merged, then the unittests checking that the data and filtered data are the same outside of flagged / skipped integrations / channels should pass.
> 
> here is the PR
> 
> [HERA-Team/uvtools#102](https://github.com/HERA-Team/uvtools/pull/102)
> 
> cheers,
> 
> -Aaron

I've also added some unittests to cover `dpss` mode. `data` and `clean_data` disagree in unflagged integrations and channels at the $10^{-6}$ level. I can't figure out why so I've made this an outstanding issue.

nkern:
@anguta very possibly it is the float-to-float comparison problem. anytime you want to compare a float you should use np.isclose, not a == b.

also you said the issue was with `dayenu_filter`--is this also a problem with the Fourier Filter method? Or was it just a problem with the dayenu wrapper? thanks

anguta:
> @anguta very possibly it is the float-to-float comparison problem. anytime you want to compare a float you should use np.isclose, not a == b.
> 
> also you said the issue was with `dayenu_filter`--is this also a problem with the Fourier Filter method? Or was it just a problem with the dayenu wrapper? thanks

Hard to say. I was already doing the `isclose` method. The `atol` that the failure was happening at is surprisingly large (for numerical issues), about `1e-6`. Not an EoR issue but I did find it surprising for numerical errors.

At any rate, this should be ready to go once the `uvtools` PR is merged.

nkern:
@anguta not sure if you are done with this. lmk if you are finished and I'll have one last look. thanks

anguta:
> @anguta not sure if you are done with this. lmk if you are finished and I'll have one last look. thanks

Thanks for checking in. I've decided to wait until after merging #625 . I'll let you know when it's ready!

cheers,

-Aaron

nkern (approved):
I think this looks good. thanks aew!

PR #625, opened 2020-08-14
Apply cal red averaging

### LINKED ISSUE #623: optional redundant averaging during apply_cal
Give users the option of computing and saving redundantly averaged visibilities after applying a calibration solution.

anguta:
Closes #623


hera_cal/apply_cal.py
```diff
@@ -192,6 +196,10 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
         flag_file: optional path to file containing flags to be ORed with flags in input data. Must have
             the same shape as the data.
         flag_filetype: filetype of flag_file to pass into io.load_flags. Either 'h5' (default) or legacy 'npz'.
+        a_priori_ex_ants_yaml : path to YAML with antenna flagging information parsable by
```
jsdillon, line 199: Why are we ignoring frequency/time flags? Wouldn't that be useful?

Also, be very careful with ex_ants when `redundant_solution=True`, since you might flag baselines that are actually good but have excluded antennas in the "exemplar" baseline for the redundant group.

hera_cal/apply_cal.py
```diff
@@ -206,10 +214,21 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+            Datacontainer containing weights to use in redundant averaging.
+            only used if redundant_average is True.
+            Default is None.
         kwargs: dictionary mapping updated UVData attributes to their new values.
             See pyuvdata.UVData documentation for more info.
     '''
-
+    if a_priori_ex_ants_yaml is not None:
+        from hera_qm.metrics_io import read_a_priori_ant_flags
+        ex_ants = ex_ants.union(set(read_a_priori_ant_flags(a_priori_ex_ants_yaml, ant_indices_only=True)))
```
jsdillon, line 228: Uh, I don't think ex_ants is ever defined...

hera_cal/apply_cal.py
```diff
@@ -227,17 +246,44 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+                    if bl in hd.antpairs or bl[:-1] in hd.antpairs:
+                        nbls += 1
+                        data_grp.append(bl)
+                if nbls > 0:
+                    # keep list of zeroth baselines of each group.
+                    red_data_bls.append(data_grp[0])
+            # couldn't get a system working where we just read in the outputs one at a time.
+            # so unfortunately, we have to load one baseline per redundant group.
+            hd_red.read(bls=red_data_bls)
+            hd_red.select(bls=red_data_bls)
```
jsdillon, line 282: This select shouldn't be necessary.

jsdillon: A thing to note, HERAData class attribute will reflect the whole file, even after select. This is intentional, so it's easier to do partial i/o. UVData class attributes will not.

hera_cal/apply_cal.py
```diff
@@ -227,17 +246,44 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+            # have baselines in the data. Each group is still labeled by the
+            # first baseline of each group regardless if that baseline is in
+            # the data file.
+            red_data_bls = []
+            for grpnum, grp in enumerate(all_red_antpairs):
+                nbls = 0
+                data_grp = []
+                for bl in grp:
+                    if bl in hd.antpairs or bl[:-1] in hd.antpairs:
+                        nbls += 1
```
jsdillon, line 274: Seems like this can be done easier with a `break` statement, rather than accumulating more variables you don't need like `data_grp` and `nbls`... like if `np.any([bl in hd.antpairs for bl in grp]) or np.any([bl[::-1] in hd.antpairs for bl in grp])`

hera_cal/apply_cal.py
```diff
@@ -227,17 +246,44 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+            # go through all redundant groups and remove the groups that do not
+            # have baselines in the data. Each group is still labeled by the
+            # first baseline of each group regardless if that baseline is in
+            # the data file.
+            red_data_bls = []
+            for grpnum, grp in enumerate(all_red_antpairs):
+                nbls = 0
+                data_grp = []
+                for bl in grp:
+                    if bl in hd.antpairs or bl[:-1] in hd.antpairs:
```
jsdillon, line 273: This should be `bl[::-1]` no?

hera_cal/apply_cal.py
```diff
@@ -227,17 +246,44 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+    no_red_weights = redundant_weights is None
     # partial loading and writing using uvh5
     if nbl_per_load is not None:
         if not ((filetype_in == 'uvh5') and (filetype_out == 'uvh5')):
             raise NotImplementedError('Partial writing is not implemented for non-uvh5 I/O.')
-        hd = io.HERAData(data_infilename, filetype='uvh5')
         for attribute, value in kwargs.items():
             hd.__setattr__(attribute, value)
-        for data, data_flags, _ in hd.iterate_over_bls(Nbls=nbl_per_load):
+        all_reds = redcal.get_reds(hd.antpos, pols=hd.pols, bl_error_tol=bl_error_tol, include_autos=True)
```
jsdillon, line 258: If `redundant_average` is false and `redundant_solution` is false, I'd suggest not running this line.

hera_cal/apply_cal.py
```diff
@@ -206,10 +214,21 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
         add_to_history: appends a string to the history of the output file. This will preceed combined histories
             of flag_file (if applicable), new_calibration and, old_calibration (if applicable).
         clobber: if True, overwrites existing file at outfilename
+        redundant_average : bool, optional
+            If True, redundantly average calibrated data and save to <data_outfilename>.red_avg.<filetype_out>
+        redundant_weights : datacontainer, optional.
+            Datacontainer containing weights to use in redundant averaging.
+            only used if redundant_average is True.
+            Default is None.
```
jsdillon, line 222: Explain that if the default is None, then the nsamples is used.

hera_cal/apply_cal.py
```diff
@@ -246,20 +292,35 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+                # redundantly average
+                utils.red_average(data=data, flags=data_flags, nsamples=data_nsamples, reds=all_red_antpairs, wgts=redundant_weights, inplace=True)
+                # update redundant data. Don't partial write.
+                hd_red.update(nsamples=data_nsamples, flags=data_flags, data=data)
+            else:
+                # partial write works for no redundant averaging.
+                hd.partial_write(data_outfilename, inplace=True, clobber=clobber, add_to_history=add_to_history, **kwargs)
+        if redundant_average:
+            # if we did redundant averaging, just write the redundant dataset out in the end at once.
+            hd_red.write_uvh5(data_outfilename, clobber=clobber)
```
jsdillon, line 319: I think you should pass through ` **kwargs`

hera_cal/apply_cal.py
```diff
@@ -268,16 +329,31 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+            if no_red_weights:
+                redundant_weights = copy.deepcopy(data_nsamples)
+                for bl in data_flags:
+                    if np.all(data_flags[bl]):
+                        redundant_weights[bl][:] = 0.
+
+            utils.red_average(hd, reds=all_red_antpairs, inplace=True, wgts=redundant_weights)
+            if filetype_out == 'uvh5':
+                # overwrite original outfile with
+                hd.write_uvh5(data_outfilename, clobber=clobber)
```
jsdillon, line 354: I think you should pass through ` **kwargs`

hera_cal/apply_cal.py
```diff
@@ -268,16 +329,31 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
         else:
             calibrate_in_place(data, new_gains, data_flags=data_flags, cal_flags=new_flags,
                                old_gains=old_gains, gain_convention=gain_convention)
-
-        io.update_vis(data_infilename, data_outfilename, filetype_in=filetype_in, filetype_out=filetype_out,
-                      data=data, flags=data_flags, add_to_history=add_to_history, clobber=clobber, **kwargs)
+        if not redundant_average:
+            io.update_vis(data_infilename, data_outfilename, filetype_in=filetype_in, filetype_out=filetype_out,
+                          data=data, flags=data_flags, add_to_history=add_to_history, clobber=clobber, **kwargs)
+        else:
```
jsdillon, line 340: There's a lot of copy-pasted code from above. Would it make sense to make a helpful function, perhaps a "leading underscore" function with a very short docstring?

hera_cal/tests/test_io.py
```diff
@@ -480,6 +480,20 @@ def test_iterate_over_bls(self):
         with pytest.raises(NotImplementedError):
             next(hd.iterate_over_bls())
 
+        hd = HERAData(self.uvh5_1)
```
jsdillon, line 483: This isn't really testing the redundancy of the chunks...

Also, I'd suggest testing Nbls < len(red).

jsdillon: (if it's not too hard, your call. I see that a lot of this is tested in `test_chunck_baselines_by_redundant_group`)

hera_cal/tests/test_utils.py
```diff
@@ -663,3 +663,40 @@ def test_echo(capsys):
...
+            [(24, 25), (37, 38), (38, 39), (52, 53)],
+            [(24, 52), (25, 53)],
+            [(25, 37), (39, 52)],
+            [(24, 39), (37, 53)],
+            [(37, 39)],
+            [(25, 52)],
+            [(24, 53)]]
+    chunked_by_four_expected = [reds[0], reds[1], reds[2], reds[3],
+                                reds[4] + reds[5], reds[6] + reds[7] + reds[8], reds[9]]
+    # unwravel redundant group.
```
jsdillon, line 693: typo

hera_cal/utils.py
```diff
@@ -1251,7 +1252,9 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
             navg = np.sum(n * f, axis=0)         # this is the new total nsample (without flagged elements)
             fmax = np.max(f, axis=2)             # collapse along freq: marks any fully flagged integrations
             iavg = np.sum(tint.squeeze() * fmax, axis=0) / np.sum(fmax, axis=0).clip(1e-10, np.inf)
-            favg = np.isclose(wsum, 0.0)         # this is getting any fully flagged pixels
+            # average flags should be where weights x flags x nsamples sum to zero.
+            wfsum = np.sum(n * w * f, axis=0).clip(1e-10, np.inf)
```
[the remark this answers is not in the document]
anguta: I don't think it's necessary.

hera_cal/utils.py
```diff
@@ -1274,6 +1277,8 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
         for bl in list(data.keys()):
             if bl not in bls:
                 del data[bl]
+                del flags[bl]
```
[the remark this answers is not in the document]
anguta: I wasn't aware of this. I'll consolidate.

hera_cal/utils.py
```diff
@@ -1311,3 +1316,73 @@ def top2eq_m(ha, dec):
...
+    This method calculates chunks of baselines to load simultaneously constrained by the conditions that
+    all baselines simultaneously loaded completely fill out redundant groups and the number of baselines
+    loaded is less then or equal to max_chunk_size.
+    If a redundant group exceeds max_chunk_size it will be loaded in its entirety anyways
+    and warning will be raised.
+
+    If is greater then the number of baselines in a redundant group
+    then return consecutive redundant groups with total baseline count
+    less then or equal to max_chunk_size.
+    If max_chunk_size is smaller then the number of baselines in a redundant group
```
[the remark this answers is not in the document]
anguta: removed.

hera_cal/utils.py
```diff
@@ -1311,3 +1316,73 @@ def top2eq_m(ha, dec):
...
+    then still return that group but raise a Warning.
+
+    Parameters
+    ----------
+    bls : list
+        list of antpairs or antpairpols of all baselines to chunk.
+    reds : list
+        list of lists with redundant groups of antpairpols
```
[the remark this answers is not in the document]
anguta: It sounds like you are suggesting that reds can be pre-filtered to only include the baselines hat we care about. I can reorganize things to do this.

hera_cal/utils.py
```diff
@@ -1311,3 +1316,73 @@ def top2eq_m(ha, dec):
...
+
+    Returns
+    -------
+
+    chunked_baselines : list
+        list of lists of antpairpols where each chunk containes entire redundant groups
+        and (if possible) includes a number of baselines less then max_chunk_size.
+    """
+    # trim reds to only include baselines in provided bls list (data).
+    reds_data = [[bl for bl in blg if bl in bls] for blg in reds]
```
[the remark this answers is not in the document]
anguta: I've moved the bls out of this function and now call `filter_reds` on `reds` before passing it in.

hera_cal/apply_cal.py
anguta: added this.

hera_cal/apply_cal.py
anguta: added

hera_cal/apply_cal.py
anguta: I could identify about twelve lines that could readily be moved to a function. Given the additional lines required for docstrings and the reduction in readability, i think i'm going to keep this as is.

hera_cal/utils.py
nkern, line 1256: this seems inconsistent--`favg` should be where `wsum==0`. If someone wants the weights to be nsample * ~flag (by default) then this is what `wsum` is equal to. In the case where the wgts are directly fed, the docstring says that this supercedes nsample and flags, in which case favg (as is written) is inconsistent with what was used to actually weight the data. what is the rationale for changing this?

anguta: The rational here is to preserve flags even when the weights are not zero where the data is flagged. Without this change, a user could provide `w` that is not zero where the data is flagged and this will effectively override the previous flags.

anguta: Because `navg` is not determined by weights, even when the user provides them, I find the most logical interpretation of the docstring to mean that user provided weights supercede `nsamples` and `flags` as weights but do not replace `nsamples` and `flags`.

hera_cal/tests/test_io.py
anguta: I can check redundancy here without much difficulty.

anguta: But it's only easy when `Nbls=1` so in all cases `Nbls < len(red)`.

hera_cal/tests/test_utils.py
anguta: thanks for catching this!

hera_cal/utils.py
anguta: At any rate, I do totally understand the concern for tweaks that change the API behavior (especially for `H1C`).

 Let me know if I should keep this change (and update the docstring?) or whether you'd like this change in behavior to be optional (so as not too interfere with `H1C`).

anguta:
Thanks for the review @jsdillon , I think this is ready for another look!

hera_cal/utils.py
```diff
@@ -1311,3 +1316,73 @@ def top2eq_m(ha, dec):
...
+            raise ValueError("All baselines provided in bls must also be in reds!")
+    # now iterate through redundancies.
+    baseline_chunks = []
+    # now iterate through redundant groups.
+    group_index = 0
+    while group_index < len(reds_data):
+        # if red group is larger then the chunk size.
+        # then give a warning and treate the red group as a chunk anyways.
+        if len(reds_data[group_index]) > max_chunk_size:
+            warnings.warn("Warning: baseline group encountered with number"
```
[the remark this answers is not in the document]
anguta: I've added this information to the warning.

hera_cal/utils.py
```diff
@@ -1311,3 +1316,73 @@ def top2eq_m(ha, dec):
...
+            reds_data_flattened.append(bl)
+    # check that baselines provided fall into one of the redundant groups in reds
+    for bl in bls:
+        if bl not in reds_data_flattened:
+            raise ValueError("All baselines provided in bls must also be in reds!")
+    # now iterate through redundancies.
+    baseline_chunks = []
+    # now iterate through redundant groups.
+    group_index = 0
+    while group_index < len(reds_data):
```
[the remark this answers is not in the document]
anguta: simplified.

hera_cal/utils.py
nkern: Its not an H1C API issue for me. I actually think that it is inconsistent with the averaging: it seems that `favg` should be consistent with the averaging actually done to the data. If you pass a `wgts` dictionary, you are effectively ignoring the flags, so the "flags" in your weighting are set by when `np.isclose(wgts, 0)`. I understand that is it weird to ignore the flags, but that is the spirit of the functionality enabled by the `wgts` dictionary. If we want to remove that functionality that is a separate question, but I think `favg` should be consistent with what we weight by. Lmk if that makes sense.

Yeah the nsamples is an outlier, that might be changed to
`navg = np.sum(n * ~np.isclose(w, 0),  axis=0)`

hera_cal/io.py
```diff
@@ -611,8 +627,39 @@ def iterate_over_bls(self, Nbls=1, bls=None):
...
+                if isinstance(self.pols, list):
+                    pols = self.pols
+                else:
+                    for k in self.pols:
+                        for pol in self.pols[k]:
+                            if pol not in pols:
+                                pols.append(pol)
+
+                reds = redcal.get_reds(antpos, pols=pols, bl_error_tol=bl_error_tol,
+                                       include_autos=True)
```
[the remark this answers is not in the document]
anguta: I've exposed it.

hera_cal/io.py
```diff
@@ -611,8 +627,39 @@ def iterate_over_bls(self, Nbls=1, bls=None):
...
+        else:
+            if reds is None:
+                if self.filetype != 'uvh5':
+                    raise NotImplementedError('Redundant group iteration without explicitly setting antpos for filetype ' + self.filetype
+                                              + ' without setting antpos has not been implemented.')
+                antpos = {}
+                pols = []
+                # generate antpos dict to feed into get_reds
+                # that accounts for possibility that
+                # HERACal was initialized from multiple
```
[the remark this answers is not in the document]
anguta: Fixed.

hera_cal/io.py
```diff
@@ -611,8 +627,39 @@ def iterate_over_bls(self, Nbls=1, bls=None):
...
-        for i in range(0, len(bls), Nbls):
-            yield self.read(bls=bls[i:i + Nbls])
+        if not chunk_by_redundant_group:
+            baseline_chunks = [bls[i:i + Nbls] for i in range(0, len(bls), Nbls)]
+        else:
+            if reds is None:
+                if self.filetype != 'uvh5':
+                    raise NotImplementedError('Redundant group iteration without explicitly setting antpos for filetype ' + self.filetype
+                                              + ' without setting antpos has not been implemented.')
+                antpos = {}
```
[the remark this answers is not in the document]
anguta: That simplified things greatly :)

hera_cal/apply_cal.py
anguta: I can add time / frequency flags.

hera_cal/apply_cal.py
anguta: I've fixed this.

hera_cal/apply_cal.py
anguta: I think that the way I'm flagging ex_ants, (setting flags of ex_ants True but not excluding them from reds) should make sure there aren't issues with removing the exemplary baseline as a key to the red groups.

hera_cal/apply_cal.py
anguta: I've removed the selection.

hera_cal/apply_cal.py
anguta: Now that I know about `filter_reds`, this is a one-liner.

hera_cal/apply_cal.py
anguta: yes. This code has been removed.

hera_cal/apply_cal.py
anguta: I've moved this into an appropriate if statement.

hera_cal/apply_cal.py
anguta: added this clarification.

anguta:
@jsdillon , I think this is ready for another look.

hera_cal/utils.py
jsdillon: I think I see where Nick is coming from here. 

Perhaps the compromise here is to make the function error if wgts is provided and flags or nsamples is also provided. And maybe the docstring can be updated for clarity.

anguta: A mode of use that I do want to support is for a user to redundantly average data, not set that data to zero, but still have flags in those integrations.

anguta: having weights override flags makes this impossible. The way my update is written still flags regions where the weights are zero, it simply preserves the flags that are input. If someone wants flags to only occur where weights are zero, they can provide flags that are False and weights that are zero where they wish for the flags to occur. 

I don't see though how we can allow for redundantly averaged data that is not set to zero in flagged regions using the method as written though. Perhaps a `flagging_mode` keyword can be used to fix this.

anguta: A way that this could be done would be to perform the averaging with the user provided weights and then restore the flags afterwards. I'm open to writing a wrapper for this.

nkern: Well, the whole idea of `wgts` is to allow the user to do what they want. If they want to use the flags but have interesting, non-standard wgts, they can simply construct that themselves (`wgts = my_wgt * ~flgs`)--that's the point of a user-fed option.

anguta: If one does this, then they will preserve the flags, but the data will now be zero in the flagged channels so by choosing to keep the flags, the user has been forced to throw away the data in those flagged channels.

nkern: I think that goes against the core philosophy of what a flag represents though. If someone wants to not weight by the flags, that's exactly what `wgts` is trying to enable--weights set by the user. At the same time, if one wants to keep the initial set of flags, they can keep them. nothing is stopping them, right? But, for the function itself to report that a pixel is "flagged" (via `favg`) but is actually not set to zero (or vice versa for that matter) is inconsistent. I mean, I think what you are asking for is perhaps achievable by making _another_ array, say `avg_inp_flags`, which takes the average of the input flags, regardless of what `wgts` is, however, this should not replace `favg` in my opinion.

anguta: I agree that it's kind of weird philosophically but I do think that it's the right kind of redundant averaging for us to do until we are super comfortable with our flags. I toyed around with the wrapper approach and things get kind of messy (you have to compute redundant baselines and filter them within the wrapper) so I think the best way forward is to add an argument to `red_average` that is defaulted to the case where we have weights determine flags and compute nsamples as you suggest: `navg = np.sum(n * ~np.isclose(w, 0), axis=0)`. In the non-default case it will compute `favg` as the product of `n`, `w`, and `f`.

hera_cal/utils.py
```diff
@@ -1158,7 +1160,12 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
         nsamples : DataContainer
             If data is a DataContainer, these are its nsamples. Default (None) is 1.0 for all pixels.
             Furthermore, if data is a DataContainer, integration_time is 1.0 for all pixels.
-
+        user_weights_determine_avg_flags : bool, optional
```
nkern, line 1163: I think an easier and clearer way to get what you want is to enable this functionality
```
apply_flags_to_wgts : bool, optional
   If True and if wgts is fed, multiply inverse of flags (if provided) into wgts dictionary
```
which does this
```
if apply_flags_to_wgts:
    for k in wgts:
        wgts[k] *= (~flags[k]).atype(np.float)
```
Now you don't have to have two different kinds of output--there is only `favg`, but this `favg` is what you @anguta want when you turn on `apply_flags_to_wgts`. lmk if that makes sense

hera_cal/utils.py
```diff
@@ -1158,7 +1160,13 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
+            If this argument is True, the the flags of the redundantly averaged data
```
nkern, line 1164: then the

hera_cal/utils.py
```diff
@@ -1170,7 +1178,18 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
     Notes:
         1. Different polarizations are assumed to be non-redundant.
         2. Default weighting is nsamples * integration_time * ~flags.
-        3. If wgts Container is fed then they supercede flag and nsample weighting.
+        3. If wgts Container is fed then the behavior of flags and nsamples in the redundantly averaged sum will depend on wheter
```
nkern, line 1181: whether

hera_cal/utils.py
```diff
@@ -1248,10 +1267,16 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
             # take the weighted average
             wsum = np.sum(w, axis=0).clip(1e-10, np.inf)  # this is the normalization
             davg = np.sum(d * w, axis=0) / wsum  # weighted average
-            navg = np.sum(n * f, axis=0)         # this is the new total nsample (without flagged elements)
```
nkern, line 1251: I actually think this should be changed to
```
binary_wgts = (~np.isclose(w, 0)).astype(np.float)  # binary weights where wgts == 0
navg = np.sum(n * binary_wgts, axis=0) / np.sum(binary_wgts, axis=0).clip(1e-10, np.inf)
```

hera_cal/utils.py
```diff
@@ -1248,10 +1267,16 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
             fmax = np.max(f, axis=2)             # collapse along freq: marks any fully flagged integrations
             iavg = np.sum(tint.squeeze() * fmax, axis=0) / np.sum(fmax, axis=0).clip(1e-10, np.inf)
-            favg = np.isclose(wsum, 0.0)         # this is getting any fully flagged pixels
+            # average flags should be where weights x flags x nsamples sum to zero.
```
nkern, line 1272: If we go with the `propagate_flags` kwarg, you can change this to
```
favg = np.isclose(wsum, 0.0)
if propagate_flags:
    favg = np.logical_or(favg, np.all(np.isclose(f, 0), axis=0))
```

hera_cal/apply_cal.py
```diff
@@ -308,7 +307,9 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
                         if np.all(data_flags[bl]):
                             redundant_weights[bl][:] = 0.
                 # redundantly average
-                utils.red_average(data=data, flags=data_flags, nsamples=data_nsamples, reds=all_red_antpairs, wgts=redundant_weights, inplace=True)
+                utils.red_average(data=data, flags=data_flags, nsamples=data_nsamples,
+                                  reds=all_red_antpairs, wgts=redundant_weights, inplace=True,
+                                  propagate_flags=True)
```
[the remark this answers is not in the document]
anguta: @nkern, @jsdillon  and I decided that putting the redundant averaging inside of `apply_cal` was the best option to deal with partial i/o. We need to be able to redundantly average partially loaded chunks of baselines that have just had their calibration solution applied before partially writing them and loading another chunk. I don't see how we can do this in a completely separate RTP step without saving the calibrated chunks into a bunch of small files which I think would be much messier.

hera_cal/utils.py
```diff
@@ -1244,14 +1260,24 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
                 n = np.asarray([data.get_nsamples(bl + (pol,)) for bl in blg])
                 tint = np.asarray([data.integration_time[data.antpair2ind(bl + (pol,))] for bl in blg])[:, :, None]
                 w = np.asarray([wgts[bl + (pol,)] for bl in blg]) * tint
-
+            if propagate_flags:
+                # if propagate flags is True, then set all of w to zero on fully flagged baselines
+                # even if the weights were provided by the user.
+                w[np.asarray([np.all(np.isclose(fslice, 0.0)) for fslice in f])][:] = 0.0
```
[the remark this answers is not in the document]
anguta: > I think this is wrong. The point of the `wgts` dictionary is to have _total control_ of `w` by the user. we should not be manipulating `wgts` if they are fed. `propagate_flags` is simply going to change the `favg` output, but should not change `wgts`.
> 
> I see why you did this though, which is because in practice, you (@anguta), want the specific functionality of not setting `w->0` when all baselines are flagged. Therefore, setting `wgts = n * ~f` and then feeding `red_average(..., wgts=wgts)` would not accomplish this, and needs to be "unflagged" where `f` is all True in order for the wgts to not be zero. However, I think my point here is that _you need to do that yourself_ outside of the function. In my mind, the point of `red_average` (or any general purpose function for that matter) is to support broad functionality, and when you want to implement a specific use-case you should conform your data to make that work outside the function, which in this case is fairly easy anyways, as its a one-liner!

@nkern, this is already implemented outside of the function (setting the weights to zero on fully flagged antennas) so I agree that the switch is not strictly necessary. 


> 
> Another way to say this, is that you are making an assumption about the form of `wgts` that is not necessarily true, i.e., that `wgts = n * ~f`
> 
> I think an alternative if you must have this fully automated is to make an experimental wrapper `_aew_red_average` function. Lmk what you think.

hera_cal/apply_cal.py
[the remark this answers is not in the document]
nkern: yeah it seemed like an RTP consideration. I think the best thing to do is, instead of adding a feature to apply_cal that should be in red_average, is to add it to red_average, and then just call red_average from within the apply_cal drive script. Does that make sense?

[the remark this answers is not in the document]
anguta: @nkern, it does seem that we would need a wrapper for `red_average` to do all the i/o.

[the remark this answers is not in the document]
anguta: Partial `i/o` is a generally useful feature for `red_average` to have so I'm open to this. Thoughts @jsdillon ?

hera_cal/utils.py
```diff
@@ -1170,7 +1175,18 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
     Notes:
         1. Different polarizations are assumed to be non-redundant.
         2. Default weighting is nsamples * integration_time * ~flags.
-        3. If wgts Container is fed then they supercede flag and nsample weighting.
+        3. If wgts Container is fed then the behavior of flags and nsamples in the redundantly averaged sum will depend on whether
```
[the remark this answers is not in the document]
anguta: I can remove this.

hera_cal/apply_cal.py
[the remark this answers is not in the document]
anguta: On the other hand, moving the `i/o` into `red_average` (or some wrapper) is a lot of work for a use case that hasn't yet materialized. We have something that works for our use case. It might not be the most optimal way to organize the code but in the interest of getting other stuff working, I personally would like to move on. I'm happy to write an issue for this.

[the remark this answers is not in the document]
nkern: Yeah I'm fine with that for now.

anguta:
Thank you for the detailed and insightful comments @nkern (and sorry if some of my initial pre-edited responses were a bit long-winded). I think this is ready for another pass through!

hera_cal/utils.py
```diff
@@ -1171,7 +1175,18 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
+            propagate_flags is True or False.
+            If False (DEFAULT BEHAVIOR):
+                the final flags are True in the channels/integrations of each redundantly averaged baseline
+                where the the sum of the weights is close to zero and False elsewhere.
+                the final nsamples are equal to the sum of nsamples over each redundant group where the weights are not
+                close to zero.
+            If True
+                the final flags are True in the channels / integrations of each redundantly averaged baseline
+                where the sum of the weights times the logical inverse of the flags times the logical inverse of nsamples
```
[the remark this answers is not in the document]
anguta: this has been removed.

hera_cal/apply_cal.py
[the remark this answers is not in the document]
anguta: Cool!

hera_cal/utils.py
```diff
@@ -1244,14 +1248,18 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
             # take the weighted average
             wsum = np.sum(w, axis=0).clip(1e-10, np.inf)  # this is the normalization
             davg = np.sum(d * w, axis=0) / wsum  # weighted average
-            navg = np.sum(n * f, axis=0)         # this is the new total nsample (without flagged elements)
             fmax = np.max(f, axis=2)             # collapse along freq: marks any fully flagged integrations
             iavg = np.sum(tint.squeeze() * fmax, axis=0) / np.sum(fmax, axis=0).clip(1e-10, np.inf)
-            favg = np.isclose(wsum, 0.0)         # this is getting any fully flagged pixels
+            binary_wgts = (~np.isclose(w, 0)).astype(np.float)  # binary weights.
+            if propagate_flags:
+                fsum = np.sum(n * binary_wgts * f, axis=0)
```
nkern, line 1258: I don't think `fsum` here should have n or binary weights, right? Shouldn't this whole block just be
```
binary_wgts = (~np.isclose(w, 0)).astype(np.float)  # binary weights.
navg = np.sum(n * binary_wgts, axis=0)
if propagate_flags:
    favg = np.all(np.isclose(w * f, 0), axis=0)
else:
    favg = np.all(np.isclose(w, 0), axis=0)
```

I could be wrong though, lmk

anguta: I agree that this makes sense and is more consistent with the operating philosophy of `red_average`. If people want `nsamples=0` to be flagged ,they can add those flags before calling the function.

PR #639, opened 2020-09-09
Apply cal match freqs

anguta:
Allow for applying calibration files to data with intersecting but not identical frequency channels.

### DIFF
--- a/hera_cal/apply_cal.py
+++ b/hera_cal/apply_cal.py
@@ -383,7 +383,8 @@ def apply_cal_argparser():
     a.add_argument("--flag_nchan_high", type=int, default=0, help="integer number of channels at the high frequency end of the band to always flag (default 0)")
     a.add_argument("--filetype_in", type=str, default='uvh5', help='filetype of input data files')
     a.add_argument("--filetype_out", type=str, default='uvh5', help='filetype of output data files')
-    a.add_argument("--nbl_per_load", type=int, default=None, help="Maximum number of baselines to load at once. uvh5 to uvh5 only. Default loads the whole file.")
+    a.add_argument("--nbl_per_load", type=int, default=None, help="Maximum number of baselines to load at once. uvh5 to uvh5 only."
+                                                                  "Default loads the whole file. If 0 is provided, also loads whole file.")
     a.add_argument("--gain_convention", type=str, default='divide',
                    help="'divide' means V_obs = gi gj* V_true, 'multiply' means V_true = gi gj* V_obs.")
     a.add_argument("--redundant_solution", default=False, action="store_true",
--- a/hera_cal/apply_cal.py
+++ b/hera_cal/apply_cal.py
@@ -390,6 +390,5 @@ def apply_cal_argparser():
     a.add_argument("--redundant_solution", default=False, action="store_true",
                    help="If True, average gain ratios in redundant groups to recalibrate e.g. redcal solutions.")
     a.add_argument("--clobber", default=False, action="store_true", help='overwrites existing file at outfile')
-    a.add_argument("--vis_units", default=None, type=str, help="String to insert into vis_units attribute of output visibility file.")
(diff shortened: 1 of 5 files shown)


hera_cal/apply_cal.py
```diff
@@ -246,11 +246,28 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
     # load old calibration solution
     if old_calibration is not None:
         old_hc = io.HERACal(old_calibration)
-        old_gains, old_flags, _, _ = old_hc.read()
+        old_hc.read()
+        # determine frequencies to load in old_hc that are close to hc
+        freqs_to_load = []
+        for f in old_hc.freqs:
+            atol = np.mean(np.diff(hc.freqs)) / 10.
```
jsdillon, line 253: It's not good to have magic numbers like this in the code. This tolerance factor, 10, should be a kwarg, though I think 10 is a reasonable default.

hera_cal/io.py
```diff
@@ -641,7 +641,7 @@ def partial_write(self, output_path, data=None, flags=None, nsamples=None,
                                   this.nsample_array, **self.last_read_kwargs)
 
     def iterate_over_bls(self, Nbls=1, bls=None, chunk_by_redundant_group=False, reds=None,
-                         bl_error_tol=1.0, include_autos=True):
+                         bl_error_tol=1.0, include_autos=True, freqs_to_load=None):
```
jsdillon, line 644: To better match other iterators, I'd call this `freqs` or `frequencies` rather than `freqs_to_load`

hera_cal/io.py
```diff
@@ -713,7 +716,10 @@ def iterate_over_bls(self, Nbls=1, bls=None, chunk_by_redundant_group=False, red
             # make sure that every baseline is in reds
             baseline_chunks = chunk_baselines_by_redundant_groups(reds=reds, max_chunk_size=Nbls)
         for chunk in baseline_chunks:
-            yield self.read(bls=chunk)
+            if freqs_to_load is None:
```
jsdillon, line 719: Do you actually need this if / else? If freqs is None, won't self.read behave appropriately?

scripts/apply_cal.py
```diff
@@ -17,6 +17,9 @@
 if args.vis_units is not None:
     kwargs['vis_units'] = args.vis_units
 
+if args.nbl_per_load == 0:
```
jsdillon, line 20: Why is this necessary? When would this ever happen? Shouldn't this be an error?

jsdillon (changes requested):
Looks pretty good. Just a few small comments/questions.

hera_cal/apply_cal.py
anguta: i've added this to the kwargs and set default to 10.

hera_cal/io.py
anguta: you're right, if freqs i None, things will work fine.

hera_cal/io.py
anguta: calling it frequencies.

scripts/apply_cal.py
anguta: this is to allow for some flexibility in the makeflow bash interface so that we can use the same scripts for when we use partial i/o vs no partial i/o. 

0 is confusing though. I've changed this to "none".

hera_cal/io.py
```diff
@@ -667,7 +667,7 @@ def iterate_over_bls(self, Nbls=1, bls=None, chunk_by_redundant_group=False, red
             include_autos: bool, optional
                 include autocorrelations in iteration if True.
                 Default is True.
-            freqs_to_load: array-like, optional
+            freqsuencies: array-like, optional
```
jsdillon, line 670: typo

jsdillon (approved):
Just one typo, then good to go.

hera_cal/io.py
anguta: fixed.

PR #643, opened 2020-09-20
Fix bug in chunk_baselines_by_redundant_groups where baseline_chunks are not necessarily initialized 

anguta:
If the first chunk is less than max_chunk_size, then baseline_chunks are not initialized. This PR fixes this bug.


jsdillon (approved):

jsdillon:
Assuming tests pass, I approve.

PR #645, opened 2020-09-22
Use sets to speed up membership checking in `joint_reds` in `abscal.match_baselines`

jsdillon:
In my tests on some H3C data, this speed up the match from several minutes to a few seconds.


anguta (approved):
looks good to me.

PR #646, opened 2020-09-23
Use sets to speed up slow list comp

jsdillon:
The line `matched_reds = [red[0] for red in all_reds if bl in red]` was taking about 3% of the total redcal runtime, which is silly. This should be much faster.


plaplant (approved):
Thanks @jsdillon!

PR #647, opened 2020-09-29
Fix abscal bias with new linearized abscal overall amplitude calibration

### LINKED ISSUE #642: Abscal amplitude calibration is biased, especially at low SNR
In the process of Step 4 validation for the H1C IDR2.2 power spectrum paper, we discovered in absolute amplitude calibration of a few percent, though it varies in frequency and LST.

As we now understand it, this bias occurs when performing absolute amplitude calibration (post redcal, so just a single real number per frequency, pol, and time) with a reference model that is less noisy (or even noise free) than the data being calibrated. When SNR on visibilities is low, thermal noise is more likely to increase the amplitude of a measurement than decrease it, leading to a bias. This was previously known in the literature: https://cas.tudelft.nl/pubs/gaincal.pdf

This was discovered in the context of our procedure for producing the abscal reference model in H1C, which involves low-pass delay-filtering externally calibrated at the horizon (plus a 50 ns buffer), and comparing that to the data. This was done to intentionally reduce noise in the reference model and to allow for interpolation over RFI gaps.

In Step 4 of Validation (the end-to-end simulation) we can compare the derived abscal gains to the true gains. We see a bias both using the low-pass filtered abscal model (which has thermal noise, though it is mostly filtered out, but no calibration errors) and using the "true" visibilities as our abscal reference model (no noise or calibration errors). The result in both cases is the same.

image

image

This effect appears pretty insensitive to how the low pass filter is performed, further confirming that it's not a consequence of the low pass filter throwing out real foreground power, as we previously suspected.

image

Following today's Validation telecon, our plan is to quantify this effect for H1C and correct for it post-hoc, which is easier in H1C because smooth_cal basically blurs out the temporal dependence of the bias. However, I think we should try to fix it for for H3C/H4C, since we expect the same effect for abscal against a simulation, e.g. the RIMEz model we're using in H3C/H4C.

jsdillon:
In #642, I explained how @nkern and I discovered a relatively small bias (though larger at low SNR) in absolute amplitude calibration in the process of performing validation for H1C IDR2. This effect arose because we were fitting for the absolute amplitude degeneracy of redundant calibration using the `abs_amp_logcal` solver that solves for amplitudes using a logarithmic linearization `ln|V_ij,xy^data / V_ij,xy^model| = eta_x + eta_y`. (Here x/y are variables standing in for pols and i/j are antenna indices.) This linearization is biased because Gaussian noise with equal variance in its real and imaginary components is always more likely to increase the amplitude of a complex number than decrease it. This effect is most pronounced at low SNR. 

This PR introduces a new solver, `abs_amp_lincal` that instead solves `V_ij,xy^data = A_x A_y * V_ij,xy^model`. When no cross-polarized visibilities are used, this is a simple linear optimization problem in `A^2`. When cross-polarized visibilities are used, this becomes non-linear, but using `abs_amp_logcal` as a starting point, one can still solve for A_x and A_y using `linsolve.LinProductSolver`.

This new function has to be run after phase calibration, since phase errors will cause decoherence that will itself lead to a bais. Here are my results with a the first 5 integrations of a single validation data file (`zen.2458098.40141.foregrounds.true.uvh5`) with manually added gains and noise (but no other systematics), following the procedure in https://github.com/HERA-Team/hera-validation/blob/master/test-series/2/test-2.0.0.ipynb :

image

This PR closes #642.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -212,7 +212,7 @@ def abs_amp_lincal(model, data, wgts=None, verbose=True, return_gains=False, gai
         else:
             data_here[k] = data[k]
         if np.any(~np.isfinite(model[k])):
-            model_here[k] =copy.deepcopy(model[k])
+            model_here[k] = copy.deepcopy(model[k])
             fill_dict_nans(model_here[k], wgts=wgts[k], nan_fill=0.0, inf_fill=0.0, array=True)
         else:
             model_here[k] = model[k]
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -168,10 +168,10 @@ def abs_amp_lincal(model, data, wgts=None, verbose=True, return_gains=False, gai
                 Converegence is measured L2-norm of the change in the solution of the 
                 variables divided by the L2-norm of the solution itself.
                 Default: None (resolves to machine precision for inferred dtype).
-                N.B. Only used when data and model include cross-polarized visibilities.
+                Note: only used when data and model include cross-polarized visibilities.
     
     maxiter : Integer maximum number of iterations to perform LinProductSolver.
-              N.B. Only used when data and model include cross-polarized visibilities.    
+              Note: only used when data and model include cross-polarized visibilities.
 
     verbose : print output, type=boolean, [default=False]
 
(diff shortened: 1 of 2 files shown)


hera_cal/abscal.py
```diff
@@ -129,6 +129,137 @@ def abs_amp_logcal(model, data, wgts=None, verbose=True, return_gains=False, gai
...
+
+    return_gains : boolean. If True, convert result into a dictionary of gain waterfalls.
+
+    gain_ants : list of ant-pol tuples for return_gains dictionary
+    
+    conv_crit : A convergence criterion below which to stop iterating LinProductSolver. 
+                Converegence is measured L2-norm of the change in the solution of the 
+                variables divided by the L2-norm of the solution itself.
+                Default: None (resolves to machine precision for inferred dtype).
+                N.B. Only used when data and model include cross-polarized visibilities.
```
nkern, line 171: had to look up what N.B. meant--I think just `Note:` is a little more commonly used in our codebase

nkern (approved):
this looks good to me! thanks for addressing this josh!

hera_cal/abscal.py
jsdillon: ok, changed it

PR #650, opened 2020-10-02
Have HERAData try reordering blts when they are not equally spaced before throwing an error

jsdillon:
It turns out that using pyuvdata's fast_concat does not ensure equally spaced blts as assumed by HERAData. This PR will fix that problem.

This is preventing https://github.com/HERA-Team/hera_qm/pull/376 from working on site.


anguta (approved):
Seems alright. I approve pending tests passing.

PR #654, opened 2020-10-16
Allow abscal in arbitrary dimensions to account for multiple tip/tilt degeneracies

### LINKED ISSUE #592: Generalize Abscal's tip/tilt solvers to include higher dimensions
As discussed in #590, an alternative solution to removing antennas so that the array only has 2 tip/tilt DoF is to enable abscal to calibrate more DoF by generalizing the tip/tilt calculation to N dimensions. This shouldn't be too terribly difficult, though it's a bit more challenging if we also want to maintain the AbsCal class. So maybe this should be addressed in concert with #577.

jsdillon:
This PR generalizes the post-redcal phase solvers, namely `TT_phs_logcal`, `delay_slope_lincal`, and `global_phase_slope_logcal` to take antenna positions in more than 2 dimensions and thus to solve for phase/delay slopes in more than 2 dimensions. This is useful for arrays with extra degeneracies, since converting antenna positions to "idealized/generalized positions" with `redcal.reds_to_antpos` will make positions with higher numbers of dimensions if there are extra redcal degeneracies. 

The new solvers default to 2D and thus return keys that look like `Phi_ew_Jee`, but by turning off `assume_2D` they return keys like `Phi_0_Jee`, etc. 

The most challenging part of this was generalizing `dft_phase_slope_solver` to N-dimensions. (See #453 for background.) That relied on a brute force whose complexity scaled to the power of the number of dimensions in the search. 2 was OK, but 3 was already pretty unbearably slow and we want to be able to do 4 as well if necessary. So I decided to write an entirely new version of this function that uses gridded FFTs rather than DFTs. The basic idea is that you grid the complex visibility (ratios) by baseline vector, FFT, and look for a peak which should correspond to a phase slope. I just use the peak voxel rather than some sort of fancy quadratic/Quinn's method, since I didn't want to bother generalizing those to N dimensions (I tried doing Quinn's along each dimension individually, but it didn't improve the accuracy appreciably). Anyway, since this solver is just to get out the phase wraps so `TT_phs_logcal` can do its job, I think we're OK without more accuracy. 

Additionally, I found a bottleneck in some of my unit testing of global_phase_slope_logcal which I traced back to repeated re-counting of baselines in data containers caused by `utils.red_average`. I've implemented a fast multi-item `del` in DataContainer to fix this. Perhaps you can think of other places in hera_cal where this would be useful, @nick?

This PR closes #592.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -945,6 +945,74 @@ def dft_phase_slope_solver(xs, ys, data, flags=None):
     return slope_x.reshape(data.shape[1:]), slope_y.reshape(data.shape[1:])
 
 
+def ndim_fft_phase_slope_solver(data, bl_vecs, zero_pad=2, bl_error_tol=1.0):
+    '''Find phase slopes across the array in the data. Similar to utils.fft_dly,
+    but can grid arbitarary bl_vecs in N dimensions (for example, when using
+    generealized antenna positions from redcal.reds_to_antpos in arrays with 
+    extra degeneracies).
+    
+    Parameters:
+    -----------
+    data : dictionary or DataContainer mapping keys to (complex) ndarrays
+    
+    bl_vecs : dictionary mapping keys in data to vectors in N dimensions
+    
+    zero_pad : float factor by which to expand the grid onto which the data is binned. 
+               Increases resolution in Fourier space at the cost of runtime/memory.
+               Must be >= 1.
+    
+    bl_error_tol : float used to define non-zero elements of baseline vectors.
+                   This helps set the fundamental resolution of the grid.
+                   
+    Output:
+    -------
+    phase_slopes : list of length N dimensions. Each element is the same shape
+                   as each entry in data. Contains the phase gradients in units
+                   of 1 / [bl_vecs]. 
+    '''
+    nDim = _count_nDims(bl_vecs, assume_2D=assume_2D)
+    if assume_2D:
+        nDim = 2
+    keys = sorted(list(bl_vecs.keys()))
+
+    # Figure out a grid for baselines and 
+    coords = []
+    all_bins = []
+    bl_vecs_array = np.array([bl_vecs[k] for k in keys])
+    assert zero_pad >= 1, f'zero_pad={zero_pad}, but it must be greater than or equal to 1.'
+    for d in range(nDim):
+        min_comp = np.min(bl_vecs_array[:, d])
+        max_comp = np.max(bl_vecs_array[:, d])
+        # pick minimum delta in this dimension inconsistent with 0 using bl_error_tol
+        dbl = np.min(np.abs(bl_vecs_array[:, d])[np.abs(bl_vecs_array[:, d]) >= bl_error_tol])
+        comp_range = max_comp - min_comp
+        bins = np.arange(min_comp - dbl - comp_range * (zero_pad - 1) / 2, 
+                         max_comp + 2 * dbl + comp_range * (zero_pad - 1) / 2, dbl)
+        all_bins.append(bins)
+        coords.append(np.digitize(bl_vecs_array[:, d], bins))
+    coords = np.array(coords).T
+
+    # create and fill grid with complex data
+    digitized = np.zeros(tuple([len(b) for b in all_bins]) + data[keys[0]].shape, dtype=complex)
+    for i, k in enumerate(keys):
+        digitized[tuple(coords[i])] = data[k]
+
+    # FFT along first nDim dimensions 
+    digitized_fft = np.fft.fftn(digitized, axes=tuple(range(nDim)))
+    # Condense the FFTed dimensions and find the max along them 
+    new_shape = (np.prod(digitized_fft.shape[0:nDim]),) + data[keys[0]].shape
+    arg_maxes = digitized_fft.reshape(new_shape).argmax(0)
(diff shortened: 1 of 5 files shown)


hera_cal/abscal.py
```diff
@@ -298,9 +303,14 @@ def TT_phs_logcal(model, data, antpos, wgts=None, refant=None, verbose=True, zer
...
     antpos : antenna position vectors, type=dictionary
-          keys are antenna integers, values are 2D
-          antenna vectors in meters (preferably centered at center of array),
-          with [0] index containing east-west separation and [1] index north-south separation
+          keys are antenna integers, values are antenna positions vectors
+          (preferably centered at center of array). If assume_2D is True, it is assumed that the 
+          [0] index contains the east-west separation and [1] index the north-south separation
+          
+    assume_2D : type=boolean, [default=False]
+                If this is true, all dimensions of antpos beyond the first two will be ignored.
```
nkern, line 311: indentation

hera_cal/abscal.py
```diff
@@ -298,9 +303,14 @@ def TT_phs_logcal(model, data, antpos, wgts=None, refant=None, verbose=True, zer
...
+                If return_gains is False and assume_2D is True, then the returned variables will
```
nkern, line 312: shouldn't this be `assume_2D = False`?

hera_cal/abscal.py
```diff
@@ -389,11 +398,28 @@ def TT_phs_logcal(model, data, antpos, wgts=None, refant=None, verbose=True, zer
     echo("...finished linsolve", verbose=verbose)
 
     if not return_gains:
+        # rename variables ew/ns instead of 0/1 to maintain backwards compatability
+        if assume_2D:
+            params = list(fit.keys())
+            for p in params:
+                if 'Phi_0' in p:
```
nkern, line 405: how do you know that Phi_0 is ew and not ns?

hera_cal/abscal.py
jsdillon: fixed.

hera_cal/abscal.py
jsdillon: I don't understand... this docstring doesn't have an obvious pattern for indentation

hera_cal/abscal.py
jsdillon: Because of this sentence in the docstring: "If assume_2D is True, it is assumed that the [0] index contains the east-west separation and [1] index the north-south separation."

hera_cal/abscal.py
nkern: you've got three indents here, all the others kwarg descriptions are two indents

jsdillon: They are actually all over the place. model is 8 spaces, data is 7 spaces, wgts is 7, refant is 4, antpos is 6, assume_2D is 12, etc.

jsdillon: I'll make them all line up with the first word after the :, which I think is the closest thing we have to a standard.

jsdillon:
I've got some preliminary pipeline results. Here's before: 

https://nbviewer.jupyter.org/github/HERA-Team/H4C_Notebooks/blob/master/stage_2_redcal/stage_2_redcal_2459122.ipynb
https://nbviewer.jupyter.org/github/HERA-Team/H4C_Notebooks/blob/master/stage_2_abscal/stage_2_abscal_2459122.ipynb

Here's after

https://nbviewer.jupyter.org/github/HERA-Team/H4C_Notebooks/blob/ndim-abscal/stage_2_redcal/stage_2_redcal_2459122.ipynb
https://nbviewer.jupyter.org/github/HERA-Team/H4C_Notebooks/blob/ndim-abscal/stage_2_abscal/stage_2_abscal_2459122.ipynb

I think abscal chi^2 went up during the fornax transit, but otherwise not crazy. 

@nkern Perhaps you can check the images? Unfortunately, your imaging script is still broken. Check out the makeflow in `/lustre/aoc/projects/hera/H4C/makeflow/2459122_ndim_abscal_test`

jsdillon:
In addition to the movies you've made, @nkern (https://eoranalysis.slack.com/archives/C3ZPGMG3E/p1605030685227300), I've verified that intersector baselines appear to be properly in phase after abscal. Here's the array on 2459122:

image

Here's three intersector baselines:

image
image
image

If there was a unsolved-for phase slope between sectors, that should appear in the above as phase offsets between the baselines. For comparison, here's a set of redundant intra-sector baselines (if there were phase errors here, they would cancel out):

image

I'm pretty convinced this is working.

nkern (approved):
@jsdillon I think this looks good. It seems that in adding new tests for the ndim_abscal you re-wrote some of the old tests (e.g. test_TT_phs_logcal)? I can't really tell. As long as our old unit-tests with the old convention pass, then I'm fine w/ this PR being merged.

jsdillon:
@nkern Thanks! 

For the record, I have not removed any unit tests for existing functions like `TT_phs_logcal`, but instead I made some new tests that have two features: 

1) They don't rely on the abscal class (so if we do deprecate it, this will make it easier to maintain coverage) the test the basic functions themselves that operate on data containers which are the ones used in the current pipeline.

2) They don't just check that the answers have the right dimension, type, etc. but I've added a bunch more testing of (very simply) simulated data where the abscal solution is known a priori as opposed to tests that mostly use test data files.

PR #658, opened 2020-10-22
Flag NaNs in abscal solutions

jsdillon:
With the recent abscal updates in #647, it seems that I've introduced the ability for abscal to insert NaNs in the the gain solutions. This happened with a single frequency, time, and polarization (but all ants) when running 2459122, which then caused smooth_cal to choke, spinning its wheels endlessly. I have verified that this fix makes smooth_cal converge in a reasonable amount of time.


anguta (approved):
looks good!

PR #659, opened 2020-10-23
Move abscal nan flagging to end

jsdillon:
While #658 didn't do what I was hoping, it appears that this does the trick.


anguta (approved):
I approve, pending tests passing.

PR #629, opened 2020-08-18
flag factorization before filtering runs.

anguta:
This pull request allows for flag factorization to be performed as part of delay and xtalk filtering. Also add baseline parallelization support for delay-filtering.

Closes #630 #632 Depends on #619 and #625

PR #668, opened 2020-11-24
Enable select on read for HERACal

jsdillon:
This is essentially the same as UVCal.read_calfits() and then UVCal.select() when loading a single file, but it saves memory if you want to load a list of calibration files, downselecting them as you go.

This came up as a useful feature in a review of #629, specifically this comment: https://github.com/HERA-Team/hera_cal/pull/629#discussion_r529113146


anguta (approved):
Looks good to me.

PR #674, opened 2020-12-16
Flagging Features: Options to skip integrations or channels based on flagging patterns that might cause trouble for linear filtering. Also add apply_flags functionality to vis_clean.

anguta:
A pull request merging in several major features in flagging that help with the stability of the linear filter (and maybe CLEAN too?) and also support external flags and yaml flags for vis_clean methods. 

New flagging features in `fourier_filter` and higher level functions include:

* `skip_flagged_edge_freqs` : If an edge channel is flagged for all integrations, then perform interpolation within the channels that are not flagged for all integration and leave the entirely flagged edge channel permanently flagged. 
*  `skip_flagged_edge_times` : The same as `skip_flagged_edge_freqs` but now applied to the time dimension.
*  `skip_gaps_larger_then_filter_period` : skip integrations or channels where the smallest wavelength interpolation fourier mode is smaller then the largest contiguous flagging gap.
*  `skip_if_flag_within_edge_distance` : skip integration or channel if a flag is within a user provided distance of the edge.
* `flag_model_rms_outliers` : skip integrations or channels where the rms of the interpolation model exceeds the rms of the unflagged data by some user provided parameter.


hera_cal/delay_filter.py
```diff
@@ -178,15 +198,21 @@ def load_delay_filter_and_write_baseline_list(datafile_list, baseline_list, calf
...
             Fully flagged integrations do not count towards triggering time_thresh.
+        external_flags : str, optional, path to external flag files to apply
         res_outfilename: path for writing the filtered visibilities with flags
         CLEAN_outfilename: path for writing the CLEAN model visibilities (with the same flags)
         filled_outfilename: path for writing the original data but with flags unflagged and replaced
             with CLEAN models wherever possible
         clobber: if True, overwrites existing file at the outfilename
         add_to_history: string appended to the history of the output file
+        polarizations: list of polarizations to include and write.
+        verboase: lots of output.
```
jsdillon, line 209: This isn't an option

jsdillon (changes requested):
Sorry this took so long. Mostly minor changes. 

One thing I don't like is the vast proliferation of scripts. If we have multiple ways of doing roughly the same thing, I feel like it should be 1 script with command line options. Right now, the large number of scripts for dayneu, clean, dpss, etc. for the same kind of filtering makes the codebase harder to maintain and introduces lots of extra un-unit-tested code.

jsdillon:
@nkern might also want to take a look, since this is more closely related to your work on visclean

hera_cal/io.py
```diff
@@ -1806,28 +1806,27 @@ def baselines_from_filelist_position(filename, filelist, polarizations=None):
         name of the file being processed.
     filelist : list of strings
         name of all files over which computations are being parallelized.
-    polarizations : list of strings
-        polarizations to include in baseline parallelization.
-
+    chunk_pols : bool, optional
```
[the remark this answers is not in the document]
anguta: got rid of it.

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,64 @@ def apply_calibration(self, input_cal, unapply=False):
...
                                      gain_convention=gain_convention)
 
+    def apply_flags(self, external_flags=None, overwrite_data_flags=False,
+                    flag_zero_times=True, a_priori_flag_yaml=None):
+        """
+        apply external flags.
+        Parameters
+        ----------
+        external_flags: str, optional.
+            Str or list of strings pointing to flag files to apply.
```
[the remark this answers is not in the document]
anguta: I've mentioned both of these things in the docstring.

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,64 @@ def apply_calibration(self, input_cal, unapply=False):
...
+        overwrite_data_flags: bool, optional
+            If true, overwrite all data flags for bls that are not entirely flagge.d
```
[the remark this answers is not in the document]
anguta: fixed.

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,64 @@ def apply_calibration(self, input_cal, unapply=False):
...
+        flag_zero_times: bool, optional
```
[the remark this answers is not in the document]
anguta: I've added an explanation in the docstring.

scripts/dpss_delay_filter_run.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
anguta: That is one option. I am a bit concerned with condensing the scripts from a usability perspective since I think different sets of parameters are best for different filtering modes. For example, linear filters tend to have issues with flagged edges so they work best when skip_flagged_edges is True. I'm worried that non-experts won't know to set these defaults and run into problems.

hera_cal/xtalk_filter.py
```diff
@@ -30,8 +30,9 @@ class XTalkFilter(VisClean):
     """
 
     def run_xtalk_filter(self, to_filter=None, weight_dict=None, max_frate_coeffs=[0.024, -0.229], mode='clean',
-                         skip_wgt=0.1, tol=1e-9, verbose=False, cache_dir=None, read_cache=False,
-                         write_cache=False, **filter_kwargs):
+                         skip_wgt=0.1, tol=1e-9, cache_dir=None, read_cache=False,
+                         write_cache=False, skip_flagged_edges=False, flag_filled=True,
```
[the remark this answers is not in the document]
anguta: added docstring.

hera_cal/xtalk_filter.py
```diff
@@ -134,31 +146,38 @@ def load_xtalk_filter_and_write(infilename, calfile=None, Nbls_per_load=None, sp
...
             if factorize_flags:
                 xf.factorize_flags(time_thresh=time_thresh, inplace=True)
-            xf.run_xtalk_filter(cache_dir=cache_dir, read_cache=read_cache, write_cache=write_cache, **filter_kwargs)
+            xf.run_xtalk_filter(cache_dir=cache_dir, read_cache=read_cache, write_cache=write_cache,
+                                skip_flagged_edges=skip_flagged_edges, **filter_kwargs)
             xf.write_filtered_data(res_outfilename=res_outfilename, CLEAN_outfilename=CLEAN_outfilename,
                                    filled_outfilename=filled_outfilename, partial_write=True,
                                    clobber=clobber, add_to_history=add_to_history,
-                                   freq_array=xf.hd.freq_array, Nfreqs=xf.Nfreqs)
+                                   freq_array=np.asarray([xf.freqs]), Nfreqs=xf.Nfreqs)
```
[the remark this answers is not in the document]
anguta: I don't think this one was intentional. I've reverted it.

hera_cal/xtalk_filter.py
```diff
@@ -134,31 +146,38 @@ def load_xtalk_filter_and_write(infilename, calfile=None, Nbls_per_load=None, sp
...
             xf.hd.data_array = None  # this forces a reload in the next loop
 
 
 def load_xtalk_filter_and_write_baseline_list(datafile_list, baseline_list, calfile_list=None, spw_range=None, cache_dir=None,
-                                              read_cache=False, write_cache=False,
+                                              read_cache=False, write_cache=False, external_flags=None,
                                               factorize_flags=False, time_thresh=0.05,
                                               res_outfilename=None, CLEAN_outfilename=None, filled_outfilename=None,
-                                              clobber=False, add_to_history='', round_up_bllens=False, **filter_kwargs):
+                                              clobber=False, add_to_history='', round_up_bllens=False, polarizations=None,
```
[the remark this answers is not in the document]
anguta: added this.

hera_cal/delay_filter.py
anguta: I've removed it.

hera_cal/io.py
```diff
@@ -1806,28 +1806,27 @@ def baselines_from_filelist_position(filename, filelist, polarizations=None):
...
+    bls = list(set([bl[:2] for bl in hd.bls]))
     file_index = filelist.index(filename)
     nfiles = len(filelist)
     # Determine chunk size
     nbls = len(bls)
     chunk_size = nbls // nfiles + 1
     lower_index = file_index * chunk_size
     upper_index = np.min([(file_index + 1) * chunk_size, nbls])
-    return bls[lower_index:upper_index]
+    # only return baselines if lower and upper indices are within number of bls
```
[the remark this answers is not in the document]
anguta: I don't think the else should get triggered. I've removed it.

scripts/dpss_delay_filter_run.py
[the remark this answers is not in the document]
anguta: I've decided to go with a system with a single script that sets different defaults based on the filtering mode.

nkern:
@anguta in the PR description could you just make a bullet point list of major changes (i.e. can you elaborate a little on the "several major flagging features" that you add? Just to make it easier to come back to this and know what happened.

hera_cal/delay_filter.py
```diff
@@ -55,13 +56,12 @@ def run_delay_filter(self, to_filter=None, weight_dict=None, horizon=1., standof
...
                 Only works properly when all weights are all between 0 and 1.
             tol : float, optional. To what level are foregrounds subtracted.
-            verbose: If True print feedback to stdout
             cache_dir: string, optional, path to cache file that contains pre-computed dayenu matrices.
                         see uvtools.dspec.dayenu_filter for key formats.
             read_cache: bool, If true, read existing cache files in cache_dir before running.
             write_cache: bool. If true, create new cache file with precomputed matrices
                                that were not in previously loaded cache files.
-            cache: dictionary containing pre-computed filter products.
+            skip_flagged_edges : bool, if true do not include edge freqs in filtering region (filter over sub-region).
```
nkern, line 64: is this only a bool? how do you specify the edge freqs if this is true? if this is mixed type, then please specify

hera_cal/io.py
```diff
@@ -1812,28 +1812,20 @@ def baselines_from_filelist_position(filename, filelist, polarizations=None):
         name of the file being processed.
     filelist : list of strings
         name of all files over which computations are being parallelized.
-    polarizations : list of strings
```
nkern, line 1815: why remove the polarizations functionality here? don't have a strong opinion just wondering what the motive is

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,70 @@ def apply_calibration(self, input_cal, unapply=False):
         apply_cal.calibrate_in_place(self.data, cal_gains, self.flags, cal_flags,
                                      gain_convention=gain_convention)
 
+    def apply_flags(self, external_flags=None, overwrite_data_flags=False,
+                    flag_zero_times=True, a_priori_flag_yaml=None):
+        """
+        apply external flags.
+        Parameters
```
nkern, line 221: line break between desc and Parameters

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,70 @@ def apply_calibration(self, input_cal, unapply=False):
...
+            If true, overwrite all data flags for bls that are not entirely flagged.
+        flag_zero_times: bool, optional
+            if true, don't overwrite flags where the entire time is flagged.
+            this can be useful if we are applying external flags that do not have
+            thresholding.
+            An example of where we might want to do this is if we are doing a second round of RFI flagging
+            starting with a more conservative set of flags but we want to keep fully flagged times.
+        a_priori_flag_yaml: str, optional
+            path to a yaml file containing manual flags.
+        Returns
```
nkern, line 237: don't need this if it doesn't return anything

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,70 @@ def apply_calibration(self, input_cal, unapply=False):
         apply_cal.calibrate_in_place(self.data, cal_gains, self.flags, cal_flags,
                                      gain_convention=gain_convention)
 
+    def apply_flags(self, external_flags=None, overwrite_data_flags=False,
```
nkern, line 217: I would consider keeping the `apply_` methods consistent with how I have them for `apply_calibration(self, input_cal)`.  Note that there I do not change the data in `self.hd` which is kept in its original form, but only change the DataContainers `self.data` and `self.flags`. Unless you think there is a specific reason why we need to change `self.hd`. For writing data, the `self.write_data` method should handle writing the data and flags in `self.data` and `self.flags`

One easy way to do this is to do something like
```
hd_flagged = copy.deepcopy(self.hd)
uvutils.apply_flags(hd_flagged, uvflags)
d,f,n = hd_flagged.build_datacontainers()
self.flags = self.flags + f
```
or something like this

hera_cal/vis_clean.py
```diff
@@ -307,7 +372,12 @@ def vis_clean(self, keys=None, x=None, data=None, flags=None, wgts=None,
                   ax='freq', horizon=1.0, standoff=0.0, cache=None, mode='clean',
                   min_dly=10.0, max_frate=None, output_prefix='clean',
                   skip_wgt=0.1, verbose=False, tol=1e-9,
-                  overwrite=False, **filter_kwargs):
+                  overwrite=False,
+                  skip_flagged_edge_freqs=False,
```
nkern, line 376: can you explain how these `skip_flagged_` params are different than the edgecut parameters?

hera_cal/vis_clean.py
```diff
@@ -335,9 +405,29 @@ def vis_clean(self, keys=None, x=None, data=None, flags=None, wgts=None,
...
+            if true, do not filter over flagged edge times (filter over sub-region)
+            defualt is False
+        skip_gaps_larger_then_filter_period : bool, optional
+            if true, skip integrations or channels with gaps that are larger then the period of
+            of the finest scale mode used for interpolation.
+            default is False.
+        clean_flags_in_resid_flags : bool, optional
+            if true, include clean flags in residual flags that will be written out in res_outfilename.
+            default is False.
+        flag_filled : bool, optional
```
nkern, line 424: Hmm, can you change this parameter name to something a little more informative, perhaps `keep_flags`
and also the term `filter flags` is a little ambiguous, that makes me thing of the flags used for performing the CLEAN. Perhaps say `set the post-filtered flags equal to ...`

hera_cal/vis_clean.py
```diff
@@ -486,6 +590,24 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
...
+            of the finest scale mode used for interpolation.
+        flag_filled : bool, optional
+            if true, set filter flags equal to the original flags (do not unflag interpolated channels)
+            This is useful for cross-talk filtering where the cross-talk modes will not completely interpolate
+            over the channel gaps since there are substantial contributions to the foreground power from fringe-rates
+            that are not being modeled as cross-talk. In this case, we may want a file with the modelled cross-talk included but
+            not used to in-paint flagged integrations.
+        clean_flags_in_resid_flags : bool, optional
+            if true, include clean flags in residual flags that will be written out in res_outfilename.
+            default is False.
```
nkern, line 610: I know this is not technically part of this PR, but I think it is good to have an enumeration of all of the filter_kwargs somewhere in hera_cal, which you can then reference from higher level wrappers like vis_clean, or delay_clean etc. In other words, we shouldn't reference all the way to `uvtools`, and should probably put all of the kwargs and their docstrings at least once at the lowest level hera_cal method, which I think is here. I guess part of my thought here is, some of the enumerated kwargs are indeed filter_kwargs--what made you decide which to enumerate and which to package up in the **filter_kwargs dict? If we can't think of a good reason why, perhaps its better to enumerate them here. But I do acknowlegde it can get tedious every time, which is why I think its probably fine to do this in higher level wrappers within the hera_cal repo

hera_cal/vis_clean.py
```diff
@@ -398,16 +488,30 @@ def vis_clean(self, keys=None, x=None, data=None, flags=None, wgts=None,
...
 
     def fourier_filter(self, filter_centers, filter_half_widths, mode,
                        x=None, keys=None, data=None, flags=None, wgts=None,
                        output_prefix='clean', zeropad=None, cache=None,
-                       ax='freq', skip_wgt=0.1, verbose=False, overwrite=False, **filter_kwargs):
+                       ax='freq', skip_wgt=0.1, verbose=False, overwrite=False,
+                       skip_flagged_edge_freqs=False, skip_flagged_edge_times=False,
+                       skip_gaps_larger_then_filter_period=False,
+                       flag_filled=False, clean_flags_in_resid_flags=False,
+                       **filter_kwargs):
```
nkern, line 514: should also note in the method description that this is just a wrapper of `uvtools.dpsec.fourier_filter`

hera_cal/vis_clean.py
```diff
@@ -664,10 +786,88 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
                         w, _ = zeropad_array(w, zeropad=zeropad[m], axis=m)
                         xp[m] = np.hstack([x[m].min() - (np.arange(zeropad[m])[::-1] + 1) * np.mean(np.diff(x[m])),
                                            x[m], x[m].max() + (1 + np.arange(zeropad[m])) * np.mean(np.diff(x[m]))])
-            mdl, res, info = dspec.fourier_filter(x=xp, data=d, wgts=w, filter_centers=filter_centers,
+            mdl, res = np.zeros_like(d), np.zeros_like(d)
+            if skip_flagged_edge_freqs:
```
nkern, line 790: FYI: the next ~80 lines of code have zero comments, so its hard to understand what you are doing. can you just add some comments at each high-level block to help comprehension? thanks. Also, this is a serious amount of code for `fourier_filter`, which should just be in effect a shallow wrapper for `dspec.fourier_filter`. is it worth making this its own function, both for modularity, but also for code comprehension here? lmk your thoughts thanks

hera_cal/vis_clean.py
```diff
@@ -664,10 +786,88 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
...
+                            on_flag = (w == 0)
+                            if on_flag:
+                                current_flag_length += 1
+                            else:
+                                if current_flag_length >= max_contiguous:
+                                    max_contiguous = current_flag_length
+                                current_flag_length = 0
+                        max_filter_delay = np.max([np.abs(fc) + np.abs(fw) for fc, fw in zip(filter_centers, filter_half_widths)])
+                        dvoxel = np.mean(np.diff(x))
+                # if width of largest contiguous flag region is greater then the largest filtering delay, then flag the whole integration.
```
nkern, line 839: indentation

nkern:
@anguta after pushing to your branch there are two unit test failures, one is the trim_model test which I'll have a look at, the other is deep in the dpss weeds that I don't fully understand--lmk if you have any insight, thanks!

hera_cal/delay_filter.py
anguta: If this is true, then the method automatically identifies edges that are fully flagged (e.g. a channel where all times are flagged) and does not include them in the filter. I'm updating the docstring to make this more clear.

hera_cal/io.py
anguta: Splitting different polarizations in the same baseline into different files runs into trouble downstream at the stage where power spectra waterfalls for different baselines are combined into single files again. `hera_pspec` does not currently support combining power-spectrum files with the same baseline times but different polarizations so it made sense to avoid splitting polarizations.

hera_cal/vis_clean.py
anguta: removed.

hera_cal/vis_clean.py
anguta: Good idea; I think its good to track the original data state.

hera_cal/vis_clean.py
anguta: I'll explain this in the CLEAN context (iterative modeling with dft modes).

The edgecut parameters set the edge channels to zero which effects the CLEAN kernel, but does not affect the set of fourier modes that you are using to model your signal (using DFT modes spaced by 1/Ndata).  skip_flagged_edges only performs CLEAN on the regions bounded by entirely flagged edge channels which changes the basis that you are using to model your data (now you are modeling your data with DFT modes spaced by (1 / (Ndata - N_edge_skipped) ).  I've found that linear filters with DFT modes and DPSS modes tend to have problems with extrapolating flagged edges and also interpolating flags that are close to an edge when there is not taper that goes to zero. TBH I'm not completely sure why.

hera_cal/vis_clean.py
anguta: I've adopted this language and gotten rid of the xtalk analogy.

hera_cal/vis_clean.py
anguta: added this.

hera_cal/vis_clean.py
anguta: It looks like this is mentioned at the beginning of the docstring (you may have added it)? Let me know if this is good.

hera_cal/vis_clean.py
```diff
@@ -664,10 +780,88 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
...
-            mdl, res, info = dspec.fourier_filter(x=xp, data=d, wgts=w, filter_centers=filter_centers,
+            mdl, res = np.zeros_like(d), np.zeros_like(d)
+            if skip_flagged_edge_freqs:
+                unflagged_chans = np.where(~np.all(np.isclose(w, 0.0), axis=0))[0]
+                if len(unflagged_chans) > 0:
+                    ind_left = np.min(unflagged_chans)
+                    ind_right = np.max(unflagged_chans) + 1
+                else:
+                    ind_left = 0
+                    ind_right = d.shape[1]
```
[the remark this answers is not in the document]
anguta: Thanks for catching this! I've cleaned this up.

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,64 @@ def apply_calibration(self, input_cal, unapply=False):
...
+            self.hd.update(flags=self.flags)
+        # explicitly keep_existing since we already reset flags.
+        if external_flags is not None:
+            flag_apply(external_flags, self.hd, force_pol=True, keep_existing=True)
+        # apply apriori flag yaml too.
+        if a_priori_flag_yaml is not None:
+            import hera_qm.utils as qm_utils
+            self.hd = qm_utils.apply_yaml_flags(self.hd, a_priori_flag_yaml)
+
+        _, self.flags, _ = self.hd.build_datacontainers()
```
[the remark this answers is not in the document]
anguta: I think the `update` statements make sure that the flags are applied to the `HERAData` object. The reason for rebuilding is that `apply_yaml_flags` has to operate on a `HERAData` object so I need to do this on the `hd` and then rebuild.

hera_cal/vis_clean.py
```diff
@@ -664,10 +780,88 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
...
+                xp[0] = xp[0][ind_lower: ind_upper]
+                xp[1] = xp[1][ind_left: ind_right]
+            elif ax == 'time':
+                xp = xp[ind_lower: ind_upper]
+            elif ax == 'freq':
+                xp = xp[ind_left: ind_right]
+
+            # skip half period gaps
+            if skip_gaps_larger_then_filter_period:
+                if ax == 'freq':
```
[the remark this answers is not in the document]
anguta: I think that's right that this should also trigger if `both`. I've cleaned this up.

hera_cal/vis_clean.py
anguta: fixed.

hera_cal/vis_clean.py
anguta: I've added more documentation.

As far as breakout functions go. I'm not sure if there is a lot of utility outside of `fourier_filter` within `hera_cal`. I am open to moving them to a breakout function if you think that would help with readability (can see this going the other way though since it requires someone to scroll around to all these different functions to figure out whats going on). 

The ideal place to have all these features might actually be in `uvtools` since they help a lot with the stability of the linear filters and it would be useful for them to be available as defaults for people that are not using them on HERA.

hera_cal/vis_clean.py
anguta: I think the `filter_kwargs` here are basically set by `uvtools.dspec.fourier_filter` and the choice of which args to make kwargs there was set by options that were different depending on whether the filtering mode chosen (`clean` vs. `dayenu` vs. `dpss_leastsq` etc...). I think these have been enumerated here.

Maybe you are saying that we should put more args into kwargs at higher levels?

nkern:
@anguta looks like there is a specific problem with numpy.testing in Python 3.7, where it is no longer allowing an array to be equated to a float. You should just update the unittest such that instead of 
np.testing.assert_array_equal(array, scalar)
it does
assert np.isclose(array, scalar).all()

hera_cal/vis_clean.py
nkern: yeah its fine, i added it

hera_cal/vis_clean.py
nkern: no, I'm just saying that we might want to enumerate all of the possible kwargs at some point here, but on second thought that's tough b/c it will change as uvtools is developed. as long as its clear that the additional kwargs are specific to additional kwargs in `uvtools.dspec.fourier_filter` I'm fine w/ it. In other words, ppl know where to look if they are wondering what else might go in this open-ended kwarg slot.

anguta: I've added a note that these kwargs are enumerated in `uvtools.dspec.fourier_filter`.

anguta:
> @anguta after pushing to your branch there are two unit test failures, one is the trim_model test which I'll have a look at, the other is deep in the dpss weeds that I don't fully understand--lmk if you have any insight, thanks!

I've fixed the DPSS issue.

anguta:
> @anguta looks like there is a specific problem with numpy.testing in Python 3.7, where it is no longer allowing an array to be equated to a float. You should just update the unittest such that instead of
> np.testing.assert_array_equal(array, scalar)
> it does
> assert np.isclose(array, scalar).all()

It looks like this has been fixed?

nkern:
> > @anguta looks like there is a specific problem with numpy.testing in Python 3.7, where it is no longer allowing an array to be equated to a float. You should just update the unittest such that instead of
> > np.testing.assert_array_equal(array, scalar)
> > it does
> > assert np.isclose(array, scalar).all()
> 
> It looks like this has been fixed?

I don't think so, I still see a Python 3.7 error on the checks

plaplant:
@anguta I think the test failures are related to a linsolve + numpy v1.20 issue that has been fixed upstream. If you rebase off the current version of `master` and rerun the tests, they should pass now.

anguta:
> @anguta I think the test failures are related to a linsolve + numpy v1.20 issue that has been fixed upstream. If you rebase off the current version of `master` and rerun the tests, they should pass now.

Looks like rebasing on master did the trick!

hera_cal/vis_clean.py
```diff
@@ -213,6 +225,69 @@ def apply_calibration(self, input_cal, unapply=False):
...
                                      gain_convention=gain_convention)
 
+    def apply_flags(self, external_flags=None, overwrite_data_flags=False,
+                    flag_zero_times=True, a_priori_flag_yaml=None):
+        """
+        apply external flags.
+
+        Parameters
+        ----------
+        external_flags: str, optional.
```
nkern, line 235: why make this only a string? shouldn't this just be a UVFlag object or a string?

hera_cal/vis_clean.py
```diff
@@ -213,6 +225,69 @@ def apply_calibration(self, input_cal, unapply=False):
...
+        flag_zero_times: bool, optional
+            if true, don't overwrite flags where the entire time is flagged.
+            this can be useful if we are applying external flags that do not have
+            thresholding.
+            An example of where we might want to do this is if we are doing a second round of RFI flagging
+            starting with a more conservative set of flags but we want to keep fully flagged times.
+        a_priori_flag_yaml: str, optional
+            path to a yaml file containing manual flags.
+        """
+        hd_flagged = copy.deepcopy(self.hd)
```
nkern, line 250: this perhaps seems unnecessary memory bloat--can we apply the flags to the DataContainer without copying the underling `self.hd`?

hera_cal/vis_clean.py
```diff
@@ -213,6 +225,69 @@ def apply_calibration(self, input_cal, unapply=False):
...
+        if external_flags is not None:
+            external_flags = UVFlag(external_flags)
+            # select frequencies and times that match data.
+            flag_times = np.unique(external_flags.time_array)
```
nkern, line 254: I think you can replace this with something simpler like
`external_flags.select(times=self.times, frequencies=self.freqs)`
and then check that external_flags is not null

hera_cal/vis_clean.py
```diff
@@ -213,6 +214,64 @@ def apply_calibration(self, input_cal, unapply=False):
...
+                if np.any(np.isclose(self.times, t)):
+                    times_overlapping.append(t)
+            for f in flag_freqs:
+                if np.any(np.isclose(self.freqs, f)):
+                    freqs_overlapping.append(f)
+            # select frequencies and times that overlap with data.
+            external_flags.select(frequencies=freqs_overlapping, times=times_overlapping)
+            # make the times and frequencies in external_flags identical to data
+            # since to_baseline uses exact equality to broadcast flags to vis shape.
+            external_flags.time_array = np.unique(self.times)
```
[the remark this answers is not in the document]
nkern: self.times is already unique, same for self.freqs

hera_cal/vis_clean.py
```diff
@@ -662,10 +752,143 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
                         w, _ = zeropad_array(w, zeropad=zeropad[m], axis=m)
                         xp[m] = np.hstack([x[m].min() - (np.arange(zeropad[m])[::-1] + 1) * np.mean(np.diff(x[m])),
                                            x[m], x[m].max() + (1 + np.arange(zeropad[m])) * np.mean(np.diff(x[m]))])
-            mdl, res, info = dspec.fourier_filter(x=xp, data=d, wgts=w, filter_centers=filter_centers,
+            mdl, res = np.zeros_like(d), np.zeros_like(d)
```
nkern, line 755: Lines 755 -- 884 is a large chunk of code that is only used for the specific case of this edge flagging, and is not relevant to other use-cases in Fourier filter. It also makes this "wrapper" method more of a behemoth and thus harder to maintain. I would ask that you make two separate, module-level functions either here in vis_clean or in utils, e.g.
```
def flag_edge_channels(data_in):
    """this is for skip_flagged_edge_freqs and skip_flagged_edge_times"""

def flag_edge_distance(data_in):
    """ this is for flag_within_edge_distance"""
```

and then all of this code here simply becomes
if skip_flagged_edge_freqs or skip_flagged_edge_times:
   # call flag_edge_channels

if flag_within_edge_distance:
    # call flag_edge_distance

You'll notice that for the default case of CLEAN, currently a lot of this code is actually run but none of it effects the actual data. This will make the separation between the different methods a lot cleaner (pun intended) I think.

Lmk your thoughts

hera_cal/vis_clean.py
```diff
@@ -662,10 +752,143 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
...
+                    if skip_if_flag_within_edge_distance:
+                        if np.any(np.isclose(win[:max_edge, rownum], 0.0)) | np.any(np.isclose(win[-max_edge - 1:, rownum], 0.0)):
+                            win[:, rownum] = 0.
+            # perform the filtering on weights and data that have been truncated to exclude flaggged edges
+            # per users preference.
+            mdl, res, info = dspec.fourier_filter(x=xp, data=din, wgts=win, filter_centers=filter_centers,
                                                   filter_half_widths=filter_half_widths,
                                                   mode=mode, filter_dims=filterdim, skip_wgt=skip_wgt,
                                                   **filter_kwargs)
+            # insert filtered model etc... ba
```
nkern, line 889: typo?

hera_cal/vis_clean.py
```diff
@@ -692,14 +915,39 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
                                 skipped[:, i] = True
                             elif dim == 1:
                                 skipped[i] = True
-
+            # A redundant system (in the airline sense of the word).
```
nkern, line 918: not sure what you mean here

hera_cal/vis_clean.py
```diff
@@ -692,14 +915,39 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode,
...
+            # just in case any artifacts make it through after our best flagging
```
nkern, line 919: I don't see any reason why this should be in the Fourier filter method. For modularity, this should be its own method, not something inherent to this function.

I would advise to make a new module level function that operates directly on the data containers. a drive script can simply call this additional method if it likes, no reason to have the `model_rms_threshold` kwarg in fourier_filter

hera_cal/vis_clean.py
anguta: The copy is here to keep the original flags but I agree that its unecessary to copy the whole `HERAData` object. Instead, I now copy the original flags only. Perform flagging and data container building on `self.hd` and then re-apply the original flags.

hera_cal/vis_clean.py
anguta: I've run into trouble using these methods because `select` requires exact floating point matches between frequencies and times which we usually dont get.

hera_cal/vis_clean.py
anguta: Sounds good. Adding UVFlag support.

hera_cal/vis_clean.py
anguta: I've moved all of this code into three different breakout util functions.

hera_cal/vis_clean.py
anguta: I think so.

hera_cal/vis_clean.py
anguta: Also clarified the Noneness in the docstring and added a ValueError if they are both None.

hera_cal/vis_clean.py
anguta: The comment isn't super helpful so I've removed it. Mean that the RMS flagging deals with artifacts that might make it through the other flags (just like airlines have many redundant copies of systems as backups in case one (or more) fail).

hera_cal/vis_clean.py
anguta: I've broken this out into `vis_clean.flag_model_rms`

anguta:
Thanks for the recommendations @nkern ! after breaking those flagging functions out, i think things are definitely cleaner (and more available for sandbox stuff).

I think this is ready for another look whenever you have the time (pending all the tests passing!)

hera_cal/vis_clean.py
nkern: ah yes, I remember something similar to this, okay

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+    ax : str, optional
+        string specifying which axis to flag edges of.
+        valid options include 'freq', 'time', 'both'.
+        default is 'freq'
+    Returns
+    -------
+    wout, array-like 2d
+        weights with rows or columns with zero weights
+        within min_flag_edge_distance set entirely to zero.
+
```
nkern, line 104: excess line break

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
-import warnings
+
+
+def truncate_flagged_edges(data_in, weights_in, x, ax='freq'):
+    """
+    cut away edge channels and integrations that are completely flagged
+
+    Parameters
+    ----------
+    data_in : array-like, 2d
```
nkern, line 29: also state here the implicit shape, i.e. `2d of shape (Ntimes, Nfreqs)`

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+    return xout, dout, wout, edges
+
+
+def flag_rows_with_flags_within_edge_distance(weights_in, min_flag_edge_distance, ax='freq'):
+    """
+    flag integrations (and/or channels) with flags within min_flag_edge_distance of edge.
+
+    Parameters
+    -----------
+    weights_in : array-like, 2d
```
nkern, line 89: same comment here about shape of input data

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+    return wout
+
+
+def flag_rows_with_contiguous_flags(weights_in, max_contiguous_flag, ax='freq'):
+    """
+    flag any row or column with contiguous zero-weights over limit.
+
+    Parameters
+    ----------
+    weights_in : array-like, 2d
```
nkern, line 127: same here

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+            elif np.any(np.isclose(wout[rownum, :min_flag_edge_distance], 0.0)) | np.any(np.isclose(wout[rownum, -min_flag_edge_distance - 1:], 0.0)):
+                wout[rownum, :] = 0.
+        if ax == 'both':
+            wout = flag_rows_with_flags_within_edge_distance(wout, min_flag_edge_distance[0], ax='time')
+    return wout
+
+
+def flag_rows_with_contiguous_flags(weights_in, max_contiguous_flag, ax='freq'):
+    """
+    flag any row or column with contiguous zero-weights over limit.
```
nkern, line 123: to make this a little clearer, can you add `zero-weights over a specified limit`

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+
+    Parameters
+    ----------
+    weights_in : array-like, 2d
+        weights to check. any row (ax='time') or col (ax='freq')
+        with contiguous regions of zero with length greater then max_contiguous_flag
+        will be set to zero.
+    max_contiguous_flag : integer (or 2-list/tuple)
```
nkern, line 131: `2-tuple for each axis if ax='both'`

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+        any row or column with contiguous weights equal to zero greater or equal to this value
+        will be set entirely to zero.
+    ax : str, optional
+        axis to perform flagging over.
```
nkern, line 135: `axis to perform flagging over. options=['time', 'freq', 'both'], default='freq'`

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+    flag any row or column with contiguous zero-weights over limit.
+
+    Parameters
+    ----------
+    weights_in : array-like, 2d
+        weights to check. any row (ax='time') or col (ax='freq')
+        with contiguous regions of zero with length greater then max_contiguous_flag
+        will be set to zero.
+    max_contiguous_flag : integer (or 2-list/tuple)
+        any row or column with contiguous weights equal to zero greater or equal to this value
```
nkern, line 132: this is a little confusing to me the way its worded, perhaps `flag any row or column when any N series of contiguous bins along the axis are zero, where N = max_contiguous_flags`

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+        x (and y) axes of data to determine maximum contiguous flags from.
+    filter_centers : list or 2-tuple/list of lists
+        centers of filtering-windows.
+    filter_half_widths : list or 2-tuple list of lists
+        half-widths of filtering windows.
+
+    Returns
+    -------
+    max_contiguous_flag: int or 2-list containing the width of a region corresponding
+                         to the largest delay in the filter centers and filter_widths
```
nkern, line 186: indentation: btw I noticed that you tend to indent line breaks to the beginning of the previous line. I think technically indented line breaks should only have 1 tab. perhaps you can rectify other instances of this in this PR? I know its confusing b/c I think we adopted slightly different standards throughout hera_cal development

i.e.
```
max_contiguous_flags: int or 2-list....
    to the largest...
```

hera_cal/vis_clean.py
```diff
@@ -7,29 +7,240 @@
...
+            elif max_contiguous >= max_contiguous_flag:
+                wout[rownum][:] = 0.
+        if ax == 'both':
+            wout = flag_rows_with_contiguous_flags(wout, max_contiguous_flag[0], ax='time')
+    return wout
+
+
+def get_max_contiguous_flag_from_filter_periods(x, filter_centers, filter_half_widths):
+    """
+    determine maximum contiguous flags from filter periods
```
nkern, line 172: perhaps we can be a little more verbose here, just b/c its such a niche function.
1. you mention filter periods in the doc but filter periods is not an input.
2. why is a maximum contiguous flag needed? This seems specific to a DPSS / linear filter approach, perhaps you can clarify briefly, thanks

hera_cal/vis_clean.py
```diff
@@ -213,6 +441,77 @@ def apply_calibration(self, input_cal, unapply=False):
         apply_cal.calibrate_in_place(self.data, cal_gains, self.flags, cal_flags,
                                      gain_convention=gain_convention)
 
+    def apply_flags(self, external_flags=None, overwrite_data_flags=False,
```
nkern, line 444: I'm still not enthused about how this is done, in the sense that it's not as consistent with `apply_calibration` as I would like it. I understand that we want to use convenience methods that operate on UVData subclasses. I'll try to see if I can make this any more elegant--I'll try to get back to you shortly on this. thanks

hera_cal/vis_clean.py
anguta: I've found that when performing linear inpainting with both DPSS and DFT modes, you tend to get  artifacts inside of gaps that are significantly larger then the smallest wavelength mode that is being used to interpolate. 

A more classical example that you might be familiar with is when you try to fit a high order polynomial to data with large gaps, you can end up getting nonsense structures inside of the gaps.

anguta: I think this arises because there are degeneracies of what should be inside of the large gap given the SNR of the data being fitted. This makes me think that there is probably a better way to fit the modes that we should be taking advantage of but better priors for how the gap structure should behave (beyond that it is composed of the linear sum of our fitting basis) for now, subject of future work ;)

nkern:
@anguta thanks for adopting the suggested changes. some last minor docstring comments and I think we are good to go, assuming Josh is okay. lmk when you've got a final draft and I'll approve, thanks

hera_cal/vis_clean.py
nkern: Okay, so the result you propose is to flag the entire channel or integration? Or to just leave that stretch of channels (or integrations) flagged?

nkern:
@anguta I don't understand the `flag_zero_times` kwarg. Let me state what I think `apply_flags` is trying to do, and you correct me
1. default behavior is to OR the existing flags in self.flags with an external set of flags, be it through `external_flags` and/or `a_priori_flags`. This can only add flags, not remove them (b/c of the OR).
2. optional feature is to __overwrite__ the flags (instead of OR), such that we can indeed remove flags if a pixel is not flagged in the provided flags (unless the antenna is fully flagged, in which case we keep it flagged)
3. the `flag_zero_times` kwarg says that if an integration (or freq channel? I don't understand based on the language whether its a horizontal or vertical stripe of flags) is flagged for all channels, keep it flagged, even if overwrite_data_flags is True. The use-case is given as a secondary round of RFI flagging on a previous set of RFI flags, where we want to keep fully flagged times.

In this case, it seems that `flag_zero_times` is wholly unnecessary--because the default behavior is to keep existing flags. In the case where we are doing secondary RFI flagging, we wouldn't want to unflag existing flags anyways. If its the case that you would like to actually overwrite the flags but specifically keep fully flagged times, then this is something so specific that the user should do it on their own before calling `apply_flags`

nkern:
furthermore, I think I saw somewhere you said that the external flags times and freqs must not necessarily match the data, but the fact that you do
```
            external_flags.time_array = self.times
            external_flags.freq_array = self.freqs
```
seems to imply the opposite

nkern:
Also, the `a_priori_flags_yaml` is also something that doesn't need to be in vis_clean. It is a two-liner, and only uses other tested and modular methods, so it can be done by the user or in a drive script.
```
V = VisClean(files)
V.read()

import hera_qm.utils as qm_utils
V.hd = qm_utils.apply_yaml_flags(V.hd, a_priori_flag_yaml)
_, V.flags, _ = V.hd.build_datacontainers()
```

anguta:
> @anguta I don't understand the `flag_zero_times` kwarg. Let me state what I think `apply_flags` is trying to do, and you correct me
> 
> 1. default behavior is to OR the existing flags in self.flags with an external set of flags, be it through `external_flags` and/or `a_priori_flags`. This can only add flags, not remove them (b/c of the OR).
> 2. optional feature is to **overwrite** the flags (instead of OR), such that we can indeed remove flags if a pixel is not flagged in the provided flags (unless the antenna is fully flagged, in which case we keep it flagged)
> 3. the `flag_zero_times` kwarg says that if an integration (or freq channel? I don't understand based on the language whether its a horizontal or vertical stripe of flags) is flagged for all channels, keep it flagged, even if overwrite_data_flags is True. The use-case is given as a secondary round of RFI flagging on a previous set of RFI flags, where we want to keep fully flagged times.
> 
> In this case, it seems that `flag_zero_times` is wholly unnecessary--because the default behavior is to keep existing flags. In the case where we are doing secondary RFI flagging, we wouldn't want to unflag existing flags anyways. If its the case that you would like to actually overwrite the flags but specifically keep fully flagged times, then this is something so specific that the user should do it on their own before calling `apply_flags`

The use case for this is kind of esoteric. There were some days that had flags that I wanted to use different flags on but also had the data set to zero at certain integrations but at the same time were not in the external flags I was applying. Never really got to the bottom of why this was. Maybe this is overspecializing the code. If you want, we can remove it.

anguta:
> Also, the `a_priori_flags_yaml` is also something that doesn't need to be in vis_clean. It is a two-liner, and only uses other tested and modular methods, so it can be done by the user or in a drive script.
> 
> ```
> V = VisClean(files)
> V.read()
> 
> import hera_qm.utils as qm_utils
> V.hd = qm_utils.apply_yaml_flags(V.hd, a_priori_flag_yaml)
> _, V.flags, _ = V.hd.build_datacontainers()
> ```
> 
> The reason why I'm pushing back a little on it is b/c it is partially responsible for the reason why its harder to not make a copy of `hd` when applying the flags (and also introducing the hera_qm jargon of `a_priori_flags` seems a little out of place for a method like this)

I'm down with removing the `a_priori_flags` jargon which I agree isn't very descriptive here since all of the flags are `a_priori` (we arent doing flagging here). I do think its nice to support applying `yaml` flags though. How about we add a `filetype` arg to `apply_flags` and let it be a `yaml` or a `uvflag` and get rid of the `a_priori_yaml` arg?

nkern:
@anguta Here is how I think apply_flags should look like, I think given how we have chosen to store and apply flags across the HERA ecosystem, its kind of inevitable to make a partial if not full copy of `self.hd` for the time being. Note that I'm using the `pyuvdata.utils.apply_uvflags` method, not the hera_qm method
```
    def apply_flags(self, external_flags, overwrite_flags=False):
        """
        Apply external set of flags to self.flags. Default is to OR existing flags,
        unless overwrite_flags is True.

        Parameters
        ----------
        external_flags: str or UVFlag object
            Str or list of strings pointing to flag files to apply.
            flag files should be in a format readable by UVFlag.
        overwrite_flags: bool, optional
            If True, overwrite flags (instead of OR) for baselines that are not entirely flagged.
        """
        if isinstance(external_flags, str):
            external_flags = UVFlag(external_flags)

        # get a census of fully flagged baselines to re-flag if overwrite_flags
        if overwrite_flags:
            full_flags = [bl for bl in self.flags if np.all(self.flags[bl])]

        # apply flags
        hd_temp = uvutils.apply_uvflag(self.hd, external_flags, unflag_first=overwrite_flags, inplace=False)

        # generate flags
        _, self.flags, _ = hd_temp.build_datacontainers()

        # flag fully flagged baselines if necessary
        if overwrite_flags:
            for bl in full_flags:
                self.flags[bl][:] = True
```

Note that pyuvdata already has a method for checking the times and freqs between `uvf` and `uvd`, which is set by the `uvf._time_array.tols` variable. So you could presumable change that before applying the flags to elegantly solve your issue, I think

nkern:
@anguta actually nvm, both of our solutions failed to neglect the possible flags that are stored in the attached calibration file, which is only applied dynamically (i.e. on-the-fly) when `attach_data()` is called.

Okay, I think this kind of makes a strong case to simply overwrite the flags in `self.hd`, which I think I'm fine with in this case. Here is a possible solution
```
    def apply_flags(self, external_flags, overwrite_flags=False):
        """
        Apply external set of flags to self.hd.flag_array (inplace!), and re-attach the data and flags.
        Default is to OR existing flags, unless overwrite_flags is True.

        Parameters
        ----------
        external_flags: str or UVFlag object
            Str or list of strings pointing to flag files to apply.
            flag files should be in a format readable by UVFlag.
        overwrite_flags: bool, optional
            If True, overwrite flags (instead of OR) for baselines that are not entirely flagged.
        """
        if isinstance(external_flags, str):
            external_flags = UVFlag(external_flags)

        # get a census of fully flagged baselines to re-flag if overwrite_flags
        if overwrite_flags:
            full_flags = [bl for bl in self.flags if np.all(self.flags[bl])]

        # apply flags
        uvutils.apply_uvflag(self.hd, external_flags, unflag_first=overwrite_flags, inplace=True)

        # re-flag fully flagged baselines if necessary
        if overwrite_flags:
            for bl in full_flags:
                tinds = self.hd.antpair2ind(bl)
                self.hd.flag_array[tinds] = True
        
        # attach data
        self.attach_data()
```

anguta:
> @anguta actually nvm, both of our solutions failed to neglect the possible flags that are stored in the attached calibration file, which is only applied dynamically (i.e. on-the-fly) when `attach_data()` is called.
> 
> Okay, I think this kind of makes a strong case to simply overwrite the flags in `self.hd`, which I think I'm fine with in this case. Here is a possible solution
> 
> ```
>     def apply_flags(self, external_flags, overwrite_flags=False):
>         """
>         Apply external set of flags to self.hd.flag_array (inplace!), and re-attach the data and flags.
>         Default is to OR existing flags, unless overwrite_flags is True.
> 
>         Parameters
>         ----------
>         external_flags: str or UVFlag object
>             Str or list of strings pointing to flag files to apply.
>             flag files should be in a format readable by UVFlag.
>         overwrite_flags: bool, optional
>             If True, overwrite flags (instead of OR) for baselines that are not entirely flagged.
>         """
>         if isinstance(external_flags, str):
>             external_flags = UVFlag(external_flags)
> 
>         # get a census of fully flagged baselines to re-flag if overwrite_flags
>         if overwrite_flags:
>             full_flags = [bl for bl in self.flags if np.all(self.flags[bl])]
> 
>         # apply flags
>         uvutils.apply_uvflag(self.hd, external_flags, unflag_first=overwrite_flags, inplace=True)
> 
>         # re-flag fully flagged baselines if necessary
>         if overwrite_flags:
>             for bl in full_flags:
>                 tinds = self.hd.antpair2ind(bl)
>                 self.hd.flag_array[tinds] = True
>         
>         # attach data
>         self.attach_data()
> ```

Thanks for the advice on `apply_flags` @nkern! I've rewritten this function incorporating most of your suggestions and I think it is a lot cleaner now. Let me know if you have any final recommendations!

-Aaron

anguta:
> > @anguta actually nvm, both of our solutions failed to neglect the possible flags that are stored in the attached calibration file, which is only applied dynamically (i.e. on-the-fly) when `attach_data()` is called.
> > Okay, I think this kind of makes a strong case to simply overwrite the flags in `self.hd`, which I think I'm fine with in this case. Here is a possible solution
> > ```
> >     def apply_flags(self, external_flags, overwrite_flags=False):
> >         """
> >         Apply external set of flags to self.hd.flag_array (inplace!), and re-attach the data and flags.
> >         Default is to OR existing flags, unless overwrite_flags is True.
> > 
> >         Parameters
> >         ----------
> >         external_flags: str or UVFlag object
> >             Str or list of strings pointing to flag files to apply.
> >             flag files should be in a format readable by UVFlag.
> >         overwrite_flags: bool, optional
> >             If True, overwrite flags (instead of OR) for baselines that are not entirely flagged.
> >         """
> >         if isinstance(external_flags, str):
> >             external_flags = UVFlag(external_flags)
> > 
> >         # get a census of fully flagged baselines to re-flag if overwrite_flags
> >         if overwrite_flags:
> >             full_flags = [bl for bl in self.flags if np.all(self.flags[bl])]
> > 
> >         # apply flags
> >         uvutils.apply_uvflag(self.hd, external_flags, unflag_first=overwrite_flags, inplace=True)
> > 
> >         # re-flag fully flagged baselines if necessary
> >         if overwrite_flags:
> >             for bl in full_flags:
> >                 tinds = self.hd.antpair2ind(bl)
> >                 self.hd.flag_array[tinds] = True
> >         
> >         # attach data
> >         self.attach_data()
> > ```
> 
> Thanks for the advice on `apply_flags` @nkern! I've rewritten this function incorporating most of your suggestions and I think it is a lot cleaner now. Let me know if you have any final recommendations!
> 
> -Aaron

Missed some of your other responses. Will get to them.

hera_cal/vis_clean.py
anguta: That's correct. If we are doing frequency axis filtering and there is a contiguous chunk of channels that are flagged that exceeds the wavelength of the smallest sinusoid wavelength being fitted then we flag that integration. Similarly when filtering with time.

hera_cal/vis_clean.py
anguta: Thanks for the feedback. I've followed your recommended template and added an additional arg to support `yaml` files.

anguta:
> > > @anguta actually nvm, both of our solutions failed to neglect the possible flags that are stored in the attached calibration file, which is only applied dynamically (i.e. on-the-fly) when `attach_data()` is called.
> > > Okay, I think this kind of makes a strong case to simply overwrite the flags in `self.hd`, which I think I'm fine with in this case. Here is a possible solution
> > > ```
> > >     def apply_flags(self, external_flags, overwrite_flags=False):
> > >         """
> > >         Apply external set of flags to self.hd.flag_array (inplace!), and re-attach the data and flags.
> > >         Default is to OR existing flags, unless overwrite_flags is True.
> > > 
> > >         Parameters
> > >         ----------
> > >         external_flags: str or UVFlag object
> > >             Str or list of strings pointing to flag files to apply.
> > >             flag files should be in a format readable by UVFlag.
> > >         overwrite_flags: bool, optional
> > >             If True, overwrite flags (instead of OR) for baselines that are not entirely flagged.
> > >         """
> > >         if isinstance(external_flags, str):
> > >             external_flags = UVFlag(external_flags)
> > > 
> > >         # get a census of fully flagged baselines to re-flag if overwrite_flags
> > >         if overwrite_flags:
> > >             full_flags = [bl for bl in self.flags if np.all(self.flags[bl])]
> > > 
> > >         # apply flags
> > >         uvutils.apply_uvflag(self.hd, external_flags, unflag_first=overwrite_flags, inplace=True)
> > > 
> > >         # re-flag fully flagged baselines if necessary
> > >         if overwrite_flags:
> > >             for bl in full_flags:
> > >                 tinds = self.hd.antpair2ind(bl)
> > >                 self.hd.flag_array[tinds] = True
> > >         
> > >         # attach data
> > >         self.attach_data()
> > > ```
> > 
> > 
> > Thanks for the advice on `apply_flags` @nkern! I've rewritten this function incorporating most of your suggestions and I think it is a lot cleaner now. Let me know if you have any final recommendations!
> > -Aaron
> 
> Missed some of your other responses. Will get to them.

OK, I think everything is addressed @nkern . I think this is ready for another look :)

hera_cal/vis_clean.py
```diff
@@ -213,6 +442,46 @@ def apply_calibration(self, input_cal, unapply=False):
         apply_cal.calibrate_in_place(self.data, cal_gains, self.flags, cal_flags,
                                      gain_convention=gain_convention)
 
+    def apply_flags(self, external_flags, overwrite_flags=False, type='uvflag'):
```
nkern, line 445: its bad to overwrite built-in python variables, so type is probably not a great choice. how about filetype? :)

hera_cal/vis_clean.py
```diff
@@ -213,6 +442,46 @@ def apply_calibration(self, input_cal, unapply=False):
...
+            uvutils.apply_uvflag(self.hd, external_flags, unflag_first=overwrite_flags, inplace=True)
+        elif type == 'yaml':
+            from hera_qm.utils import apply_yaml_flags
+            self.hd = apply_yaml_flags(self.hd, external_flags, unflag_first=overwrite_flags)
+        else:
+            raise ValueError(f"{type} is an invalid type! Must be 'yaml' or 'uvflag'.")
+        # re-flag fully flagged baselines if necessary
+        if overwrite_flags:
+            for bl in full_flags:
+                tinds = self.hd.antpair2ind(bl)
```
nkern, line 479: codecov says this is uncovered btw

nkern (approved):
okay there are a couple of comments that I left that I think you can address but after those are done I think I'm good with this PR--thanks for your patience! Also I believe @jsdillon hasn't yet approved, so perhaps ask him to take a look, thanks!

hera_cal/vis_clean.py
anguta: Added this.

hera_cal/vis_clean.py
anguta: added this.

hera_cal/vis_clean.py
anguta: Good point. I've changed this to filetype 👍

hera_cal/vis_clean.py
anguta: Thanks for pointing this out. I've added some coverage.

hera_cal/vis_clean.py
plaplant: Unfortunately I don't know of a tool that will automatically reformat docstrings for you. However, once a consistent docstring format is adopted and adhered through throughtout the repo (e.g., `numpy`-style, which this is pretty close to), then you can have `flake8` check that they are formatted correctly. We do this on a few of our repos, here is how it's configured in pyuvdata: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/setup.cfg. You just run `flake8` in the top-level of the repo and it'll check everything for you. (Flake8 does a lot more than docstrings, but you can pick and choose what checks to run.)

It also doesn't have to be an all-or-nothing thing, you can set it up so you only check certain files while the repo moves toward being compliant to a single style. (That's what pyuvdata does.) Also, for completeness, here's what this docstring should be in numpy style:
```
max_contiguous_flags : int or 2-list of ints
    The width of a region corresponding to the largest delay in
    the filter centers and filter_widths.
```
Basically, the return type is all that goes on the first line, and then the longer description is indented below. The full style guide is [here](https://numpydoc.readthedocs.io/en/latest/format.html).

jsdillon (approved):
Looks like questions have been addressed. Thanks @anguta!

One suggestion, either before or after you merge, would be to update the PR description (and maybe make the name a bit more detailed) with the most up-to-date info based on the feedback from Nick and I. Since we don't really do changelogs, this is the next best thing.

PR #680, opened 2021-02-04
Update Iterable imports

plaplant:
Since python3.3, the `Iterable` object should be imported from `collections.abc`, rather than `collections`. The `DeprecationWarning` issued suggests this will no longer be supported in python3.9, so this updates how things are imported to make sure it still works going into the future. There is also a small change to how numpy strings are handled in `io.py` and `utils.py`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.6, 3.7]
+        python-version: [3.7, 3.8, 3.9]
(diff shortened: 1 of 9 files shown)


jsdillon:
While you're doing this, do you want to address the other deprecated collections in the repo, especially OrderedDict?

plaplant:
It turns out the OrderedDict import is fine. The code/tests still produce a fair number of warnings, but I think these are the only ones related to deprecated imports. Though similar to the comment on hera_opm, I wonder if we want to build for python 3.8 and 3.9 as well.

jsdillon (approved):
Looks good to me.

PR #682, opened 2021-02-11
Bring over per_antenna_modified_z_scores from hera_qm

jsdillon:
This function is being eliminated in https://github.com/HERA-Team/hera_qm/pull/393 and since it's only used in hera_cal, it should live here.


dstorer (approved):
Looks fine to me!

PR #684, opened 2021-02-15
Lstbin redundant groups

aewallwi:
Add support in lst binner for averaging baselines that are redundant between nights but are formed from different antenna pairs. Also supports lstbinning different nights with different sets of baselines. Options are also available for performing redundant averaging within each night.

Also added support for LSTbinning sets of nights where some nights have completely different sets of baselines then other nights (and in particular the last night). The current lstbin implementation only supports lstbinning data where the last night contains every baseline that appears on previous nights.


hera_cal/apply_cal.py
```diff
@@ -192,8 +192,7 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
         data_infilename: filename of the data to be calibrated.
         data_outfilename: filename of the resultant data file with the new calibration and flags.
         new_calibration: filename of the calfits file (or a list of filenames) for the calibration
-            to be applied, along with its new flags (if any).
-        old_calibration: filename of the calfits file (or a list of filenames) for the calibration
+        old_calibration: filename of the calfits file (or a list of filenames) for the calibration	        old_calibration: filename of the calfits file for the calibration
```
jsdillon, line 195: What happened here?

hera_cal/io.py
```diff
@@ -1184,6 +1185,8 @@ def load_vis(input_data, return_meta=False, filetype='miriad', pop_autos=False,
         pick_data_ants: boolean, if True and return_meta=True, return only antennas in data
         nested_dict: boolean, if True replace DataContainers with the legacy nested dictionary filetype
             where visibilities and flags are accessed as data[(0,1)]['nn']
+        return_nsamples: boolean, if True, return nsamples data container.
```
jsdillon, line 1188: I don't have any problem with adding this kwarg, per se, but `load_vis` is a really outdated use pattern that we mostly keep around to support old code. Maybe one day we'll deprecate it. Using `HERAData` is the more modern approach and the easier one to maintain.

So all that said, it appears that you're only updating this to support unit tests. I'd probably prefer leaving the old code as is and re-rewring new unit tests with HERAData. Thoughts?

hera_cal/lstbin.py
```diff
@@ -62,6 +63,12 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
     freq_array : type=ndarray, 1D array of unique data frequencies channels in Hz. Needed for rephase.
     lat : type=float, latitude of array in degrees North. Needed for rephase.
     verbose : type=bool, if True report feedback to stdout
+    nsamp_list : list of nsamples arrays
```
jsdillon, line 66: This should appear after `flags_list` and should be called `nsamples_list` for consistency.

hera_cal/lstbin.py
```diff
@@ -62,6 +63,12 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
+    bl_list : optional list of antenna pairs that includes baselines that may not be in the data for the chunk of lsts
+              being processed but may be present in other lst chunks.
+              baselines not in data will be spoofed in the average to keep baselines in lst-averaged files.
+              flags set to True, nsamples set to zero, data set to one.
```
jsdillon, line 70: In general, I set spoofed data to 0 not 1 (but spoofed calibration solutions to 1).

hera_cal/lstbin.py
```diff
@@ -62,6 +63,12 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
     freq_array : type=ndarray, 1D array of unique data frequencies channels in Hz. Needed for rephase.
     lat : type=float, latitude of array in degrees North. Needed for rephase.
     verbose : type=bool, if True report feedback to stdout
+    nsamp_list : list of nsamples arrays
+    bl_list : optional list of antenna pairs that includes baselines that may not be in the data for the chunk of lsts
```
jsdillon, line 67: Does it only take antenna pairs or does it take bl tuples like `(0, 1, 'ee')`?

hera_cal/lstbin.py
```diff
@@ -111,8 +118,12 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
     # themselves hold lists of ndarrays
     data = odict()
     flags = odict()
+    nsamples = odict()
     all_lst_indices = set()
-
+    pols = []
```
jsdillon, line 123: I'd do `list(set([pol for dc in data_list for pol in dc.pols()]))`

hera_cal/lstbin.py
```diff
@@ -175,13 +189,31 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
+                for pol in pols:
+                    key = antpair + (pol,)
+                    if key not in data and utils.reverse_bl(key) not in data:
+                        nsamples[key] = odict({ind: [] for ind in grid_indices})
+                        data[key] = odict({ind: [] for ind in grid_indices})
+                        flags[key] = odict({ind: [] for ind in grid_indices})
+                        for k, ind in enumerate(grid_indices):
+                            nsamples[key][ind].append(np.zeros(Nfreqs))
+                            flags[key][ind].append(np.ones(Nfreqs, dtype=bool))
+                            data[key][ind].append(np.ones(Nfreqs, dtype=complex))
```
jsdillon, line 216: I'd do zeros, not ones here.

hera_cal/lstbin.py
```diff
@@ -267,19 +302,45 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
 
             # take bin average: real and imag separately
             if median:
-                real_avg.append(np.nanmedian(d.real, axis=0))
-                imag_avg.append(np.nanmedian(d.imag, axis=0))
+                # for median to account for varying nsamples, copy each day by nsamps
```
jsdillon, line 305: AFAIK we don't really use the median option. I'd be fine not supporting both `median==True` and `nsamp_list!=None` as long as that is documented.

hera_cal/lstbin.py
```diff
@@ -406,6 +467,8 @@ def lst_bin_arg_parser():
     a.add_argument("--vis_units", default='Jy', type=str, help="visibility units of output files.")
     a.add_argument("--ignore_flags", default=False, action='store_true', help="Ignore flags in data files, such that all input data is included in binning.")
     a.add_argument("--Nbls_to_load", default=None, type=int, help="Number of baselines to load and bin simultaneously. Default is all.")
+    a.add_argument("--average_redundant_baselines", action="store_true", default=False, help="Redundantly average baselines within and between nights.")
```
jsdillon, line 470: You should also propagate these options to the script.

hera_cal/lstbin.py
```diff
@@ -190,13 +222,16 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
     else:
         # keep all lst_grid bins and fill empty ones with unity data and mark as flagged
         for index in range(len(lst_grid)):
-            if index in all_lst_indices:
-                # skip if index already in data
-                continue
+            # I got rid of this statement because it causes failures if a
```
jsdillon, line 225: Once this PR is going to be pulled in, this will be a very confusing comment!

While I think I'm fine taking out this check, I'm wondering why it was here... Feels like a Chesterton's Fence...

hera_cal/lstbin.py
```diff
@@ -556,6 +621,14 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
     Nbls_to_load : int, default=None, Number of baselines to load and bin simultaneously. If Nbls exceeds this
         than iterate over an outer loop until all baselines are binned. Default is to load all baselines at once.
     ignore_flags : bool, if True, ignore the flags in the input files, such that all input data in included in binning.
+    average_redundant_baselines : bool, if True, baselines that are redundant between nights will be averaged together.
+                                  When this is set to true, Nbls_to_load is interpreted as the number of redundant groups
+                                  to load simultaneously. The number of data waterfalls can be substantially larger in some
+                                  cases.
+    include_autos : bool, if True, include autocorrelations in redundant baseline averages.
```
jsdillon, line 628: I'd clarify that autocorrelations are not considered "redundant" with one another.

hera_cal/lstbin.py
```diff
@@ -601,39 +673,45 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
+        for a in hd.antpos:
+            if a not in antpos:
+                antpos[a] = hd.antpos[a]
+    # generate a list of dictionaries which contain the nights occupied by each unique baseline
+    # (or unique baseline group if average_redundant_baselines is true)
+    bldicts = gen_bldicts([io.HERAData(dlists[-1]) for dlists in data_files], bl_error_tol=bl_error_tol, include_autos=include_autos,
+                          redundant=average_redundant_baselines)
+    # store all baselines in this list. They need to be spoofed
+    # if there are some times that dont overlap for nights with different baselines.
+    # all_key_baselines = [list(bldict.values())[0] for bldict in blgroups]
```
jsdillon, line 688: What's this line for?

hera_cal/lstbin.py
```diff
@@ -601,39 +673,45 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
+    # iterate over output LST files
     if Nbls_to_load in [None, 'None', 'none']:
-        Nbls_to_load = Nbls
-    Nblgroups = Nbls // Nbls_to_load + 1
-    blgroups = [bls[i * Nbls_to_load:(i + 1) * Nbls_to_load] for i in range(Nblgroups)]
+        Nbls_to_load = len(bldicts)
+    Nblgroups = len(bldicts) // Nbls_to_load + 1
```
jsdillon, line 692: If `Nbls_to_load` is `None`, then `Nblgroups` is 2... shouldn't it be 1?

hera_cal/lstbin.py
```diff
@@ -672,10 +763,19 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
                             apply_cal.calibrate_in_place(data, gains, data_flags=flags, cal_flags=cal_flags,
                                                          gain_convention=uvc.gain_convention)
 
+                    # redundantly average baselines, keying to baseline group key
+                    # on earliest night.
+                    if average_redundant_baselines:
+                        if ignore_flags:
+                            raise NotImplementedError("average_redundant_baselines with ignore_flags True is not implemented.")
+                        utils.red_average(data=data, flags=flags, nsamples=nsamps,
```
jsdillon, line 771: Does this function handle `include_autos` correctly?

hera_cal/lstbin.py
```diff
@@ -855,3 +961,90 @@ def sigma_clip(array, flags=None, sigma=4.0, axis=0, min_N=4):
...
+        # entry matches (or conjugate matches). If yes, append to that bldict
+        for grp in reds:
+            if redundant:
+                present = False
+                for bldict in bldicts:
+                    # check if baseline group occured in previous nights
+                    # if it did, add it to the appropriate bldict.
+                    for i in bldict:
+                        if np.linalg.norm(blvecs[grp[0]] - blvecs[bldict[i][0]]) <= bl_error_tol or np.linalg.norm(blvecs[grp[0]] + blvecs[bldict[i][0]]) <= bl_error_tol:
+                            # I was having a lot of trouble getting a unittest for two separate cases here because
```
jsdillon, line 1020: I'd probably prefer untested code to unreadable code if the untested thing is so similar to a tested thing. I'm not a dogmatist when it comes to unit testing.

jsdillon: Also, you can do something like 

```
sign = -1
if np.linalg.norm(blvecs[grp[0]] - blvecs[bldict[i][0]]) <= bl_error_tol:
    sign = 1
```

hera_cal/lstbin.py
```diff
@@ -855,3 +961,90 @@ def sigma_clip(array, flags=None, sigma=4.0, axis=0, min_N=4):
...
+                            # I'm not sure how to conveniently create a UVData object with baselines conjugated.
+                            # two cases is more readable then one so I'd prefer to have two.
+                            sign = int((-1) ** int(np.linalg.norm(blvecs[grp[0]] + blvecs[bldict[i][0]]) <= bl_error_tol))
```
jsdillon, line 1023: This is the most unreadable line, so if you can get rid of this exponentiation, that'd be good.

hera_cal/lstbin.py
```diff
@@ -855,3 +961,90 @@ def sigma_clip(array, flags=None, sigma=4.0, axis=0, min_N=4):
...
+                    present = False
+                    # check if baseline occured in previous nights
+                    # if it did, add it to its corresponding baseline dictionary.
+                    for bldict in bldicts:
+                        for i in bldict:
+                            if bl in bldict[i] or bl[::-1] in bldict[i]:
+                                # I was having a lot of trouble getting a unittest for two separate cases here because
+                                # I'm not sure how to conveniently create a UVData object with baselines conjugated.
+                                # two cases is more readable then one so I'd prefer to have two.
+                                sign = int((-1) ** int(bl[::-1] in bldict[i]))
```
jsdillon, line 1042: See above

hera_cal/lstbin.py
```diff
@@ -601,39 +673,45 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
-    Nbls = len(bls)
+    # get antpos over all nights
+    for dlist in data_files:
+        hd = io.HERAData(dlist[-1])
+        for a in hd.antpos:
+            if a not in antpos:
+                antpos[a] = hd.antpos[a]
+    # generate a list of dictionaries which contain the nights occupied by each unique baseline
+    # (or unique baseline group if average_redundant_baselines is true)
+    bldicts = gen_bldicts([io.HERAData(dlists[-1]) for dlists in data_files], bl_error_tol=bl_error_tol, include_autos=include_autos,
```
jsdillon, line 684: can we rename this variable and this function. `bldict` is so generic and this is a new and very specialized object. The code would be more readable if it were something like `bl_nightly_dict` or maybe even more specific than that...

hera_cal/tests/test_lstbin.py
```diff
@@ -29,15 +28,16 @@ def setup_method(self):
...
                            sorted(glob.glob(DATA_PATH + '/zen.2458045.4*XRAA.uvh5'))]
 
-        (self.data1, self.flgs1, self.ap1, a, self.freqs1, t, self.lsts1,
-         p) = io.load_vis(self.data_files[0], return_meta=True, filetype="uvh5")
-        (self.data2, self.flgs2, ap, a, self.freqs2, t, self.lsts2,
-         p) = io.load_vis(self.data_files[1], return_meta=True, filetype="uvh5")
-        (self.data3, self.flgs3, ap, a, self.freqs3, t, self.lsts3,
-         p) = io.load_vis(self.data_files[2], return_meta=True, filetype="uvh5")
+        (self.data1, self.flgs1, self.nsmps1, self.ap1, a, self.freqs1, t, self.lsts1,
+         p) = io.load_vis(self.data_files[0], return_meta=True, filetype="uvh5", return_nsamples=True)
```
jsdillon, line 32: Is the only place you want to use the new features in load_vis? If so, can we just eliminate the use of load_vis?

hera_cal/tests/test_lstbin.py
```diff
@@ -268,6 +317,132 @@ def test_lst_bin_files(self):
...
+                hd.write_uvh5(out_path, clobber=True)
+
+        lstbin.lst_bin_files(redundantly_averaged_filepaths, outdir=tmp_path, overwrite=True, median=False,
+                             verbose=False, file_ext=file_ext, ignore_flags=False, average_redundant_baselines=True)
+        # assert that the keys when only two antennas are present on the first night data
+        # are identical to the data set when all antennas are present on the first night data
+        uv5 = io.HERAData(output_lst_file)
+        d5, f5, n5 = uv5.read()
+        assert sorted(list(d5.keys())) == sorted(list(d4.keys()))
+
```
jsdillon, line 445: gen_bldicts should probably have its own broken-out unit test. How hard would that be to do?

jsdillon (changes requested):
Got a few suggestions for clarity and readability, but generally looks good. 

I also just wanted to give @nkern a heads up in case you wanted to skim the new features and see if anything interferes with his work, since he wrote the original lst binner.

hera_cal/apply_cal.py
anguta: not sure fixed.

hera_cal/io.py
anguta: i've reverted this.

anguta: I've replaced the pattern.

hera_cal/lstbin.py
anguta: I've changed the ordering.

hera_cal/lstbin.py
anguta: I agree.

hera_cal/lstbin.py
anguta: antenna pairs. I've clarified this.

hera_cal/lstbin.py
anguta: I've adopted this.

hera_cal/lstbin.py
anguta: Switched to zeros.

hera_cal/lstbin.py
anguta: i've removed median support for non-None nsamples_list.

hera_cal/lstbin.py
anguta: They are in the script, passed as kwargs.

hera_cal/lstbin.py
anguta: I think the reason it was here was that there was no reason to fill in fake data because this lst had been visited by a previous night and if we assume that all nights have the same baselines, then this baseline has already been spoofed. If we let baselines vary between nights, this assumption is no longer true so we can't necessarily skip spoofing.

anguta: To keep the check, and whatever performance speedups where intended, i've added another check that checks whether baselines are the same across all nights. If they are, then we can keep this check and skip the following spoofing for loop.

hera_cal/lstbin.py
anguta: If you do include `average_redundant_baselines` I think the autos do get averaged. For this reason, we want to lstbin autocorrs seperately (at least in this implementation).

hera_cal/lstbin.py
anguta: I've switched to `bl_nightly_dict`.

hera_cal/lstbin.py
anguta: orphaned code.

hera_cal/lstbin.py
anguta: I've fixed this.

hera_cal/lstbin.py
anguta: it does if you want the autos to be averaged together.

hera_cal/lstbin.py
anguta: I will use your more readable suggestion.

hera_cal/lstbin.py
anguta: yeah.

hera_cal/lstbin.py
anguta: went with your suggestion.

hera_cal/tests/test_lstbin.py
anguta: I've replaced this with `HERAData` convention.

hera_cal/tests/test_lstbin.py
anguta: I've broken out a short unittest.

nkern:
@anguta I would like to have a look at this in more detail. I'll try to get to this soon, thanks for your patience :)

anguta:
Thanks for all the feedback @jsdillon , I think this is ready for another look!

hera_cal/lstbin.py
jsdillon: I'd say change the existing spoof.

hera_cal/lstbin.py
jsdillon: Oh, ok. Then I'd clarify that that's what's going on in the documentation. 

I was thinking that it is generally useful to have LST-binned autos to get the expected noise level, but if we're already talking about redundant averaging, then maybe that's less useful. So I guess this is fine...

jsdillon (approved):
I'm happy with this PR as is (modulo a few tiny comments above). But I think we should wait to pull until @nkern has had a look.

hera_cal/lstbin.py
aewallwi: It is pretty annoying having separate tomls to lstbin autos so maybe this is a feature we should add.

nkern:
@aewallwi okay finally getting to this. Correct me if I'm wrong, but this is essentially just enabling a slight computation speed-up in terms of the way in which we average the data, in the sense that we can with existing infrastructure average across nights and across redundant baselines, but never simultaneously. With the linearity of an average I don't expect there to be a difference between either if I'm using the same weights.

The only difference I can think this may have is for running the median clipping if we are averaging across baselines, which I am *very* hesitant about, b/c the assumptions that go into median clipping RFI are very different than the assumptions about median clipping non-redundancies. I'm not sure yet if that's what your aim is here, I'll try to take a look at it today. But if it's not, then it seems that this PR is mainly a way to bypass excess IO, (in other words, average the data while we have them in memory, rather than writing them out, only to read them back in at a later stage of the pipeline) in which case I'll be looking for how we use existing "in-memory" functions to accomplish the goal, rather than re-writing our averaging functions.

Lmk if you have any thoughts on that thus far, thanks

aewallwi:
> @aewallwi okay finally getting to this. Correct me if I'm wrong, but this is essentially just enabling a slight computation speed-up in terms of the way in which we average the data, in the sense that we can with existing infrastructure average across nights and across redundant baselines, but never simultaneously. With the linearity of an average I don't expect there to be a difference between either if I'm using the same weights.
> 
> The only difference I can think this may have is for running the median clipping if we are averaging across baselines, which I am _very_ hesitant about, b/c the assumptions that go into median clipping RFI are very different than the assumptions about median clipping non-redundancies. I'm not sure yet if that's what your aim is here, I'll try to take a look at it today. But if it's not, then it seems that this PR is mainly a way to bypass excess IO, (in other words, average the data while we have them in memory, rather than writing them out, only to read them back in at a later stage of the pipeline) in which case I'll be looking for how we use existing "in-memory" functions to accomplish the goal, rather than re-writing our averaging functions.
> 
> Lmk if you have any thoughts on that thus far, thanks

I think thats right. Another problem that this PR solves is if our array is not monotonically increasing in size so that different nights have different sets of baselines (earlier nights might have baselines that are not a subset of the last night)

jsdillon:
FWIW, I ran the LST-binner of the four epochs into one final stack for H1C IDR3.2 on this branch. It seems to be working well: https://nbviewer.jupyter.org/github/HERA-Team/H1C_IDR3_Notebooks/blob/main/lst_bin_inspect/h1c_idr3_lst_bin_inspect_all_epochs.ipynb

jsdillon:
@nkern are you OK with this PR? I'd love to get it merged into master, especially since I used it on H1C IDR3

nkern:
@jsdillon I left a few comments for aaron on the `lstbin()` part of this and am not sure if he has gotten back to me yet. The comments say "pending" and aren't resolved. lmk aaron if you have any thoughts on my comments. I'll leave any comments I have on `lstbin_files` below

jsdillon:
@nkern I think if it says "pending" it means you go to "Files Changed" --> "Review Changes" --> "Submit review"

hera_cal/apply_cal.py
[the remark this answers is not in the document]
aewallwi: Looks like it. This has been fixed.

hera_cal/lstbin.py
```diff
@@ -37,6 +38,9 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
         These LST arrays must be monotonically increasing, except for a possible wrap at 2pi.
     flags_list : type=list, list of DataContainer dictionaries holding flags for each data dict
         in data_list. Flagged data do not contribute to the average of an LST bin.
+    nsamples_list : type=list. List of DataContainer dictionaries holding nsamples for each data dict
+        in data_list. Default is None
+        median=False not supported if nsamples_list is not None!
```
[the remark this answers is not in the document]
aewallwi: I'll note that it is used for weighting the average. 

As an aside on memory footprint/performance, I think one of the biggest low-hanging fruit for the lstbinner is to perform the averaging and standard deviations in place on a single numpy array rather then extracting every sample into a list and computing an std and mean on that list which is the current practice. For a future PR though.

hera_cal/lstbin.py
```diff
@@ -78,8 +89,23 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
             un-averaged complex visibilities in each LST bin as values.
         flags_min : dictionary with data flags
     """
+    if nsamples_list is not None and median:
+        raise NotImplementedError("LST binning with median not supported with nsamples_list is not None.")
     # get visibility shape
     Ntimes, Nfreqs = data_list[0][list(data_list[0].keys())[0]].shape
+    # check whether baselines are the same across all nights
```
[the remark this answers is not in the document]
aewallwi: I've broken it out.

hera_cal/lstbin.py
```diff
@@ -147,7 +174,15 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
             d = utils.lst_rephase(d, bls, freq_array, lst_shift, lat=lat, inplace=False)
 
         # iterate over keys in d
-        for j, key in enumerate(d.keys()):
+        klist = list(d.keys())
+        for j, key in enumerate(klist):
+            # if bl_list is not None, use it to determine conjugation:
+            # this is to prevent situations where conjugation of bl in
+            # data_list is different from bl in data which can cause
```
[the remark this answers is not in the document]
aewallwi: different LST chunks can have different sets of nights in them which can have different conjugation conventions on the baselines. Feeding `bl_list` enforces a single conjugation convention.

hera_cal/lstbin.py
```diff
@@ -175,13 +212,39 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
                     data[key][ind].append(d[key][k])
                     # also insert flags if fed
                     if flags_list is None:
                         flags[key][ind].append(np.zeros_like(d[key][k], np.bool))
                     else:
                         flags[key][ind].append(flags_list[i][key][k])
+                    if nsamples_list is None:
+                        nsamples[key][ind].append(np.ones_like(d[key][k], np.int))
+                    else:
+                        nsamples[key][ind].append(nsamples_list[i][key][k])
```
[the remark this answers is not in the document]
aewallwi: added this.

hera_cal/lstbin.py
```diff
@@ -62,6 +67,12 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
     freq_array : type=ndarray, 1D array of unique data frequencies channels in Hz. Needed for rephase.
     lat : type=float, latitude of array in degrees North. Needed for rephase.
     verbose : type=bool, if True report feedback to stdout
+    bl_list : optional list of antenna pairs that includes baselines that may not be in the data for the chunk of lsts
+              being processed but may be present in other lst chunks.
+              baselines not in data will be spoofed in the average to keep baselines in lst-averaged files.
+              flags set to True, nsamples set to zero, data set to zero
```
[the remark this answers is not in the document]
aewallwi: I call this "placeholder data" now.

hera_cal/lstbin.py
```diff
@@ -175,13 +212,39 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
+                    if nsamples_list is None:
+                        nsamples[key][ind].append(np.ones_like(d[key][k], np.int))
+                    else:
+                        nsamples[key][ind].append(nsamples_list[i][key][k])
+        # add in spoofed baselines to keep baselines in different LST files consistent.
+        if bl_list is not None:
+            for antpair in bl_list:
+                for pol in pols:
+                    key = antpair + (pol,)
+                    if key not in data:
```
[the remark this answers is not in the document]
aewallwi: I've combined it with the if statement below as an `and`.

hera_cal/lstbin.py
```diff
@@ -175,13 +212,39 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
+                        if (key[0] != key[1] and utils.reverse_bl(key) not in data) or key[0] == key[1]:
+                            # last part lets us spoof ne and en for autocorrs.
```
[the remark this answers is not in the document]
aewallwi: This makes sure both the `en` and `ne` pols get spoofed for autocorrs. The line you wrote will work fine for cross-correlations but it will only spoof `ne` or `en` (whichever comes first in the polarization list) for an autocorrelation.

hera_cal/lstbin.py
```diff
@@ -175,13 +212,39 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
+                            nsamples[key] = odict({ind: [] for ind in range(len(lst_grid))})
+                            data[key] = odict({ind: [] for ind in range(len(lst_grid))})
+                            flags[key] = odict({ind: [] for ind in range(len(lst_grid))})
+
+                    for ind in range(len(lst_grid)):
```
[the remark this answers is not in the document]
aewallwi: Since different nights have different sets of baselines and different LST bins have different sets of nights, it is possible to get a baseline that appears in a subset of the LSTs within an LST chunk (for example, a baseline that exists in one of the nights that only contained a subset of the LSTs in the LST chunk being processed). I've added a comment explaining this.

hera_cal/lstbin.py
```diff
@@ -175,13 +212,39 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
+                        if ind not in nsamples[key]:
+                            nsamples[key][ind] = []
+                            flags[key][ind] = []
+                            data[key][ind] = []
+                        if len(nsamples[key][ind]) == 0:
+                            nsamples[key][ind].append(np.zeros(Nfreqs))
```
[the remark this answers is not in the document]
aewallwi: Good catch. I've added this.

hera_cal/lstbin.py
```diff
@@ -190,13 +253,16 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
     else:
         # keep all lst_grid bins and fill empty ones with unity data and mark as flagged
         for index in range(len(lst_grid)):
-            if index in all_lst_indices:
-                # skip if index already in data
+            if index in all_lst_indices and baselines_same_across_nights:
```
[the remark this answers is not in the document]
aewallwi: `all_lst_indices` represents the union of LSTs across all nights and baselines. We can have situations where `index` is in `all_lst_indices` but it may not be present for a particular baseline which occurs on a subset of the nights being averaged. This is not a problem if the baselines sets are all the same on all the nights which is why I added the additional `bls_same_across_nights` condition.

hera_cal/lstbin.py
```diff
@@ -190,13 +253,16 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
                 continue
-            for key in list(data.keys()):
-                # fill data with blank data
-                data[key][index] = [np.ones(Nfreqs, np.complex)]
-                flags[key][index] = [np.ones(Nfreqs, np.bool)]
+            else:
+                for key in list(data.keys()):
+                    # fill data with blank data
+                    # if the index is not present.
+                    if index not in data[key]:
```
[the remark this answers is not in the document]
aewallwi: `all_lst_indices` is the union of indices present in the data across all baselines and nights. Some baselines might be present in a subset of nights and those nights might only cover a subset of the LSTs in the lst chunk. Thus, `index` can be in `all_lst_indices` but can be absent for a particular baseline.

hera_cal/lstbin.py
```diff
@@ -267,19 +333,30 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
 
             # take bin average: real and imag separately
             if median:
+                # for median to account for varying nsamples, copy each day by nsamps
```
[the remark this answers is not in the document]
aewallwi: removed.

hera_cal/lstbin.py
```diff
@@ -267,19 +333,30 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
-                real_avg.append(np.nanmean(d.real, axis=0))
-                imag_avg.append(np.nanmean(d.imag, axis=0))
+                # for mean ot account for varying nsamples, take nsamples weighted sum.
+                # (inverse variance weighted sum).
+                resum = np.asarray([np.sum((d.real[:, chan] * n[:, chan])[np.isfinite(d.real[:, chan])]) for chan in range(d.shape[1])])
+                re_nsum = np.asarray([np.sum((n[np.isfinite(d.real[:, chan]), chan])) for chan in range(d.shape[1])])
+                imsum = np.asarray([np.sum((d.imag[:, chan] * n[:, chan])[np.isfinite(d.imag[:, chan])]) for chan in range(d.shape[1])])
+                im_nsum = np.asarray([np.sum((n[np.isfinite(d.imag[:, chan]), chan])) for chan in range(d.shape[1])])
+
+                resum[re_nsum > 0.] /= re_nsum[re_nsum > 0.]
```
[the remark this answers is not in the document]
nkern: well, the clip is only invoked when all weights across the sum axis are zero, in which case the data itself are also zero, so whether the clip is 1e-40, or 1e-99, the result is zero so long as the clip value doesn't somehow get cast as 0 in the computer before taking the division

hera_cal/lstbin.py
```diff
@@ -267,19 +333,30 @@ def lst_bin(data_list, lst_list, flags_list=None, dlst=None, begin_lst=None, lst
...
             # take bin average: real and imag separately
             if median:
+                # for median to account for varying nsamples, copy each day by nsamps
                 real_avg.append(np.nanmedian(d.real, axis=0))
                 imag_avg.append(np.nanmedian(d.imag, axis=0))
             else:
-                real_avg.append(np.nanmean(d.real, axis=0))
-                imag_avg.append(np.nanmean(d.imag, axis=0))
+                # for mean ot account for varying nsamples, take nsamples weighted sum.
+                # (inverse variance weighted sum).
```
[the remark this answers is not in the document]
aewallwi: This is much cleaner. I've adopted it.

hera_cal/lstbin.py
```diff
@@ -552,6 +634,16 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
     Nbls_to_load : int, default=None, Number of baselines to load and bin simultaneously. If Nbls exceeds this
         than iterate over an outer loop until all baselines are binned. Default is to load all baselines at once.
     ignore_flags : bool, if True, ignore the flags in the input files, such that all input data in included in binning.
+    average_redundant_baselines : bool, if True, baselines that are redundant between nights will be averaged together.
+                                  When this is set to true, Nbls_to_load is interpreted as the number of redundant groups
```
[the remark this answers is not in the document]
aewallwi: fixed.

hera_cal/lstbin.py
```diff
@@ -552,6 +634,16 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
+                                  to load simultaneously. The number of data waterfalls can be substantially larger in some
+                                  cases.
+    include_autos : bool, if True, include autocorrelations in redundant baseline averages.
+                    default is True.
+    bl_error_tol : float, tolerance within which baselines are considered redundant
```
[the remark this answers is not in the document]
aewallwi: I've changed this to "within and between nights".

hera_cal/lstbin.py
```diff
@@ -856,3 +996,103 @@ def sigma_clip(array, flags=None, sigma=4.0, axis=0, min_N=4):
...
     return clip_flags
+
+
+def gen_bl_nightly_dicts(hds, bl_error_tol=1.0, include_autos=True, redundant=False, ex_ant_yaml_files=None):
+    """
+    Helper function to generate baseline dicts to keep track of reds between nights.
+
+    Parameters:
+    -----------
+
```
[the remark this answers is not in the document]
aewallwi: fixed.

hera_cal/lstbin.py
```diff
@@ -597,39 +688,43 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
     if Nbls_to_load in [None, 'None', 'none']:
-        Nbls_to_load = Nbls
-    Nblgroups = Nbls // Nbls_to_load + 1
-    blgroups = [bls[i * Nbls_to_load:(i + 1) * Nbls_to_load] for i in range(Nblgroups)]
+        Nbls_to_load = len(bl_nightly_dicts) + 1
+    Nblgroups = len(bl_nightly_dicts) // Nbls_to_load + 1
+    blgroups = [bl_nightly_dicts[i * Nbls_to_load:(i + 1) * Nbls_to_load] for i in range(Nblgroups)]
     blgroups = [blg for blg in blgroups if len(blg) > 0]
-
-    # iterate over output LST files
```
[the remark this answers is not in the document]
aewallwi: It accidentally got moved up a few lines. I've put it back.

hera_cal/lstbin.py
```diff
@@ -647,16 +742,43 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
                     hd = io.HERAData(data_files[j][k], filetype='uvh5')
-                    antpairs = set(hd.antpairs)
-                    bls_in_data = [bl for bl in blgroup if bl in antpairs or bl[::-1] in antpairs]
-                    if len(bls_in_data) == 0:
-                        utils.echo(f"No baselines from blgroup {bi + 1} found in {data_files[j][k]}, skipping file for these bls", verbose=verbose)
-                    data, flags, nsamps = hd.read(bls=bls_in_data, times=tarr[tinds])
+                    try:
+                        bls_to_load = []
+                        key_baselines = []  # map first baseline in each group to
+                        # first baseline in group on earliest night.
```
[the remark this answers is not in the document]
aewallwi: I mean earliest night that has the baseline.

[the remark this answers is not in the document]
aewallwi: Updated comment to make this clear.

hera_cal/lstbin.py
```diff
@@ -647,16 +742,43 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
+                        reds = []
+                        for bl_nightly_dict in blgroup:
+                            # only load group if present in the current night.
+                            if j in bl_nightly_dict:
+                                # key to earliest night with this redundant group.
+                                key_bl = bl_nightly_dict[np.min(list(bl_nightly_dict.keys()))][0]
+                                key_baselines.append(key_bl)
+                                reds.append(bl_nightly_dict[j])
+                                for bl in bl_nightly_dict[j]:
```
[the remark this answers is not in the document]
aewallwi: Good point. I use extend now.

aewallwi:
> @jsdillon I left a few comments for aaron on the `lstbin()` part of this and am not sure if he has gotten back to me yet. The comments say "pending" and aren't resolved. lmk aaron if you have any thoughts on my comments. I'll leave any comments I have on `lstbin_files` below

Thanks for the comments @nkern! I've managed to go through and respond/address your comments. I think this is ready for another look.

nkern (approved):
looks good, thanks aew!

PR #685, opened 2021-02-16
Ensure that flagged antennas do not affect the number of abscal degeneracies

jsdillon:
This PR addresses an edge case where flagging an antenna changes the number of tip-tilt degeneracies in abscal. Now `post_redcal_abscal` identifies which antennas are completely flagged, removes them when computing the idealized antenna positions and the associated number of dimensions, and then puts them back in at the origin. Since they have no weight, they don't affect any other calculation.

This is coming up in H1C IDR3 when the outriggers were included, e.g.: 
image


anguta (approved):
This looks good to me. If you think it'd be prudent, we might want to add a unittest that explicitely addresses this edge case or some before/after screen grabs from your H3C analysis.

PR #686, opened 2021-02-18
Remove vestigial parameter that's breaking apply_cal script

jsdillon:
The apply_cal script is calling a parameter that was removed mid-PR. This fixes that.


anguta (approved):
Approved.

PR #688, opened 2021-02-19
Fix np comparisons

anguta:
Fix issue where `np.testing.assert_array_equal` was being applied to scalars which is no longer allowed. To address this, I've changed comparisons between arrays in scalars of the form `np.testing.assert_array_equal(array, scalar)` to `np.testing.assert_array_equal(array, np.ones_like(array) * scalar)`  or, when appropriate `np.zeros_like` in place of `np.ones_like`.

### DIFF
--- a/hera_cal/tests/test_abscal.py
+++ b/hera_cal/tests/test_abscal.py
@@ -266,11 +266,11 @@ class Test_Abscal_Solvers(object):
         model[0, 1, 'ee'][0, 0] = np.nan
         model[0, 1, 'ee'][0, 1] = np.inf
         fit = abscal.abs_amp_lincal(model, data)
-        np.testing.assert_array_equal(fit['A_Jee'], 2.0)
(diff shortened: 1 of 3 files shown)


jsdillon:
Are you sure this isn't allowed? The documentation says it's fine: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

anguta:
> Are you sure this isn't allowed? The documentation says it's fine: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

I see. I was under the impression that this was causing the macos python 3.7 test to fail in #674 but this may not be the case.

jsdillon:
Looks like that PR is passing its tests... Should we close this one?

plaplant:
@anguta as I said on #674, I believe the tests were failing before because of linsolve + numpy v1.20 incompatibilities that have since been fixed. I don't know if we need to make these changes, unless tests are still failing.

anguta:
It looks like all the test now pass on #674 (after rebasing). Thanks @plaplant (sorry I missed your comment). I will close this PR.

PR #689, opened 2021-03-02
Fix edge case in firstcal with antennas that only appear in length-1 redundant groups

jsdillon:
Generally speaking, we pick `max_dims` so that antennas that aren't redundant with anything get flagged (since otherwise they'd introduce an extraneous extra dimension). While we're in the middle of expanding the array over the sector gap, this is a bit tricky to get right.

Today I found an edge case in RTP where an antenna wasn't getting flagged, but never participated in any redundant groups of length greater than 1. In this case, firstcal fails to find a solution for the antenna delay, though logcal and omnical can solve for it. This creates key errors in the pipeline. 

This PR solves that problem by setting the antenna's delay and delay offset to 0. We are free to do this, because both are completely degenerate and unsolvable with redcal. In theory, we could solve for them with abscal, though in the future we'll probably just want to flag the antenna.

### DIFF
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -1539,7 +1539,6 @@ def redundantly_calibrate(data, reds, freqs=None, times_by_bl=None, fc_conv_crit
     '''
     rv = {}  # dictionary of return values
     filtered_reds = filter_reds(reds, max_dims=max_dims)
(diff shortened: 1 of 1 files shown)


anguta (approved):
looks good to me.

PR #691, opened 2021-03-04
Update HERAData to account for telescope antennas vs. data antennas in its auxiliary metadata

jsdillon:
Recent data has all antenna positions in the telescope, which breaks a bunch of my plotting. This will help fix that.


hera_cal/io.py
```diff
@@ -321,6 +322,7 @@ def get_metadata_dict(self):
         '''
         antpos, ants = self.get_ENU_antpos()
```
nkern, line 323: let's make the data_ants=False kwarg explicit here to safeguard us against pyuvdata api changes

nkern (approved):
approved pending minor update

PR #692, opened 2021-03-05
Data antpos attribute

jsdillon:
This PR follows up on #691, adding a new attribute `data_antpos` to HERAData and DataContainer objects. `data_antpos` is to `antpos` as `data_ants` is to `ants`.

This PR also fixes a number of places in the repo where antenna positions were used (most notably to get a list of lists of redundant baselines) where we really meant to use just the antennas in the data.

 I'm not 100% sure that I've found all the problems introduced by the change in the correlator, since a lot of our testing files have antpos matching data_antpos. I think It's probably best to just give things a shot and see if anything breaks.

Also, I accidentally committed all these changes to master instead of to this branch. So I reverted all the changes on master and then re-reverted them on this branch.

### DIFF
--- a/hera_cal/tests/test_io.py
+++ b/hera_cal/tests/test_io.py
@@ -331,8 +331,12 @@ class Test_HERAData(object):
             assert isinstance(dc, DataContainer)
             for k in dc.antpos.keys():
                 assert np.all(dc.antpos[k] == hd.antpos[k])
+            assert len(dc.antpos) == 52
+            assert len(hd.antpos) == 52
             for k in dc.data_antpos.keys():
                 assert np.all(dc.data_antpos[k] == hd.data_antpos[k])
+            assert len(dc.data_antpos) == 2
+            assert len(hd.data_antpos) == 2
             assert np.all(dc.freqs == hd.freqs)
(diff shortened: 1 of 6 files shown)


hera_cal/tests/test_io.py
```diff
@@ -331,6 +331,8 @@ def test_build_datacontainers(self):
             assert isinstance(dc, DataContainer)
             for k in dc.antpos.keys():
                 assert np.all(dc.antpos[k] == hd.antpos[k])
+            for k in dc.data_antpos.keys():
+                assert np.all(dc.data_antpos[k] == hd.data_antpos[k])
```
nkern, line 335: I know this may be trivial, but technically don't we also want a test that does a down-select and shows that len(data_ants) < len(ants)?

jsdillon: The test doesn't actually do a downselect, it's just making sure the attribute is there and it matches between the HERAData object and the datacontainer

jsdillon: Good point. Done

nkern (approved):

PR #694, opened 2021-03-18
Fix Bug in Abscal to Allow Calibration of Antennas Flagged in Model

jsdillon:
In #685, I introduced a bug in abscal where antennas that were flagged in the abscal model, but not in the data, were left unflagged but not fully calibrated. This created problems in H1C IDR3 like this where a few baselines in a redundant group were highly discrepant with the rest.

image

This PR takes a different approach to the problem #685 was trying to solve. Instead of using the `data_wgts` to figure out which antennas to exclude, I use the redcal gain flags. However, I wanted to preserve the functionality where flagged antennas still get calibrated as well as possible, so I added the ability for them to get calibrated as well as long as they don't introduce new redcal/abscal degeneracies (this is enforced with the `max_dims` parameter). 

I've also broken out  this functionality into a new helper function, `_get_idealized_antpos()` which makes it easier to unittest.


jsdillon:
Reran 2458085 on a branch and it seemed to work. Here's the before:

image

https://nbviewer.jupyter.org/github/HERA-Team/H1C_IDR3_Notebooks/blob/abf2dfd8f4f8773e968b29d370cfe52c2c389198/abscal_inspect/abscal_inspect_2458085.ipynb


And here's the after:

image
https://nbviewer.jupyter.org/github/HERA-Team/H1C_IDR3_Notebooks/blob/main/abscal_inspect/abscal_inspect_2458085.ipynb

hera_cal/abscal.py
```diff
@@ -3336,6 +3336,72 @@ def build_data_wgts(data_flags, data_nsamples, model_flags, autocorrs, auto_flag
...
+    all_ants = list(cal_flags.keys())
+    unflagged_ants = [ant for ant in cal_flags if not np.all(cal_flags[ant])]
+    all_reds = redcal.get_reds(antpos, bl_error_tol=tol, pols=pols)
+    unflagged_reds = redcal.filter_reds(all_reds, ants=unflagged_ants)
+    
+    # count the number of dimensions describing the redundancies of unflagged antennas
+    unflagged_idealized_antpos = redcal.reds_to_antpos(unflagged_reds, tol=redcal.IDEALIZED_BL_TOL)
+    unflagged_nDims = _count_nDims(unflagged_idealized_antpos, assume_2D=False)
+    
+    # get the potentially calibratable ants, reds, and idealized_antpos. These are antennas may be
```
anguta, line 3369: Thes are antennas may be -> These are antennas that may be

hera_cal/tests/test_abscal.py
```diff
@@ -1144,6 +1144,68 @@ def test_build_data_wgts(self):
                         else:
                             assert wgts[bl][t, f] == 2
 
+    def test_get_idealized_antpos(self):
+
+        # build 7 element hex with 1 outrigger
+        antpos = hex_array(2, split_core=False, outriggers=0)
+        antpos[7] = np.array([100, 0, 0])
```
anguta, line 1151: Add a little more clarification on what you are checking in this test. Something like "ensure that _get_idealised_antpos gives the expected number of dimensions and idealized antennas"

anguta: Also discuss the test cases below.

anguta (changes requested):
The code itself looks solid to me. Just a few docstring clarifications might be helpful.

hera_cal/abscal.py
```diff
@@ -3336,6 +3336,72 @@ def build_data_wgts(data_flags, data_nsamples, model_flags, autocorrs, auto_flag
...
+        tol: float distance for baseline match tolerance in units of baseline vectors (e.g. meters)
+        keep_flagged_ants: If True, flagged antennas that are off-grid (i.e. would introduce an 
+            additional degeneracy) are placed at the origin. Otherwise, flagged antennas in cal_flags
+            are excluded from idealized_antpos.
+        data_wgts: DataContainer mapping baselines like (0, 1, 'ee') to weights. Used to check if 
+            flagged antennas off the calibratable grid have no weight. Ignored if keep_flagged_ants
+            is False.
+
+    Returns:
+        idealized_antpos: dictionary mapping antenna numbers to antenna positions on a perfect grid,
```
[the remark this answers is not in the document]
jsdillon: I've expanded this docstring significantly. It now reads
``` 
idealized_antpos: dictionary mapping antenna numbers to antenna positions on an N-dimensional
            grid where redundant real-world baselines (up to the tol) are perfectly redundant (up to
            numerical precision). These baselines will be arbitrarily linearly transformed (stretched,
            skewed, etc.) and antennas that introduce extra degeneracies will introduce extra dimensions.
            See redcal.reds_to_antpos() for more detail.```

hera_cal/abscal.py
jsdillon: Fixed.

hera_cal/tests/test_abscal.py
jsdillon: I've significantly expanded the comments on the unit test

anguta (approved):

PR #696, opened 2021-03-25
Fix bug in logic for lst_stop in config_lst_bin_files

jsdillon:
When `lst_stop < begin_lst` but `lst_stop` is inferred from the files instead of provided directly (as is the case in using a lstbin toml in hera_opm), then there's no opportunity to increment lst_stop by 2pi. 

In trying to LST-bin h1c_idr3 epoch 2, this lead to lst_stop being before lst_start, and thus no wrapper scripts being produced by build_makeflow_from_config.sh.

I've also removed an unnecessary casting of a list to an array that produces the following deprecation warning:
```VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray```

EDIT:
The above didn't quite work, I think because now the `file_lsts` and the `lst_arrays` are over 2 pi out of phase. I think this PR also fixes that, but it's tricky to unittest and so I'm trying it with real data.

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -468,9 +468,6 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
         lst_arrays.append(larrs)
         time_arrays.append(tarrs)
 
-    lst_arrays = np.asarray(lst_arrays)
-    time_arrays = np.asarray(time_arrays)
-
     # get starting LST for output binning
     if lst_start is None:
         lst_start = lmin
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -638,9 +638,9 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
                     larr[larr < larr[0]] += 2 * np.pi
 
                     # phase wrap larr to get it to fall within 2pi of file_lists
-                    while larr[-1] < fmin - 2 * np.pi:
+                    while larr[0] + 2 * np.pi < fmax:
                         larr += 2 * np.pi
-                    while larr[0] > fmax + 2 * np.pi:
+                    while larr[-1] - 2 * np.pi > fmin:
                         larr -= 2 * np.pi
 
                     # check if this file has overlap with output file
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -490,9 +490,8 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
     # get stopping LST for output binning
     if lst_stop is None:
         lst_stop = lmax
-    else:
-        if lst_stop < begin_lst:
-            lst_stop += 2 * np.pi
+    if lst_stop < begin_lst:
+        lst_stop += 2 * np.pi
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jsdillon** [comment]
Some encouraging news. I ran LST-binning of epoch 2 of H1C IDR3 on this branch and it worked well. The range of LSTs is precisely what I'd expect (the files run from 0.11282323 hours to 14.70586176 hours) and everything looks normal:

Check out https://nbviewer.jupyter.org/github/HERA-Team/H1C_IDR3_Notebooks/blob/main/lst_bin_inspect/h1c_idr3_lst_bin_inspect_epoch_2.ipynb

File: hera_cal/lstbin.py [hunk 1]
```diff
@@ -675,7 +677,7 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
                     file_list.append(data_files[j][k])
                     nightly_data_list.append(data)  # this is data
                     nightly_flgs_list.append(flags)  # this is flgs
-                    nightly_lst_list.append(larr[tinds])  # this is lsts
+                    nightly_lst_list.append(data.lsts)  # this is lsts
```
nkern, line 680: can you remind me if this was a necessary change, or a cosmetic change? I know that a unique copy of the lsts are attached to the data DataContainer, but I'm wondering why larr[tinds] doesn't work also. My concern is that it is very clear what `larr[tinds]` is, but with the change we are now tying ourselves to be dependent on however data.lsts is defined, which I'm wary may for some reason change without us knowing in the future. 

lmk thanks

hera_cal/lstbin.py
```diff
@@ -641,6 +637,12 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
                     tarr = time_arrs[j][k]
                     larr[larr < larr[0]] += 2 * np.pi
 
+                    # phase wrap larr to get it to fall within 2pi of file_lists
+                    while larr[0] + 2 * np.pi < fmax:
+                        larr += 2 * np.pi
+                    while larr[-1] - 2 * np.pi > fmin:
```
nkern, line 643: it seems odd to shift down an octave here when we shifted up before. is this necessary?

jsdillon:
Given that @nkern approved #701, which included this, I'm taking that as effective approval of this as well.

hera_cal/lstbin.py
jsdillon: I think this is necessary. We were previously not passing unwrapped LSTs to the lstbinner with `larr`. Since `larr` is now getting modified, this change prevents the modifed version from escaping outside this function. I could just copy larr before modifying and pass the copy instead.

nkern:
@jsdillon ha, I actually didn't mean for that (was going to make it explicit in that post but thought it was obvious, my bad). anyways I think this is probably fine as is. I would have liked to see larr[tinds] stay as it was, as I think the lst arrays are always wrapped in our data writing, but I don't expect this to be a problem as you've written it.

jsdillon:
Sorry for the misunderstanding there. Glad it's all OK.

PR #697, opened 2021-03-26
Handle baseline -> time cornerturn when on data falls within the time limits of the template file.

aewallwi:
Handle situation in baseline_chunk_to_time_chunk where the baseline chunk has no times within the bounds of the template time chunk file by raising a warning and not writing any output.

### DIFF
--- a/hera_cal/vis_clean.py
+++ b/hera_cal/vis_clean.py
@@ -1835,8 +1835,12 @@ def time_chunk_from_baseline_chunks(time_chunk_template, baseline_chunk_files, o
         # this is to prevent integrations that straddle file boundaries from being dropped.
         # when we perform reconstitution.
         t_select = (hd_baseline_chunk.times >= tmin) & (hd_baseline_chunk.times < tmax)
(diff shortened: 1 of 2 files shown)


hera_cal/vis_clean.py
```diff
@@ -1835,8 +1835,11 @@ def time_chunk_from_baseline_chunks(time_chunk_template, baseline_chunk_files, o
         # this is to prevent integrations that straddle file boundaries from being dropped.
         # when we perform reconstitution.
         t_select = (hd_baseline_chunk.times >= tmin) & (hd_baseline_chunk.times < tmax)
-        hd_combined.read(times=hd_baseline_chunk.times[t_select], axis='blt')
-        hd_combined.write_uvh5(outfilename, clobber=clobber)
+        if np.count_nonzero(t_select) > 0:
```
jsdillon, line 1838: To make this more readable, I'd suggest `np.any`

jsdillon (approved):
Just one minor suggestion, but otherwise looks fine.

hera_cal/vis_clean.py
aewallwi: changed this.

PR #700, opened 2021-04-04
Time inpainting

### LINKED ISSUE #733: Make sure to pass filter_spw_ranges to `topat_frfillter_run.py`
`tophat_frfilter_run.py` in #700 does not pass the `filter_spw_ranges` arg in #725 . Make sure to add this to #700 once #725 gets merged.

aewallwi:
Depends on #704. Closes #733 

Enables the production of "cats-eye" fringe-rate delay plots without flagging sidelobes. Getting rid of these sidelobes requries us to time-inpaint with all of the fringe-rates of the sky for each baseline. This PR allows us to produce these plots as part of RTP.

Also removes `xtalk_filter.py` filter module and moves contents to `frf.py`. The `FRFilter` object now supports both the new sky-fringe-rate inpainting mode and the old `max_frate_coeffs` mode used in `xtalk_filter.py`.

* removes`xtalk_filter.py`.  Move functionality of `xtalk_filter.run_xtalk_filter()` to `frf.run_tophat_frfilter()`.
* adds a new filtering mode where `max_frate`  is determined by the maximum and minimum fringe-rates of each baseline that can correspond to sky emission. This is necessary to perform fringe-rate inpainting of sky-modes and allows us to make the "cats-eye" plots. 

This should be merged after #725 so we can add the `filter_spw_ranges` arg to `tophat_frfilter_run.py`.


aewallwi:
> * Driver methods in `frf.py` to perform time-inpainting of data using the range of fringe-rates on each baseline that are occupied by sky emission. The `model` files are effectively fringe-rate filtered waterfalls assuming a top-hat fringe-rate filter that covers the range of fringe-rates on the sky. Right now, the methods use `fr_filter` but it make sense to rename this as a `tophat_fr_filter` instead.
> * Two driver scripts: `fr_filter_run.py` and `fr_filter_baseline_parallelized.py`.
> * Unittests for driver methods in `test_frf.py`.
> 
> Also adjusted a lot of the equalities in the in-painting drivers scripts to have spaces around the equals signs which makes this PR look a lot larger then it actually is.

A lot of the code is copy-pasted from `xtalk_filter` drivers. The main difference is that the `max_frate` dictionary fed to `vis_clean` now contains the max / min fringe-rates that we would expect on the sky given each baseline's length and orientation (rather then a manual linear fit using `max_frate_coeffs`. A more elegant solution may therefore be to subsume the `xtalk_filter` functionality back into `fr_filter` and add an option specifying the method for calculating `max_frates`. I'd be interested to hear people's thoughts on this alternative solution which would get rid of the `xtalk_filter` classes and drivers.

hera_cal/vis_clean.py
```diff
@@ -1412,6 +1412,61 @@ def write_filtered_data(self, res_outfilename=None, CLEAN_outfilename=None, fill
...
+            to_filter = self.data.keys()
+        # compute maximum fringe rate dict based on baseline lengths.
+        blcosines = {k: self.blvecs[k[:2]][0] / np.linalg.norm(self.blvecs[k[:2]]) for k in to_filter}
+        frateamps = {k: 1. / (24. * 3.6) * self.freqs.max() / 3e8 * 2 * np.pi * np.linalg.norm(self.blvecs[k[:2]]) for k in to_filter}
+        # set autocorrs to have blcose of 0.0
+        for k in blcosines:
+            if np.isnan(blcosines[k]):
+                blcosines[k] = 0.0
+        sinlat = np.sin(np.abs(self.hd.telescope_location_lat_lon_alt[0]))
+        max_frates = io.DataContainer({})
```
nkern, line 1452: Should these technically be in DataContainers? There are some special methods on DataContainer specifically for working with complex visibilities (automatic conjugations, etc). @jsdillon might weigh in. perhaps these are just better as normal dicts

hera_cal/vis_clean.py
```diff
@@ -1412,6 +1412,61 @@ def write_filtered_data(self, res_outfilename=None, CLEAN_outfilename=None, fill
                         self.write_data(data_out, outfilename, filetype=filetype, overwrite=clobber, flags=flags_out,
                                         add_to_history=add_to_history, extra_attrs=extra_attrs, **kwargs)
 
+    def sky_frates(self, to_filter=None, frate_standoff=0.0, frac_frate_sky_max=1.0, min_frate=0.025):
```
nkern, line 1415: this would be nice as a standalone function that we can use in sandbox mode. you can still keep a class method sky_frates(self, ...) but let's have that call the standalone function as needed. lmk if that makes sense

hera_cal/frf.py
```diff
@@ -497,6 +498,85 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+    def run_tophat_frfilter(self, to_filter=None, weight_dict=None, mode='clean',
+                            frate_standoff=0.0, frac_frate_sky_max=1.0, min_frate=0.025,
+                            max_frate_coeffs=None,
+                            skip_wgt=0.1, tol=1e-9, verbose=False, cache_dir=None, read_cache=False,
+                            write_cache=False,
+                            data=None, flags=None, **filter_kwargs):
+        '''
+        Interpolate / filter data in time using the physical fringe-rates of the sky. (or constant frate)
+        Arguments:
+          to_filter: list of visibilities to filter in the (i,j,pol) format.
```
nkern, line 510: I know the delay filter kwarg uses `to_filter`, but in `VisClean` and `FRF` we use the convention `keys` for this kwarg. if possible let's keep this consistent as its a subclass

hera_cal/frf.py
```diff
@@ -497,6 +498,85 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+              If None (the default), all visibilities are filtered.
+          weight_dict: dictionary or DataContainer with all the same keys as self.data.
```
nkern, line 512: can we use `wgts` for this? That is also more consistent with other usages of this kind of term

hera_cal/frf.py
```diff
@@ -497,6 +498,85 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+            Maximum fringe-rate coefficients for the model max_frate [mHz] = x1 * EW_bl_len [ m ] + x2."
+            Providing these overrides the sky-based fringe-rate determination! Default is None.
+          skip_wgt: skips filtering rows with very low total weight (unflagged fraction ~< skip_wgt).
+              Model is left as 0s, residual is left as data, and info is {'skipped': True} for that
+              time. Skipped channels are then flagged in self.flags.
+              Only works properly when all weights are all between 0 and 1.
+          tol : float, optional. To what level are foregrounds subtracted.
+          verbose: If True print feedback to stdout
+          cache_dir: string, optional, path to cache file that contains pre-computed dayenu matrices.
+                      see uvtools.dspec.dayenu_filter for key formats.
```
nkern, line 535: different indentation

hera_cal/frf.py
```diff
@@ -497,6 +498,85 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+          read_cache: bool, If true, read existing cache files in cache_dir before running.
+          write_cache: bool. If true, create new cache file with precomputed matrices
+                             that were not in previously loaded cache files.
```
nkern, line 538: same

hera_cal/frf.py
```diff
@@ -497,6 +498,85 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+                            skip_wgt=0.1, tol=1e-9, verbose=False, cache_dir=None, read_cache=False,
+                            write_cache=False,
+                            data=None, flags=None, **filter_kwargs):
+        '''
+        Interpolate / filter data in time using the physical fringe-rates of the sky. (or constant frate)
+        Arguments:
+          to_filter: list of visibilities to filter in the (i,j,pol) format.
+              If None (the default), all visibilities are filtered.
+          weight_dict: dictionary or DataContainer with all the same keys as self.data.
+              Linear multiplicative weights to use for the delay filter. Default, use np.logical_not
```
nkern, line 513: `...delay filter..`

hera_cal/frf.py
```diff
@@ -497,6 +498,85 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
             filt_flags[k] = f
             filt_nsamples[k] = eff_nsamples
 
+    def run_tophat_frfilter(self, to_filter=None, weight_dict=None, mode='clean',
```
nkern, line 501: So this is kind of just a wrapper around `VisClean.fourier_filter` with extra tools for handling sky_frate kwarsg. Can we make that more clear in the name and in docstring, just b/c there is now a proliferation of "filtering" functions that this subclass inherits. (I also think `run_` should be reserved for module-level functions used directly by drive scripts) Perhaps something like

```
def tophat_frfilter(self, keys=None, wgts=None, ...):
    A wrapper around VisClean.fourier_filter specifically for
    filtering along the time axis with uniform fringe-rate weighting.
````

also is there a reason why this is specifically tophat weighting? It seems it would be nice to allow
for arbitrary fr-weighting. but perhaps that is actually much harder when we are using the `fourier_filter`
function and all of its possible modes. I just raise this as it seems like this is inching towards the direction
that ARP brought of up optimal or semi-optimal FR filtering. it would be nice to be able to do this simultaneously with inpainting.

File: hera_cal/frf.py [hunk 9]
```diff
@@ -558,6 +638,154 @@ def time_avg_data_and_write(input_data_list, output_data, t_avg, baseline_list=N
...
+        Nbls_per_load: int, the number of baselines to load at once.
+            If None, load all baselines at once. default : None.
+        spw_range: 2-tuple or 2-list, spw_range of data to filter.
+        cache_dir: string, optional, path to cache file that contains pre-computed dayenu matrices.
+            see uvtools.dspec.dayenu_filter for key formats.
+        read_cache: bool, If true, read existing cache files in cache_dir before running.
+        write_cache: bool. If true, create new cache file with precomputed matrices
+            that were not in previously loaded cache files.
+        factorize_flags: bool, optional
+            If True, factorize flags before running delay filter. See vis_clean.factorize_flags.
```
nkern, line 694: delay filter

hera_cal/vis_clean.py
aewallwi: I think that makes sense. I will change these.

hera_cal/frf.py
aewallwi: will change to be consistent.

hera_cal/frf.py
aewallwi: Yep.

hera_cal/frf.py
aewallwi: 👍

hera_cal/frf.py
aewallwi: I've changed this to keys.

hera_cal/frf.py
aewallwi: I agree with the name change.  I've fixed up the docstring.


Regarding the tophat weighting, this is the only possible functionality I can imagine for `fourier_filter` since the filter fits modes that span a compact region of fringe-rate space to the data with modes that are not necessarily Fourier modes (and can be mapped to specific fringe-rates). For example, DPSS modes are not necessarily super localized to a specific fringe-rate (although they are super concentrated within the bandwidth of the filter). For fringe-rate (fourier) modes, we can up/down weight different modes but I think this is basically what `apply_fir` function does. 

I think the most sensible way to support non-tophat weights in the future is to use `apply_fir` on the clean components after running the `tophat_frfilter` to achieve different fringe-rate weights. This step can be added to the `load_tophat_frfilter_and_write` function and when we do this, we can also change the name to be something more general. The next PR  ( #736 ) has functions to calculate fringe-rate profiles from uvbeam objects so we can use this in conjunction with `frp_to_fir` and `apply_fir`.

hera_cal/frf.py
aewallwi: changed.

hera_cal/frf.py
```diff
@@ -16,11 +16,85 @@
...
+def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, min_frate_half_width=0.025):
+    """Automatically compute sky fringe-rate ranges based on baselines and telescope location.
+
+    Parameters
+    ----------
+    uvd: UVData object
+        uvdata object of data to compute sky-frate limits for.
+    keys: list of antpairpol tuples, optional
+        list of antpairpols to generate sky fringe-rate centers and widths for.
+        Default is None -> use all keys in self.data.
```
nkern, line 37: all keys in uvd

hera_cal/frf.py
```diff
@@ -16,11 +16,85 @@
...
+    """
+    if keys is None:
+        keys = uvd.get_antpairpols()
+    antpos, antnums = uvd.get_ENU_antpos()
+    sinlat = np.sin(np.abs(uvd.telescope_location_lat_lon_alt[0]))
+    frate_centers = {}
+    frate_half_widths = {}
+
+    # compute maximum fringe rate dict based on baseline lengths.
+    for k in keys:
```
nkern, line 64: can you leave a comment on the eqn. number of Parsons 2016 you are using here?

hera_cal/frf.py
```diff
@@ -16,11 +16,85 @@
...
 
 
-def timeavg_waterfall(data, Navg, flags=None, nsamples=None, wgt_by_nsample=True, 
-                      wgt_by_favg_nsample=False, rephase=False, lsts=None, freqs=None, 
+SPEED_OF_LIGHT = const.c.si.value
+SDAY_KSEC = 86163.93 / 1000.
+
+
+def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, min_frate_half_width=0.025):
+    """Automatically compute sky fringe-rate ranges based on baselines and telescope location.
```
nkern, line 29: Compute sky fringe-rate....baselines, telescope location, and frequencies in uvdata

hera_cal/frf.py
```diff
@@ -16,11 +16,85 @@
...
+
+    Parameters
+    ----------
+    uvd: UVData object
+        uvdata object of data to compute sky-frate limits for.
+    keys: list of antpairpol tuples, optional
+        list of antpairpols to generate sky fringe-rate centers and widths for.
+        Default is None -> use all keys in self.data.
+    frate_standoff: float, optional
+        Additional fringe-rate standoff in mHz to add to Omega_E b_{EW} nu/c for fringe-rate inpainting.
```
nkern, line 39: additional fringe-rate to add to computed fringe-rate bounds [mHz]

mathematic terms are somewhat confusing without definition, and no need to specify fringe-rate inpainting as this func should be general anyways

hera_cal/frf.py
```diff
@@ -16,11 +16,85 @@
...
+        default = 0.0.
+    frate_width_multiplier: float, optional
+        fraction of horizon to fringe-rate filter.
```
nkern, line 42: what is horizon here? frate bounds don't necessarily come from observer horizon. did you mean frate_width of computed frate bounds?

hera_cal/frf.py
```diff
@@ -16,11 +16,85 @@
...
+        default is 1.0
+    min_frate_half_width: float, optional
```
nkern, line 44: other funcs call this variable `min_frate`, shall we use this instead? its not really a minimum half_width, its just a minimum frate, right?

hera_cal/frf.py
aewallwi: The expressions used for max / min fringe-rates don't appear in Parsons 2016 (they are derived from expressions in that paper though). I've added a link to this memo:

https://www.overleaf.com/read/chgpxydbhfhk

hera_cal/frf.py
aewallwi: fixed.

hera_cal/frf.py
aewallwi: fixed.

hera_cal/frf.py
aewallwi: fixed and changed math to "to add to analytic calculation of fringe-rate bounds for emission on the sky."

hera_cal/frf.py
aewallwi: I agree that horizon was a confusing term. I've changed this to "        fraction of width of range of fringe-rates associated with sky emission to filter.
"

hera_cal/frf.py
aewallwi: It is actually a half width. I'm going to change the other functions to agree with this.

aewallwi:
Thanks for the next round of comments! I think I've addressed these now.

hera_cal/frf.py
nkern: for some reason I still see math and not the comment you said above--is my browser page stale or something, or did you forget to push this up?

hera_cal/frf.py
nkern: okay, but the description still reads "minimum frate to filter". does this assume the filter is centered at 0 mHz and is symmetric? Perhaps this can be a *little* more verbose to explain, thanks. also if this works similar to min-frate in other funcs, its not the min-frate to filter (which admittedly is a confusing term), but the minimum bounds on the extent of any fr-filter (i.e. all fr-filters will model out to min_frate or greater)

hera_cal/frf.py
nkern: i don't see the link

hera_cal/frf.py
aewallwi: I think I changed this in one place but not the other. I've fixed it here too.

hera_cal/frf.py
aewallwi: It's here:

 https://github.com/HERA-Team/hera_cal/blob/6a1055b0a6edd46b498300e3272c88844f2cdbbc/hera_cal/frf.py#L65

for some reason it isn't showing up in the diff 🤔

aewallwi: now it's showing up?

hera_cal/frf.py
aewallwi: I've updated this docstring (and other equivalent docstrings for the same arg in other functions) to clarify that this is the minimum half width of the fringe-rate filter, not literally the minimum fringe-rate that will be filtered (equivalent to min_dly arg in the delay_filters).

hera_cal/frf.py
```diff
@@ -16,11 +16,85 @@
...
+            max_frate = np.max([f0 * max_frate_df for f0 in uvd.freq_array[0]])
+        else:
+            max_frate = 0.0
+            min_frate = 0.0
+
+        frate_centers[k] = (max_frate + min_frate) / 2.
+        frate_centers[utils.reverse_bl(k)] = -frate_centers[k]
+
+        frate_half_widths[k] = np.abs(max_frate - min_frate) / 2. * frate_width_multiplier + frate_standoff
+        frate_half_widths[k] = np.max([frate_half_widths[k], min_frate_half_width])  # Don't allow frates smaller then min_frate
```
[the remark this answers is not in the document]
aewallwi: I think there are several good reasons to be ouputing and keeping track of the half-width and center of the filter instead of min/max.

First, it is to use a similar API as delay filtering where `min_dly` means the minimum half_width of the delay filter (the current name is misleading, it does not mean the minimum delay of the filter because that is the negative of that value provided while we are centered at zero). lmk if you think we should refactor the delay-filter API and use something like min_dly_half_width instead to make this name less confusing.

Second, using center and width lets us easily perform non-zero centered fringe-rate filtering by using the Fourier shift trick where we just divide the data by the center fringe-rate and then perform a zero-centered filter using the half-width. 

Third, the half-width is useful to output and carry around for auxiliary calculations (like error bars) because it is the factor that is really important in determining the effective averaging time and temporal coherence of the filtered data.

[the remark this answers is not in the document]
nkern: I see. FYI--`min_dly` is not meant to signify the minimum filtering delay, as its docstring explains it sets a lower bound on the |tau| filter window (although its explanation can certainly be improved). Your wording in the docstring here actually said "the minimum frate to filter", hence the confusion, but I think it is more clear now. Although I actually think equating it to `min_dly in the delay filter` is actually more confusing b/c even though they are both a half-width, they actually have different implications b/c there is no `center_dly`, so min_dly is an absolute parameter if you will whereas min_frate_half_width is relative. I think the doc is good to go if you just get rid of that line. thanks for humoring me on this

hera_cal/frf.py
```diff
@@ -497,6 +576,94 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+        frate_standoff: float, optional
+            additional fringe-rate to add to min and max of computed fringe-rate bounds [mHz]
+            to add to analytic calculation of fringe-rate bounds for emission on the sky.
+            default = 0.0.
+        frate_width_multiplier: float, optional
+         fraction of horizon to fringe-rate filter.
+         default is 1.0
+        min_frate_half_width: float, optional
+            minimum half-width of fringe-rate filter, regardless of baseline length in mHz.
+            Serves the same function as "min_dly" in delay filters.
```
nkern, line 606: same here

hera_cal/frf.py
```diff
@@ -558,6 +725,158 @@ def time_avg_data_and_write(input_data_list, output_data, t_avg, baseline_list=N
...
+
+    '''
+    ap = vis_clean._filter_argparser()
+    filt_options = ap.add_argument_group(title='Options for the fr-filter')
+    ap.add_argument("--frate_width_multiplier", type=float, default=1.0, help="Fraction of maximum sky-fringe-rate to interpolate / filter."
+                                                                              "Used if select_mainlobe is False and max_frate_coeffs not specified.")
+    ap.add_argument("--frate_standoff", type=float, default=0.0, help="Standoff in fringe-rate to filter [mHz]."
+                                                                      "Used of select_mainlobe is False and max_frate_coeffs not specified.")
+    ap.add_argument("--min_frate_half_width", type=float, default=0.025, help="minimum half-width of fringe-rate filter, regardless of baseline length in mHz."
+                                                                              "Serves the same function as 'min_dly' in delay filters."
```
nkern, line 750: same here

hera_cal/frf.py
aewallwi: Fixed.

hera_cal/frf.py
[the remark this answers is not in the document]
aewallwi: Sounds good. I've removed the mention of the min_dly equivalence in the docstring.

hera_cal/frf.py
aewallwi: Fixed.

nkern (approved):
looks good thanks

PR #701, opened 2021-04-06
Fix data loading in LST binner to gracefully handle some (but not all) antennas missing from a group

jsdillon:
This PR fixes a problem I discovered in H1C IDR3 Epoch 3. Basically, if there's an antenna number in the final day that's not in previous days, it causes the whole baseline loading group to fail and to give a very misleading message (just because there's a ValueError doesn't mean that there's no baselines found, just that one is missing). 

This builds on the fix in #696, so that should be pulled in before this.

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -652,13 +652,12 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
 
                     # load data: only times needed for this output LST-bin file
                     hd = io.HERAData(data_files[j][k], filetype='uvh5')
-                    try:
-                        data, flags, nsamps = hd.read(bls=blgroup, times=tarr[tinds])
-                        data.phase_type = 'drift'
-                    except ValueError:
-                        # if no baselines in the file, skip this file
-                        utils.echo("No baselines from blgroup {} found in {}, skipping file for these bls".format(bi + 1, data_files[j][k]), verbose=verbose)
-                        continue
+                    antpairs = set(hd.antpairs)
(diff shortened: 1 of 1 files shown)


nkern:
@jsdillon in your description you single out "the final day", is this error only sensitive to missing antennas in the final day? Or is that simply b/c we choose the final day to construct our master list of baselines and baseline groups?

hera_cal/lstbin.py
```diff
@@ -650,13 +652,12 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
 
                     # load data: only times needed for this output LST-bin file
                     hd = io.HERAData(data_files[j][k], filetype='uvh5')
-                    try:
-                        data, flags, nsamps = hd.read(bls=blgroup, times=tarr[tinds])
```
nkern, line 654: okay so to see if I have this right: this line can fail when only one (or more) baselines in `blgroup` are not present in the `hd` object, even if there are other baselines in `blgroup` that do exist. if that's right, then I think this is indeed a bug. thanks for catching

nkern (approved):
looks good to me, thanks.

hera_cal/lstbin.py
jsdillon: Yes, that's right. It fails if you ask for a baseline with an antenna not in the data.

jsdillon:
> @jsdillon in your description you single out "the final day", is this error only sensitive to missing antennas in the final day? Or is that simply b/c we choose the final day to construct our master list of baselines and baseline groups?

It's the latter. There are baselines on the final day (which we currently use to construct our master list of baselines and baseline groups, though I believe @aewallwi is generalizing that over in #684) which are not in some of the previous days, causing an error which is then caught too broadly.

PR #704, opened 2021-04-13
Fixes from Battletesting

aewallwi:
Fix problems with argparsers and filtering scripts identified during battletesting.

### DIFF
--- a/hera_cal/vis_clean.py
+++ b/hera_cal/vis_clean.py
@@ -1714,6 +1714,7 @@ def _filter_argparser():
     linear_options.add_argument("--max_contiguous_edge_flags", type=int, default=1, help="Skip integrations with at least this number of contiguous edge flags.")
     return ap
 
+
 def time_chunk_from_baseline_chunks(time_chunk_template, baseline_chunk_files, outfilename, clobber=False, time_bounds=False):
     """Combine multiple waterfall files (with disjoint baseline sets) into time-limited file with all baselines.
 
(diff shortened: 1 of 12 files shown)


aewallwi:
Fixed argparsers and filtering scripts in response to script battletesting.

hera_cal/vis_clean.py
```diff
@@ -1672,95 +1672,47 @@ def _filter_argparser():
...
-        None
-    Returns
-    -------
-        Arg-parser for dpss filtering.
-    '''
-    a = _linear_argparser()
-    a.add_argument("--CLEAN_outfilename", default=None, type=str, help="path for writing the filtered model visibilities (with the same flags)")
-    a.add_argument("--filled_outfilename", default=None, type=str, help="path for writing the original data but with flags unflagged and replaced with filtered models wherever possible")
-    return a
+    cache_options = ap.add_argument_group(title='Options for caching (arguments only used if mode!="clean")')
```
jsdillon, line 1709: Can I suggest comments and whitespace between argument groups?

scripts/delay_filter_run.py
```diff
@@ -12,24 +12,24 @@
 a = parser.parse_args()
```
jsdillon, line 12: I think you meant to make this `ap = parser.parse_args()`, otherwise you'll get errors below because `ap` is not defined.

scripts/xtalk_filter_run.py
```diff
@@ -14,25 +14,25 @@
 a = parser.parse_args()
```
jsdillon, line 14: I think you meant for this to be `ap = `

jsdillon (changes requested):
A couple of small bugs and one cosmetic change, but otherwise good to go.

hera_cal/vis_clean.py
aewallwi: I've added white space and comments.

scripts/delay_filter_run.py
aewallwi: thanks for catching this.

scripts/xtalk_filter_run.py
aewallwi: That's correct. Thanks for catching.

aewallwi:
Thanks for the review @jsdillon. I think I've addressed all of the issues.

jsdillon (approved):

PR #705, opened 2021-04-16
Add zeropad kwarg to filter scripts

jsdillon:
This PR allows the `zeropad` kwarg for delay/xtalk filtering to be accessed via the command line (though only as a single integer, not as a tuple).


aewallwi (approved):
looks good.

PR #706, opened 2021-04-16
Create simple redundant averaging script

jsdillon:
This PR creates a very simple script for redundantly averaging data that's already been calibrated. Possible future improvements include better memory management via partial i/o and/or moving parts of this script into its own unit-tested module.

### DIFF
--- a/scripts/red_average.py
+++ b/scripts/red_average.py
@@ -7,7 +7,7 @@
 
 import sys
 import argparse
-from hera_cal import io, utils
+from hera_cal import io, utils, redcal
 
 # Parse arguments
 a = argparse.ArgumentParser(description="Apply (and optionally, also unapply) a calfits file to visibility file.")
--- a/scripts/red_average.py
+++ b/scripts/red_average.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright 2019 the HERA Project
+# Licensed under the MIT License
+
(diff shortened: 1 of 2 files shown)


scripts/red_average.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright 2019 the HERA Project
```
aewallwi, line 3: Maybe update the date.

scripts/red_average.py
```diff
@@ -0,0 +1,29 @@
...
+# Licensed under the MIT License
+
+"""Command-line script for redundant averaging of (calibrated) data. To calibrate and then average, try apply_cal.py"""
+
+import sys
+import argparse
+from hera_cal import io, utils, redcal
+
+# Parse arguments
+a = argparse.ArgumentParser(description="Redundantly average a data file that's already been calibrated.")
```
aewallwi, line 13: maybe consider using a non-single character filename like `a`.

aewallwi (changes requested):
Looks good to me. Minor style changes to consider.

aewallwi (approved):

scripts/red_average.py
jsdillon: Done.

scripts/red_average.py
jsdillon: Done.

PR #709, opened 2021-04-17
Delay script fix

aewallwi:
Fix new bugs in `delay_filter_run.py` associated with linear filtering (introduced in fixing CLEAN).


jsdillon (approved):
Looks good to me

PR #714, opened 2021-04-21
Flag all

aewallwi:
Add a simple scrip to flag all data in a file. This is useful if we want to use the sum to spoof a diff or the diff to spoof a sum but we want to set all data in the spoofed data to zero and all the flags to True.


scripts/flag_all.py
```diff
@@ -0,0 +1,37 @@
...
+ap.add_argument("infilename", type=str, help="path to visibility data to completely flag.")
+ap.add_argument("outfilename", type=str, help="path to new visibility file to write out completely flagged data")
+ap.add_argument("--clobber", default=False, action="store_true", help='overwrites existing file at outfile')
+ap.add_argument("--fill_data_with_zeros", default=False, action="store_true", help='Fills the data array with zeros.')
+ap.add_argument("--fill_nsamples_with_zeros", default=False, action="store_true", help='Fills the nsamples array with zeros.')
+
+args = ap.parse_args()
+
+# Load data
+hd = io.HERAData(args.infilename)
```
jsdillon, line 22: This is a case where we probably want to use pyuvdata directly instead of HERAData, since we're not doing any per-baseline manipulation.

jsdillon (approved):
One suggestion, but otherwise good to go.

PR #716, opened 2021-04-24
Throw away baselines

aewallwi:
A convenient script to throw away all flagged antennas. Adding this to the beginning of RTP stage 2 can save time and space. Depends on #714

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -19,6 +19,7 @@ import random
 import glob
 from pyuvdata.utils import POL_STR2NUM_DICT
 from . import redcal
+import argparse
 
 try:
     import aipy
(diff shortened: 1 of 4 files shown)


scripts/throw_away_flagged_antennas.py
```diff
@@ -0,0 +1,33 @@
...
+
+"""Command-line script for throw away all data with apriori flags. Saves downstream I/O, memory etc..."""
+
+import sys
+import argparse
+from hera_cal import io
+from hera_qm import utils
+
+# Parse arguments
+ap = argparse.ArgumentParser(description="Completely Flag a data file.")
```
jsdillon, line 14: Docstring is wrong.

scripts/throw_away_flagged_antennas.py
```diff
@@ -0,0 +1,33 @@
...
+ap.add_argument("infilename", type=str, help="path to visibility data to completely flag.")
+ap.add_argument("outfilename", type=str, help="path to new visibility file to write out completely flagged data")
+ap.add_argument("--yaml_file", default=False, action="store_true", help='overwrites existing file at outfile')
```
jsdillon, line 17: Docstring is wrong

jsdillon (changes requested):
Couple of docstring problems.

Also, do you want this script to also throw away any baselines that are completely flagged already? I feel like that would be a nice, useful feature.

scripts/throw_away_flagged_antennas.py
aewallwi: fixed.

scripts/throw_away_flagged_antennas.py
aewallwi: fixed.

aewallwi:
> Couple of docstring problems.
> 
> Also, do you want this script to also throw away any baselines that are completely flagged already? I feel like that would be a nice, useful feature.

I agree. 

This was a bit more then could be put in a shallow script so I've moved the baseline throwing away function into `io` with some unittests.

hera_cal/io.py
```diff
@@ -1829,3 +1830,63 @@ def baselines_from_filelist_position(filename, filelist):
...
+    if throw_away_fully_flagged_data_baselines:
+        antpairs_to_keep = []
+        for antpair in hd.get_antpairs():
+            fully_flagged = True
+            for pol in hd.pols:
+                fully_flagged = fully_flagged & np.all(hd.get_flags(antpair + (pol, )))
+            if not fully_flagged:
+                antpairs_to_keep.append(antpair)
+        hd.select(bls=antpairs_to_keep)
+    hd.write_uvh5(outfilename, clobber=clobber)
```
jsdillon, line 1880: Can I suggest adding something to the history explaining which baselines were thrown out?

hera_cal/io.py
```diff
@@ -1829,3 +1830,63 @@ def baselines_from_filelist_position(filename, filelist):
...
+
+
+def throw_away_flagged_ants_parser():
+    # Parse arguments
+    ap = argparse.ArgumentParser(description="Throw away baselines whose antennas are flagged in a yaml file or which have all integrations/chans flagged.")
+    ap.add_argument("infilename", type=str, help="path to visibility data to completely flag.")
```
jsdillon, line 1886: Wrong docstirng

hera_cal/io.py
```diff
@@ -1829,3 +1830,63 @@ def baselines_from_filelist_position(filename, filelist):
...
+    ap.add_argument("outfilename", type=str, help="path to new visibility file to write out completely flagged data.")
```
jsdillon, line 1887: Wrong docstring

hera_cal/io.py
```diff
@@ -1829,3 +1830,63 @@ def baselines_from_filelist_position(filename, filelist):
...
+            if True, also throw away baselines where all data is flagged.
+            Warning: Don't use this for files with a small number of time integrations.
+                     since this can easily happen by chance in files like this.
+            Default is False
+        clobber: bool, optional
+            overwrite output file if it already exists.
+            Default is False.
+    Returns
+    -------
+        N/A
```
jsdillon, line 1857: I'd suggest returning `hd`

jsdillon (changes requested):
Looks good. Got a few suggestions/comments

hera_cal/io.py
aewallwi: fixed.

hera_cal/io.py
aewallwi: fixed.

hera_cal/io.py
aewallwi: added history string.

hera_cal/io.py
aewallwi: I've added this.

jsdillon (approved):

jsdillon (approved):
Approved once you fix the test passing issue, which I think is just `from . import version` rather than `import version`.

PR #717, opened 2021-04-28
Stop `red_average` from dropping redundant baseline-pairs that are conjugates of the baseline-pairs in the baseline_array.

aewallwi:
As written, `red_average` excludes redundant baselines whose pair ordering is opposite of what appears in the `uvdata.baseline_array` accessed by `uvdata.get_blpairs()`. This PR fixes this.  Also fixes incompatibility with single-integration data discovered while writing unittests for the conjugation issue.


hera_cal/tests/test_utils.py
```diff
@@ -648,6 +649,40 @@ def test_red_average():
     pytest.raises(ValueError, utils.red_average, 'foo')
 
 
+def test_red_average_conjugate_baseline_case():
+    # this test covers the case where there baselines that are redundant
+    # sans conjugation.
+    to_test = []
+    for filenum in range(2):
```
jsdillon, line 656: Is this supposed to be range(3)?

jsdillon (changes requested):
Looks good. One minor error, I think.

hera_cal/tests/test_utils.py
aewallwi: Yep.

jsdillon (approved):

PR #718, opened 2021-04-29
Propagate vis units

aewallwi:
propagate visibility units of model to calibration file in `post_redcal_abscal_run`. Also make sure `gain_scale` is propagated to calibrated data in `apply_cal`. `vis_units` arg in `apply_cal` still overwrites `gain_scale`.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -163,13 +163,13 @@ def abs_amp_lincal(model, data, wgts=None, verbose=True, return_gains=False, gai
     return_gains : boolean. If True, convert result into a dictionary of gain waterfalls.
 
     gain_ants : list of ant-pol tuples for return_gains dictionary
-    
-    conv_crit : A convergence criterion below which to stop iterating LinProductSolver. 
-                Converegence is measured L2-norm of the change in the solution of the 
+
+    conv_crit : A convergence criterion below which to stop iterating LinProductSolver.
+                Converegence is measured L2-norm of the change in the solution of the
                 variables divided by the L2-norm of the solution itself.
                 Default: None (resolves to machine precision for inferred dtype).
                 Note: only used when data and model include cross-polarized visibilities.
-    
+
     maxiter : Integer maximum number of iterations to perform LinProductSolver.
               Note: only used when data and model include cross-polarized visibilities.
 
@@ -195,7 +195,7 @@ def abs_amp_lincal(model, data, wgts=None, verbose=True, return_gains=False, gai
         if ant0[1] != ant1[1]:
             cross_pols_used = True
             break
-       
+
     # make weights if None
     if wgts is None:
         wgts = odict()
@@ -225,11 +225,11 @@ def abs_amp_lincal(model, data, wgts=None, verbose=True, return_gains=False, gai
         pol0, pol1 = split_pol(k[-1])
         if cross_pols_used:
             re_eq_str = f'model_re_{i}*A_{pol0}*A_{pol1}'
-            im_eq_str = f'model_im_{i}*A_{pol0}*A_{pol1}'   
+            im_eq_str = f'model_im_{i}*A_{pol0}*A_{pol1}'
         else:
             re_eq_str = f'model_re_{i}*Asq_{pol0}'
             im_eq_str = f'model_im_{i}*Asq_{pol0}'
-            
+
         ls_data[re_eq_str] = np.real(data_here[k])
         ls_wgts[re_eq_str] = wgts[k]
         ls_consts[f'model_re_{i}'] = np.real(model_here[k])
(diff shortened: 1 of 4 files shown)


hera_cal/abscal.py
```diff
@@ -3566,6 +3566,7 @@ def post_redcal_abscal_run(data_file, redcal_file, model_files, raw_auto_file=No
         echo("The following model files overlap with data files in LST:\n" + "\n".join(matched_model_files), verbose=verbose)
         hd = io.HERAData(data_file)
         hdm = io.HERAData(matched_model_files)
+        hc.gain_scale = hdm.vis_units  # set vis_units of hera_cal based on model files.
```
jsdillon, line 3569: I'd suggest that if this changes the units on hc, a warning is issued. This seems like the right thing to do, but I don't think it should be silent.

hera_cal/apply_cal.py
```diff
@@ -383,10 +386,20 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
                 # update redundant data. Don't partial write.
                 hd_red.update(nsamples=data_nsamples, flags=data_flags, data=data)
             else:
-                # partial write works for no redundant averaging.
-                hd.partial_write(data_outfilename, inplace=True, clobber=clobber, add_to_history=add_to_history, **kwargs)
+                if vis_units is None:
+                    if hasattr(hc, 'gain_scale') and hc.gain_scale is not None:
+                        vis_units = hc.gain_scale
```
jsdillon, line 391: Again, I'd suggest issuing a warning if this changes the units.

hera_cal/apply_cal.py
```diff
@@ -383,10 +386,20 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+                    else:
+                        vis_units = hd.vis_units
+                    # partial write works for no redundant averaging.
+                hd.partial_write(data_outfilename, inplace=True, clobber=clobber, add_to_history=add_to_history, vis_units=vis_units, **kwargs)
+
         if redundant_average:
             # if we did redundant averaging, just write the redundant dataset out in the end at once.
+            if hasattr(hc, 'gain_scale') and hc.gain_scale is not None:
+                hd_red.vis_units = hc.gain_scale
```
jsdillon, line 400: And again here.

hera_cal/apply_cal.py
```diff
@@ -416,8 +429,15 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
             calibrate_in_place(data, new_gains, data_flags=data_flags, cal_flags=new_flags,
                                old_gains=old_gains, gain_convention=gain_convention)
         if not redundant_average:
-            io.update_vis(data_infilename, data_outfilename, filetype_in=filetype_in, filetype_out=filetype_out,
-                          data=data, flags=data_flags, add_to_history=add_to_history, clobber=clobber, **kwargs)
+            if vis_units is None:
+                if hasattr(hc, 'gain_scale') and hc.gain_scale is not None:
+                    vis_units = hc.gain_scale
```
jsdillon, line 434: And here

hera_cal/abscal.py
jsdillon: Specifically, I think the warning should say what the units were and what they are getting set to.

jsdillon (changes requested):
One suggestion: if the units are getting changes it should happen more verbosely. Otherwise, looks great!

hera_cal/abscal.py
aewallwi: Should I even issue a warning if the units were 'UNCALIB'?

hera_cal/apply_cal.py
aewallwi: Let me know if it makes sense to not issue a warning when `vis_units="UNCALIB"`. This is the default calibration scenario so it would seem strange issue a warning if changing units from `"UNCALIB"`.

hera_cal/apply_cal.py
aewallwi: see above.

hera_cal/apply_cal.py
aewallwi: Also see above.

hera_cal/apply_cal.py
jsdillon: I'm fine not issuing a warning in that case.

hera_cal/abscal.py
jsdillon: I'd issue a warning in any case where the units change and the original units weren't UNCALIB

hera_cal/apply_cal.py
```diff
@@ -416,8 +429,15 @@ def apply_cal(data_infilename, data_outfilename, new_calibration, old_calibratio
...
+            if vis_units is not None:
+                io.update_vis(data_infilename, data_outfilename, filetype_in=filetype_in, filetype_out=filetype_out,
+                              data=data, flags=data_flags, add_to_history=add_to_history, clobber=clobber, **kwargs, vis_units=vis_units)
+            else:
```
nkern, line 438: I think you can just add vis_units to kwargs, rather than making this if else

aewallwi: good catch. Will do!

hera_cal/abscal.py
aewallwi: I've added a warning.

hera_cal/apply_cal.py
aewallwi: I've added this when hd.vis_units is not None and hd.vis_units != 'uncalib'

hera_cal/apply_cal.py
aewallwi: added.

hera_cal/apply_cal.py
aewallwi: changed this.

hera_cal/apply_cal.py
aewallwi: added warning.

aewallwi:
Thanks for the feedback @jsdillon  and @nkern . This is ready for another look whenever you have a chance.

jsdillon (approved):
Looks good to me!

PR #721, opened 2021-05-07
Fixes to handling of JD to LST conversion

### LINKED ISSUE #720: Problems with LST-Binning metadata, especially LST2JD
I've noticed a few problems with the LST-binned files I created for H1C IDR3.2.

~1) This is trivial, but the `telescope_location` is not set. This issues a warning `telescope_location is not set. Using known values for HERA.` which is probably fine for now but should be fixed. Same problem for `antenna_positions`.~

EDIT: this turns out to be an issue with the calfits files created by reflection calibration. I'm making a new issue for this: #722

The next ones appear to be problems with LST2JD.

2) Every file I load with pyuvdata gives me a warning like `LST values stored in /lustre/aoc/projects/hera/H1C_IDR3/IDR3_2/LSTBIN/all_epochs/preprocess/zen.grp1.of1.LST.5.46255.sum.P.uvh5 are not self-consistent with time_array and telescope location. Consider recomputing with utils.get_lst_for_time.` Of course, I don't want to do this because the LST is right and the JD is wrong.

3) When a file straddles the 2 pi boundary in LST, there's a slight discontinuity in JD. This later breaks reflection calibration, since calfits objects with unevenly spaced.

image

4) When a start_jd is provided to LST2JD, but the file would straddle a JD boundary, it wraps rather than going to the next day. This creates discontinuities in time, which again UVCal doesn't like.

image

jsdillon:
This PR does a few things:

- Removes longitude independent of telescope_location in`write_vis`
- Replaces to the guts of JD2LST to wrap around pyuvdata's equivalent function
- Updates the algorithm for computing JDs from LSTs to increase accuracy by roughly 1 OOM with out an appreciable impact on speed. (This will hopefully fix some pyuvdata warnings about JDs and LSTs not matching.)
- Enables LST2JD to come up with LSTs that span multiple days to avoid discontinuities in LST-binned data
- Enables the user to specify the lst_branch_cut, which is essentially the starting LST of the data set. Values below the branch cut to map to later LSTs than values above the branch cut.
- Uses more precise locations for HERA as defaults
- Adds new tests for LST2JD

This PR resolves #720.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -1312,6 +1312,11 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
     Npols = len(pols)
     polarization_array = np.array(list(map(lambda p: polstr2num(p, x_orientation=x_orientation), pols)))
 
+    # get telescope ants
+    antenna_numbers = np.unique(list(antpos.keys()))
+    Nants_telescope = len(antenna_numbers)
+    antenna_names = list(map(lambda a: "HH{}".format(a), antenna_numbers))
+
     # get antenna positions in ITRF frame
     tel_lat_lon_alt = uvutils.LatLonAlt_from_XYZ(telescope_location)
     antenna_positions = np.array(list(map(lambda k: antpos[k], antenna_numbers)))
@@ -1376,11 +1381,6 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
     # get antennas in data
     data_ants = np.unique(np.concatenate([ant_1_array, ant_2_array]))
     Nants_data = len(data_ants)
-
-    # get telescope ants
-    antenna_numbers = np.unique(list(antpos.keys()))
-    Nants_telescope = len(antenna_numbers)
-    antenna_names = list(map(lambda a: "HH{}".format(a), antenna_numbers))
 
     # set uvw assuming drift phase i.e. phase center is zenith
     uvw_array = np.array([antpos[k[1]] - antpos[k[0]] for k in zip(ant_1_array, ant_2_array)])
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -1321,7 +1321,7 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
     if time_array is None:
         if start_jd is None:
             raise AttributeError("if time_array is not fed, start_jd must be fed")
-        time_array = LST2JD(lst_array, start_jd, longitude=(tel_lat_lon_alt[1] *180 / np.pi))
+        time_array = LST2JD(lst_array, start_jd, allow_other_jd=True, longitude=(tel_lat_lon_alt[1] *180 / np.pi))
     Ntimes = len(time_array)
 
     # get freqs
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -1326,7 +1326,10 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
     if time_array is None:
         if start_jd is None:
             raise AttributeError("if time_array is not fed, start_jd must be fed")
-        time_array = LST2JD(lst_array, start_jd, allow_other_jd=True, longitude=(tel_lat_lon_alt[1] * 180 / np.pi))
+        time_array = LST2JD(lst_array, start_jd, allow_other_jd=True, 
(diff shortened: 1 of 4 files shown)


hera_cal/utils.py
```diff
@@ -497,80 +502,78 @@ def JD2LST(JD, longitude=21.42830):
...
     at the 'start_JD' anchor point.
 
     Input:
     ------
-    LST : type=float, local apparent sidereal time [radians]
+    LST : type=float or array-like, local apparent sidereal time [radians]
 
     start_jd : type=int, integer julian day to use as starting point for LST2JD conversion
+    
+    allow_other_jd : treat the LST at the very beginning of the start_jd as the phase wrap
```
nkern, line 526: bool type

nkern: and I don't really understand what this is saying--is there a more straightforward way to explaining it? :)
what do you mean as 'phase wrap reference'. If LST is at the beginning, then it wouldn't wrap to previous JDs, only future JDs, no?

jsdillon: I think I've made this clearer now.

hera_cal/io.py
```diff
@@ -1278,10 +1278,11 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
 
     return_uvd : type=boolean, if True return UVData instance.
 
-    longitude : type=float, longitude of observer in degrees East
-
     start_jd : type=float, starting integer Julian Date of time_array if time_array is None.
 
+    lst_branch_cut : type=float, LST of data start, ensures that LSTs lower than this are wrapped around
```
nkern, line 1283: can you specify radians

hera_cal/utils.py
```diff
@@ -497,80 +502,94 @@ def JD2LST(JD, longitude=21.42830):
...
     ------
-    LST : type=float, local apparent sidereal time [radians]
+    LST : type=float or array-like, local apparent sidereal time [radians]
 
     start_jd : type=int, integer julian day to use as starting point for LST2JD conversion
+    
+    allow_other_jd : type=bool, ensure that the lst_branch_cut falls on start_jd but allow 
+                     LSTs to correspond to previous or subsequent days if necessary
+        
+    lst_branch_cut : type=float, LST that must fall during start_jd even when allow_other_jd
```
nkern, line 529: specify radians

hera_cal/utils.py
```diff
@@ -497,80 +502,94 @@ def JD2LST(JD, longitude=21.42830):
...
 
-    # get start_JD
-    base_jd = float(start_jd)
-
-    # iterate over LST
-    jd_array = []
-    for lst in LST:
+    # increase LSTs so no values fall below the branch cut (avoiding wraps in interpolation)
+    while np.any(LST < lst_branch_cut):
+        LST[LST < lst_branch_cut] += 2 * np.pi
```
nkern, line 553: are you sure this doesn't affect the input LST array? Perhaps we should make a copy of it at the top just in case

nkern: also I don't think a while is needed

hera_cal/utils.py
```diff
@@ -497,80 +502,94 @@ def JD2LST(JD, longitude=21.42830):
...
-            JD = (lst - offset) / slope
-
-            # redo if JD isn't on starting JD
-            if JD - base_jd < 0:
-                start_jd += 1
-            elif JD - base_jd > 1:
-                start_jd -= 1
+            lt_indices = np.floor(jd_array) < np.floor(start_jd)
+            gt_indices = np.floor(jd_array) > np.floor(start_jd)
+            if np.any(lt_indices):
```
nkern, line 577: can you add a single comment before each conditional briefly explaining what its catching, thanks

hera_cal/utils.py
jsdillon: Yes, because of the line `LST = np.array(LST)` or `LST = np.array([LST])` which was `copy=True` by default.

hera_cal/utils.py
jsdillon: Done.

nkern (approved):
looks good-can you explain why codecov is failing?

jsdillon:
Yeah, there's a couple of uncovered lines that involve decrementing days / 2pi radians to make sure everything lines up. These are completely symmetric with lines that handle incrementing by a day / 2pi radians and are covered, so I didn't bother adding more tests.

PR #723, opened 2021-05-10
Script for selecting multiple spws

aewallwi:
support for selecting out multiple spw ranges in apply_cal. This will allow us to process multiple disjoint spw-ranges in postprocessing pipe and use a single toml.


setup.py
```diff
@@ -43,7 +43,7 @@ def package_files(package_dir, subdirectory):
                 'scripts/query_ex_ants.py', 'scripts/red_average.py',
                 'scripts/xtalk_filter_run.py', 'scripts/time_average.py',
                 'scripts/time_chunk_from_baseline_chunks_run.py', 'scripts/chunk_files.py',
-                'scripts/flag_all.py', 'scripts/throw_away_flagged_antennas.py'],
+                'scripts/flag_all.py', 'scripts/throw_away_flagged_antennas.py', 'scripts/select_spw_ranges_run.py'],
```
jsdillon, line 46: I'd probably remove the word `_run` since that's most there to distinguish this from a module filename and that's not a problem here.

hera_cal/tests/test_utils.py
```diff
@@ -765,3 +765,57 @@ def test_chunck_baselines_by_redundant_group():
...
+    args = ap.parse_args()
+    assert args.spw_ranges == [(0, 20), (30, 100), (120, 150)]
+    # test tilde formatting.
+    sys.argv = [sys.argv[0], 'a', 'b', '--clobber', '--spw_ranges', '0~20,30~100,120~150']
+    ap = utils.select_spw_ranges_argparser()
+    args = ap.parse_args()
+    assert args.spw_ranges == [(0, 20), (30, 100), (120, 150)]
+
+
+def test_select_spw_ranges_run_script_code(tmpdir):
```
jsdillon, line 800: This is a great idea. We should do more of this!

jsdillon (approved):
Looks great! One tiny suggestion which you can take or leave.

setup.py
aewallwi: I've removed `_run`.

hera_cal/tests/test_utils.py
aewallwi: Definitely!

PR #726, opened 2021-05-24
Allow weighting by frequency-averaged samples in timeavg_waterfall

jsdillon:
Currently, timeavg_waterfall allows averaging together visibilities with either uniform weighting or nsamples weighting. The former is suboptimal (in terms of noise) while the latter risks adding spectral structure.

This PR introduces an intermediate option, `wgt_by_favg_nsample`. When averaging together integrations, this allows one to weight each integration by the average nsamples in that integration over all frequency channels. While still statistically suboptimal, this is closer to optimal while

While I don't expect this to make a big difference, I calculated the expected variance of waterfall with both uniform and `wgt_by_favg_nsample`time-averaging of a baseline in H1C IDR3.  This was done with 214 second averages. Here is the ratio:

image

Much of the time, it doesn't matter. However, there are a non-trivial fraction of times where doing this decreases the variance by 50% or more.


aewallwi (changes requested):
Looks good! One feature I'd like to see is propagating the `wgt_by_favg_nsample` arg from the bash script interface.

aewallwi (approved):
Thanks! Approved, pending tests passing ;)

PR #727, opened 2021-05-25
GitHub ci update

mkolopanis:
some general updates to the github action ci yaml.

Uses an action to setup the conda env rather than the old clunky bash script.
~~makes an empty directory to execute the tests in, this actually tests the installed version of hera_cal and breaks a degeneracy in the namespace between installed and cloned repo.~~ this was getting too complicated because of how test_utils is set up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,10 +54,7 @@ jobs:
 
       - name: Run Tests
         run: |
-          mkdir -p empty
-          cd empty
-          pytest --pyargs hera_cal --cov=hera_cal --cov-config=../.coveragerc --cov-report xml:../coverage.xml
-          cd ..
+          pytest --pyargs hera_cal --cov=hera_cal --cov-config=./.coveragerc --cov-report xml:./coverage.xml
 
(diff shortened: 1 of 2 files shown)


mkolopanis:
@jsdillon @aewallwi  made some tidy up changes for your github actions yaml if you'd like to take a look. There's an action specifically to download and install via mini-conda that I have found quite convenient in other projects.

aewallwi (changes requested):
Thanks for streamlinging things @mkolopanis ! Would it be possible to include a link to documentation of the features being used in the conversation or inline comments? Other then that, I think this basically looks good to go.

mkolopanis:
I think I'm a little confused by the request here. Are you wanting a link to https://github.com/marketplace/actions/setup-miniconda in the yaml?

aewallwi:
> I think I'm a little confused by the request here. Are you wanting a link to https://github.com/marketplace/actions/setup-miniconda in the yaml?

I think just pasting it in the conversation is sufficient! Thanks. I've approved.

aewallwi (approved):

PR #728, opened 2021-05-25
Allow abs_amp_lincal to be turned on or off

jsdillon:
This is largely for the purposes of validation and lets us selectively turn on or off the final step of abscal that is meant to remove the abscal bias.


aewallwi (approved):
looks good to me.

PR #730, opened 2021-06-16
Add fix for UVData.filename attribute

plaplant:
An upcoming PR on pyuvdata (https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1015) is adding a `filename` attribute to UVData objects, similar to the `filepaths` of HERAData objects. Because this attribute will be different when two objects are read from different files, it caused one of the tests to break. This PR fixes the broken test in a backwards-compatible way such that the test should pass both before and after the relevant PR is merged on pyuvdata.


jsdillon (approved):
Looks good to me!

PR #732, opened 2021-06-25
Update test_vis_clean.py after reorganization of hera_sim

jsdillon:
hera_sm change the interface to beam area interpolation and to `noise. sky_noise_jy`. This PR fixes a test to use the new interface. 

@nkern I'm also tagging you because I had to turn down the atol on one test to get it to work properly... for reasons I don't understand `assert not np.isclose(np.mean(np.abs(d1)[select]), np.mean(np.abs(d2)[select]), atol=10)` wasn't working. Is that unexepected?

### DIFF
--- a/hera_cal/tests/test_vis_clean.py
+++ b/hera_cal/tests/test_vis_clean.py
@@ -10,6 +10,8 @@ import sys
 import shutil
 from scipy import constants, interpolate
 from pyuvdata import UVCal, UVData
+from hera_sim.interpolators import Beam
+from hera_sim import DATA_PATH as HS_DATA_PATH
 from hera_sim import noise
 from uvtools import dspec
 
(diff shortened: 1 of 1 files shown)


r-pascua (approved):
The `hera_sim`-related changes look fine to me. I take it the commented out line (749 in `test_vis_clean.py`) is just there as a reference?

hera_cal/tests/test_vis_clean.py
```diff
@@ -742,8 +744,10 @@ def test_trim_model(self):
         # delay width = intended delay width x (manually set dnu / original dnu of the attached data)
         np.random.seed(0)
         k = (23, 24, 'ee')
-        Op = noise.bm_poly_to_omega_p(V.freqs / 1e9)
-        V.data[k] += noise.sky_noise_jy(V.data[(23, 23, 'ee')], V.freqs / 1e9, V.lsts, Op, inttime=50)
+        beam_interp = Beam(HS_DATA_PATH / 'HERA_H1C_BEAM_POLY.npy')
+        Op = beam_interp(V.freqs / 1e9)
+        # V.data[k] += noise.sky_noise_jy(autovis=V.data[(23, 23, 'ee')], freqs=V.freqs / 1e9, lsts=V.lsts, omega_p=Op)
```
nkern, line 749: can you rm this comment, unless its pertinent

nkern:
I don't see why turning *down* the atol would make the test pass?

nkern:
@jsdillon yes I'm fine w/ your change to the unit-test--that unit-test needs to be overhauled anyways b/c its not really testing what it should be

PR #734, opened 2021-08-04
Add DataContainer to data_wgts

kian466:
https://github.com/HERA-Team/hera_cal/blob/e7b4c12eeb3be6eb7c160b10d0d1d75b235bab7a/hera_cal/redcal.py#L1423

This line with give an error if the ordering of the antenna pairs in `data_wgts` is different from the `model`. This was fixed by initializing the `data_wgts` as a `DataContainer` object.


aewallwi (approved):
Looks good to me (pending test passes).

PR #736, opened 2021-09-04
Select main lobe 🔭 🛰️ 

### LINKED ISSUE #737: Select fringe-rates in the main beam lobe
A simple tophat fringe-rate filter that extracts the fringe-rates from the main-lobe.

aewallwi:
Filter the fringe-rates from the main lobe using instantaneous fringe rates from https://ui.adsabs.harvard.edu/abs/2016ApJ...820...51P/abstract . Depends on #700 and #747, #757 #759. Closes #737.


scripts/tophat_frfilter_run.py
```diff
@@ -60,6 +59,7 @@
                                    overwrite_flags=ap.overwrite_flags, skip_autos=ap.skip_autos,
                                    skip_if_flag_within_edge_distance=ap.skip_if_flag_within_edge_distance,
                                    zeropad=ap.zeropad, tol=ap.tol, skip_wgt=ap.skip_wgt, max_frate_coeffs=ap.max_frate_coeffs,
-                                   frate_width_multiplier=ap.frate_width_multiplier, frate_standoff=ap.frate_standoff,
-                                   min_frate_half_width=ap.min_frate_half_width,
-                                   clean_flags_in_resid_flags=True, **filter_kwargs)
+                                   frate_width_multiplier=ap.frate_width_multiplier, frate_standoff=ap.frate_standoff, fr_freq_skip=ap.fr_freq_skip,
+                                   min_frate_half_width=ap.min_frate_half_width, uvbeam=ap.uvbeam, percentile_low=ap.percentile_low, percentile_high=ap.percentile_high
+                                   clean_flags_in_resid_flags=True, pre_filter_modes_between_lobe_minimum_and_zero=ap.pre_filter_modes_between_lobe_minimum_and_zero,
```
hughbg, line 64: There's a comma missing after percentile_high=ap.percentile_high

hughbg:
I've run the code without any problems, and it runs as expected (although my definition of "expected" is based on limited experience). I also ran the test scripts "by hand" and there was 1 failure which I assume is due to my environment. I'll list it just for the record:  in test_select_spw_ranges_run_script_code the message was "OSError: Cannot find file /tmp/pytest-of-hgarsden/pytest-2/test_select_spw_ranges_run_scr0/test_calibrated_output.uvh5".

nkern:
the data files seem to be non-negligibly small. in hera_pspec we solve this by generating unittest data dynamically with a set function. can this be done here?

aewallwi:
> the data files seem to be non-negligibly small. in hera_pspec we solve this by generating unittest data dynamically with a set function. can this be done here?

I'm a little bit cautious about putting `hera_sim` into the dependencies for the unittests. I think the best approach might be to run the tests on lighter files.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+    antnums = uvd.antenna_numbers  # antenna numbers.
+
+    lat, lon, alt = uvd.telescope_location_lat_lon_alt_degrees
+    location = EarthLocation(lon=lon * units.deg, lat=lat * units.deg, height=alt * units.m)
+
+    # get topocentricl AzEl Beam coordinates.
+    npix = uvb.data_array.shape[-1]
+    nside = hp.npix2nside(npix)
+    polar, az = hp.pix2ang(nside=nside, ipix=range(npix))
+    # cancel out super-horizon beam
```
jsdillon, line 174: What does this comment have to do with the next line?

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+    # generate fringe_rate grid in mHz.
+    dt = SDAY_KSEC * np.median(np.diff(np.unique(uvd.time_array)))  # times in kSec
+    if nfr is None:
+        nfr = uvd.Ntimes
+    if dfr is None:
+        # if no dfr provided, set to 1 / (ntimes * dt)
+        dfr = 1. / (dt * nfr)
+
+    # build grid.
+    fr_grid = np.arange(-nfr // 2, nfr // 2) * dfr
```
jsdillon, line 197: I'm not sure if this is guaranteed to be length nfr (when things are perfectly divisible, for example), but it would be if you use linspace.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+        if True, set profiles for all pols equal to the sum of the profiles of individual pols.
+    dfr: float, optional.
+        spacing of fringe-rate grid used to perform binning and percentile calc
+        in units of mHz.
+        default is None -> set to 1 / (dtime * Ntime) of uvd.
+    nfr: float, optional.
+        number of points on fringe-rate grid to perform binning and percentile calc.
+        default is None -> set to uvd.Ntimes.
+    taper: str, optional
+        taper expected for power spectrum calculations. Fringe-rates from different frequencies
```
jsdillon, line 135: Note where one can find the options for this parameter.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+
+    profiles = {}  # store profiles here.
+    ap_blkeys = {}
+    # keeps track of different polarizations for each antenna pair
+    # for if we are going to sum over polarizations.
+    # get redundancies (will only compute fr-profile once for each red group).
+    antpos, antnums = uvd.get_ENU_antpos()
+    antpos = {an: ap for an, ap in zip(antnums, antpos)}
+    reds = redcal.get_reds(antpos, pols=list(unique_pols), include_autos=True)
+    reds = [[bl for bl in rg if bl in keys or utils.reverse_bl(bl) in keys] for rg in reds]
```
jsdillon, line 217: I believe you can do these two lines with `redcal.filter_reds`

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+    reds = [rg for rg in reds if len(rg) > 0]
+
+    for redgrp in reds:
+        # only explicitly calculate fr profile for the first vis in each redgroup.
+        bl = redgrp[0]
+        echo("Generating FR-Profile of {} at {}".format(bl, str(datetime.datetime.now())), verbose=verbose)
```
jsdillon, line 223: I personally find these easier to read with f-strings, but it's your call.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
     return frate_centers, frate_half_widths
 
 
+def build_fringe_rate_profiles(uvd, uvb, keys=None, normed=True, combine_pols=True, nfr=None, dfr=None,
+                               taper='none', fr_freq_skip=1, verbose=False):
+    """
+    Calculate fringe-rate profiles to either directly apply as an FIR filter or set a range to filter.
```
jsdillon, line 112: I'd clarify what you mean by FR profile and how you integrate over frequency.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+        ind1 = np.where(antnums == bl[0])[0][0]
+        ind2 = np.where(antnums == bl[1])[0][0]
+        blvec = uvd.antenna_positions[ind2] - uvd.antenna_positions[ind1]
+        # initialize binned power.
+        # we will bin frate power together for all frequencies, weighted by taper.
+        binned_power = np.zeros_like(fr_grid)
+        binned_power_conj = np.zeros_like(fr_grid)
+        # iterate over each frequency and ftaper weighting.
+        # use linspace to make sure we get first and last frequencies.
+        chans_to_use = np.linspace(0, uvd.Nfreqs - 1, int(uvd.Nfreqs / fr_freq_skip)).astype(int)
```
jsdillon, line 237: Should the choice of chans_to_use be aware of the flags in `uvd`?

hera_cal/frf.py
jsdillon: Also, I think you stopped mid-sentence.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+            frates = np.dot(np.cross(np.array([0, 0, 1.]), blvec), eq_xyz) * 2 * np.pi * f0 / SPEED_OF_LIGHT / SDAY_KSEC
+            # square of power beam values in directions of sky pixels
+            bsq = np.abs(uvb.data_array[0, 0, polnum, np.argmin(np.abs(f0 - uvb.freq_array[0])), :].squeeze()) ** 2.
+            # set beam below horizon to be zero.
+            bsq[polar >= np.pi / 2.] = 0.
+            # get fringe-rate bin membership for each pixel.
+            fr_bins = np.round(frates / dfr + nfr / 2).astype(int)
+            fr_bins_conj = np.round(frates / -dfr + nfr / 2).astype(int)
+            # bin power.
+            for binnum in range(nfr):
```
jsdillon, line 248: I imagine that a lot of the reason this is slow is because of this triple-nested for loop. Can this be vectorized?

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+                # For each bin, find all pixels that fall in that fr bin and add the sum beam-square values in each pixel
+                # times the frequency weighing value set by taper.
+                binned_power[binnum] += np.sum(bsq[fr_bins == binnum]) * fw  # add sum of beam squared times taper weight.
+                binned_power_conj[binnum] += np.sum(bsq[fr_bins_conj == binnum]) * fw
```
jsdillon, line 252: Is there a way to avoid doing both of these?

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+        list of antpairpol tuples of baselines to calculate fringe-rate limits for.
+    dfr: float, optional.
+        spacing of fringe-rate grid used to perform binning and percentile calc
+        in units of mHz.
+        default is None -> set to 1 / (dtime * Ntime) of uvd.
+    nfr: float, optional.
+        number of points on fringe-rate grid to perform binning and percentile calc.
+        default is None -> set to uvd.Ntimes.
+    taper: str, optional
+        taper expected for power spectrum calculations. Fringe-rates from different frequencies
```
jsdillon, line 312: Same notes as above

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+    ----------
+    uvd: UVData object
+        UVData holding baselines for which we will build fringe-rate profiles.
+    uvb: UVBeam object, optional
+        UVBeam object holding beams that we will build uvbeam profiles for.
+        default is None -> use pre-computed frate_profiles
+    frate_profiles: dict object, optional
+        Dictionary mapping antpairpol keys to fringe-rate profiles.
+        centered on a grid of length nfr spaced by dfr centered at 0.
+        default is None -> automatically calculate profiles using uvb.
```
jsdillon, line 295: Note that an error will be raised if neither is provided.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+    percentile_low: float, optional
+        Percent of beam-squared power below lower fringe rate.
+    keys: dict, optional
```
jsdillon, line 298: Your docstring is in a different order than the kwargs

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+        spacing of fringe-rate grid used to perform binning and percentile calc
+        in units of mHz.
+        default is None -> set to 1 / (dtime * Ntime) of uvd.
+    nfr: float, optional.
+        number of points on fringe-rate grid to perform binning and percentile calc.
+        default is None -> set to uvd.Ntimes.
+    taper: str, optional
+        taper expected for power spectrum calculations. Fringe-rates from different frequencies
+    frate_standoff: float, optional
+        Additional fringe-rate standoff in mHz to add to Omega_E b_{EW} nu/c for fringe-rate inpainting.
```
jsdillon, line 314: Inpainting? I don't think this function inpaints.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+
+    # get redundancies (will only compute fr-profile once for each red group).
+    unique_pols = set({})
+    for bl in keys:
+        if bl[-1] not in unique_pols:
+            unique_pols.add(bl[-1])
+    antpos, antnums = uvd.get_ENU_antpos()
+    antpos = {an: ap for an, ap in zip(antnums, antpos)}
+    reds = redcal.get_reds(antpos, pols=list(unique_pols), include_autos=True)
+    reds = [[bl for bl in rg if bl in keys or utils.reverse_bl(bl) in keys] for rg in reds]
```
jsdillon, line 361: Same comment as before. Also, this is duplicated code from above and should be its own function (can be private and not unittested)

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+            # save low and high fringe rates for bl and its conjugate
+            for cnum, bl in enumerate([blt, utils.reverse_bl(blt)]):
+                frate_centers[bl] = .5 * (frlows[cnum] + frhighs[cnum])
+                frate_half_widths[bl] = .5 * np.abs(frlows[cnum] - frhighs[cnum]) * frate_width_multiplier + frate_standoff
+                frate_half_widths[bl] = np.max([frate_half_widths[bl], min_frate_half_width])
+
+    return frate_centers, frate_half_widths
+
+
+def sky_mainlobe_complement(uvd, uvb, percentile_low=5., percentile_high=95., keys=None,
```
jsdillon, line 396: Uh, this function doesn't look done...

hera_cal/frf.py
```diff
@@ -575,12 +890,12 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
             filt_flags[k] = f
             filt_nsamples[k] = eff_nsamples
 
-    def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
+    def tophat_frfilter(self, keys=None, wgts=None, mode='clean', uvb=None, percentile_low=5., percentile_high=95.,
```
jsdillon, line 893: I feel like this function has two _very_ different operating modes (with and w/o the beam). It strikes me that you should write the version that uses the beam as a separate function that calls this one.

hera_cal/frf.py
```diff
@@ -642,22 +985,100 @@ def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
...
+                if not (frate_centers[k] + frate_half_widths[k] >= 0. and frate_centers[k] - frate_half_widths[k] <= 0.):
+                    self.fourier_filter(keys=[k], filter_centers=[0.], filter_half_widths=[min_frate_abs],
+                                        mode=mode, x=self.times * SDAY_KSEC,
+                                        wgts=wgts, output_prefix='pre_filter',
+                                        ax='time', cache=filter_cache, skip_wgt=skip_wgt, verbose=verbose, **filter_kwargs)
+                else:
+                    if 'data' not in filter_kwargs:
+                        self.pre_filter_resid[k] = self.data[k]
+                    else:
+                        self.pre_filter_resid[k] = filter_kwargs['data'][k]
```
jsdillon, line 1032: Several lines not covered by tests here and below.

hera_cal/frf.py
```diff
@@ -709,7 +1130,7 @@ def time_avg_data_and_write(input_data_list, output_data, t_avg, baseline_list=N
                       "in your dataset. Exiting without writing any output.", RuntimeWarning)
     else:
         fr = FRFilter(input_data_list, filetype=filetype)
-        fr.read(bls=baseline_list, axis='blt')
+        fr.read(bls=baseline_list)
```
jsdillon, line 1133: Why this change?

hera_cal/frf.py
```diff
@@ -749,6 +1170,17 @@ def tophat_frfilter_argparser(mode='clean'):
     ap.add_argument("--max_frate_coeffs", type=float, default=None, nargs=2, help="Maximum fringe-rate coefficients for the model max_frate [mHz] = x1 * EW_bl_len [ m ] + x2."
                                                                                   "Providing these overrides the sky-based fringe-rate determination! Default is None.")
     ap.add_argument("--skip_autos", default=False, action="store_true", help="Exclude autos from filtering.")
+    ap.add_argument("--uvbeam", default=None, type=str, help="Path to UVBeam beamfits file to use for determining isotropic sky fringe-rates to filter.")
```
jsdillon, line 1173: If you want to avoid adding a new script too, perhaps have the associated script call the new function only when uvbeam is provided.

hera_cal/frf.py
```diff
@@ -797,23 +1229,28 @@ def load_tophat_frfilter_and_write(datafile_list, baseline_list=None, calfile_li
...
+        verbose: bool, optional
+            Helpful text outputs.
         clean_flags_in_resid_flags: bool, optional. If true, include clean flags in residual flags that get written.
                                     default is True.
         filter_kwargs: additional keyword arguments to be passed to FRFilter.tophat_frfilter()
     '''
     if baseline_list is not None and Nbls_per_load is not None:
         raise NotImplementedError("baseline loading and partial i/o not yet implemented.")
-    hd = io.HERAData(datafile_list, filetype='uvh5', axis='blt')
+    hd = io.HERAData(datafile_list, filetype='uvh5')
```
jsdillon, line 1243: Why?

hera_cal/frf.py
```diff
@@ -797,23 +1229,28 @@ def load_tophat_frfilter_and_write(datafile_list, baseline_list=None, calfile_li
...
     if baseline_list is not None and len(baseline_list) == 0:
         warnings.warn("Length of baseline list is zero."
                       "This can happen under normal circumstances when there are more files in datafile_list then baselines."
                       "in your dataset. Exiting without writing any output.", RuntimeWarning)
     else:
         if baseline_list is None:
             if len(hd.filepaths) > 1:
-                baseline_list = list(hd.bls.values())[0]
+                baseline_list = list(hd.antpairs.values())[0]
```
jsdillon, line 1251: This conflicts with the docstring, which describes baseline_list has being a list of antpairpols

jsdillon (changes requested):

hera_cal/frf.py
r-pascua: You should be able to vectorize this with `scipy.stats.binned_statistic`. For example, I *think* this should work:
```
from scipy.stats import binned_statistic
binned_power = binned_statistic(frates, bsq * fw, statistic="sum", bins=fr_grid)[0]
binned_power_conj = binned_statistic(-frates, bsq * fw, statistic="sum", bins=fr_grid)[0]
```
If you do this (which I think is the right thing to do), then you need to make sure that you define the fringe-rate grid correctly. In particular, it specifies the *edges* of the bins, and so should have length `nfr + 1`.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+                               taper='none', fr_freq_skip=1, verbose=False):
+    """
+    Calculate fringe-rate profiles to either directly apply as an FIR filter or set a range to filter.
+
+    Parameters
+    ----------
+    uvd: UVData object
+        UVData holding baselines for which we will build fringe-rate profiles.
+    uvb: UVBeam object
+        UVBeam object holding beams that we will build uvbeam profiles for.
```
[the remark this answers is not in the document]
r-pascua: Is that necessary? It looks like the operations are done on a copy of the beam.

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+        centered on a grid of length nfr spaced by dfr centered at 0.
+        default is None -> automatically calculate profiles using uvb.
+    percentile_low: float, optional
+        Percent of beam-squared power below lower fringe rate.
+    keys: dict, optional
+        antpairpol keys to compute fringe-rate limits for.
+    percentile_high: float, optional
+        Percent of beam-squared power above upper fringe rate.
+    to_filter: list of antpairpol tuples
+        list of antpairpol tuples of baselines to calculate fringe-rate limits for.
```
r-pascua, line 303: It looks like this isn't a parameter anymore?

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+        default is 1 -> takes a long time. We recommend setting this to be larger.
+    verbose: bool, optional
+        lots of text output, default is False.
+
+    Returns
+    -------
+    frate_centers: dict object,
+        Dictionary with the center fringe-rate of each baseline in to_filter in units of mHz.
+    frate_half_widths: dict object
+        Dictionary with the half widths of each fringe-rate window around the frate_centers in units of mHz.
```
r-pascua, line 333: What are the keys of these dictionaries?

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+            raise ValueError("Must either supply uvb or fr_profiles!")
+    else:
+        if nfr is None:
+            nfr = uvd.Ntimes
+        if dfr is None:
+            dfr = 1. / (nfr * np.mean(np.diff(np.unique(uvd.time_array))) * SDAY_KSEC)
+        fr_grid = np.arange(-nfr // 2, nfr // 2) * dfr
+
+    # get redundancies (will only compute fr-profile once for each red group).
+    unique_pols = set({})
```
r-pascua, line 354: Again, you can just do `set()`

hera_cal/frf.py
```diff
@@ -97,6 +106,312 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, m
...
+    for bl in keys:
+        if bl[-1] not in unique_pols:
+            unique_pols.add(bl[-1])
+    antpos, antnums = uvd.get_ENU_antpos()
+    antpos = {an: ap for an, ap in zip(antnums, antpos)}
```
r-pascua, line 359: Same comment as before. If you want to be extra fancy, you could even do
`antpos = dict(zip(*uvd.get_ENU_antpos()[::-1]))`.

hera_cal/frf.py
```diff
@@ -10,16 +10,26 @@
     HAVE_UVTOOLS = False
 
 from . import utils
-
+import scipy.interpolate as interp
```
r-pascua, line 13: You only ever use `interp1d` from this module, so why not replace this with `from scipy.interpolate import interp1d`?

hera_cal/frf.py
```diff
@@ -10,16 +10,26 @@
...
 from .datacontainer import DataContainer
 from .vis_clean import VisClean
-from pyuvdata import UVData, UVFlag
+from pyuvdata import UVData, UVFlag, UVBeam
 import argparse
 from . import io
 from . import vis_clean
 import warnings
+from astropy.coordinates import SkyCoord, EarthLocation, AltAz, ITRS
+import astropy.units as units
```
r-pascua, line 22: You can just do `from astropy import units`

hera_cal/frf.py
```diff
@@ -10,16 +10,26 @@
...
+from pyuvdata import utils as uvutils
+from . import utils
+import copy
+from .utils import echo
+import datetime
+import astropy.constants as const
+from . import redcal
 
 SPEED_OF_LIGHT = const.c.si.value
 SDAY_KSEC = 86163.93 / 1000.
```
r-pascua, line 35: Since you've imported `astropy.units`, wouldn't it make more sense to replace this with `SDAY_KSEC = units.sday.to("ks")`? On that note, it looks like the value here doesn't exactly agree with what `astropy` says (though it's off by less than a second).

hera_cal/frf.py
r-pascua: I think I agree with Josh on this.

r-pascua (changes requested):
It looks like a fair amount of this has been fleshed out, though there are a number of changes that should be made. I'll try to be succinct:
* I think you need to be more careful with how you're setting the fringe-rate bins, and I would strongly recommend that you use `scipy.stats.binned_statistic` to do the calculation in `build_fringe_rate_profiles`. If you update this, then be sure to propagate the changes to `get_fringe_rate_limits`. In particular, I think there is an important (perhaps subtle) difference between bin centers and bin edges that needs careful attention.
* I agree with Josh that you should make a new function for beam-based tophat FR filtering rather than overload the `tophat_frfilter` function.
* Finally, I think the `sky_mainlobe_complement` function either needs to be completed or scrapped before this PR can be merged. I suppose a middle-ground is to just throw a `NotImplementedError` into the function body.

I think that about summarizes my takes on a first inspection. I made a handful of style-related comments, but tried to not nitpick. I should have more feedback when I get to validating these tools.

Thanks for the hard work! Hope the review is helpful.

hera_cal/frf.py
aewallwi: I think I've added coverage.

aewallwi:
Thanks for the reviews @hughbg @nkern @jsdillon @r-pascua.

 I think this is ready for another look!

hera_cal/chunker.py
```diff
@@ -68,7 +68,7 @@ def chunk_files(filenames, inputfile, outputfile, chunk_size, type="data",
                 polarizations = chunked_files.pols
         if spw_range is None:
             spw_range = (0, chunked_files.Nfreqs)
-        data, flags, nsamples = chunked_files.read(axis='blt', polarizations=polarizations,
+        data, flags, nsamples = chunked_files.read(polarizations=polarizations,
```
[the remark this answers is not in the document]
jsdillon: Fair, but this will make reading slower because this enables fast concatenation. Perhaps you want to expose this parameter to the user?

hera_cal/tests/test_frf.py
```diff
@@ -423,6 +432,95 @@ def test_load_tophat_frfilter_and_write_multifile(self, tmpdir):
...
+        test_data = os.path.join(DATA_PATH, "fr_unittest_data_ds.uvh5")
+        uvd = UVData()
+        uvd.read_uvh5(test_data)
+        uvb = UVBeam()
+        uvb.read_beamfits(test_beam)
+        fr_grid, profiles = frf.build_fringe_rate_profiles(uvd, uvb)
+        assert len(fr_grid) == uvd.Ntimes
+
+    def test_get_fringe_rate_limits(self):
+        # simulations constructed with the notebook at https://drive.google.com/file/d/1jPPSmL3nqQbp7tTgP77j9KC0802iWyow/view?usp=sharing
```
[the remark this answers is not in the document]
jsdillon: Uh, I still can't access it. Does it work for you in an incognito window?

hera_cal/frf.py
jsdillon: I still think 3-4 functions might be cleaner than 1 (and also more modular, since you could calculate the fringe rate limits without actually having to do the filter). That said, if you really want to leave it as one function, perhaps you can group the kwargs into sections for different operating modes? It's really hard to tell which options are relevant in which cases and which options are necessary when others are/aren't provided. 

Also, I don't get the line "if provided, will supercede main-lobe" What does that mean?

Finally, can you make the tabbing of the docstring consistent? It's very hard to read right now.

jsdillon (changes requested):
I'm mostly converged now. Just a few option issues.

nkern:
@aewallwi thanks, I would like to have a closer look at this as well, although I suspect josh and bobby got most of our bases covered. I'm aiming to look at it sometime this week

aewallwi:
> @aewallwi thanks, I would like to have a closer look at this as well, although I suspect josh and bobby got most of our bases covered. I'm aiming to look at it sometime this week

Sounds good! I've added you as a reviewer.

hera_cal/abscal.py
```diff
@@ -2020,8 +2020,9 @@ def match_times(datafile, modelfiles, filetype='uvh5', atol=1e-5):
 
     # shift model files relative to first file & first index if needed
     for ml in model_lsts:
-        if ml[0] < model_lsts[0][0]:
-            ml += 2 * np.pi
+        for i in range(len(ml)):
```
nkern, line 2023: i'm generally wary of changing functions in other parts of the codebase not directly related to the PR, as they can have unintended, reverberating impacts--can you explain the motivation here? if the issue is a LST wrap then the easier thing to do is to just ensure each ml is unwrapped, I think

hera_cal/frf.py
nkern: perhaps I'm echoing the sentiment here: we have methods for applying an FIR filter given a fringe-rate profile (FRP) in frf.py already. currently, if you provide kwargs (e.g. frate_standoff, max_frate_coeffs) then it constructs the FRP and uses the existing code to apply the filter. part of the new additions is to generate an FRP given a beam, but I think we will likely want to filter with any arbitrary FRP in the near future, not just the ones generated by the function here. I think this is possibly simpler if you just have this high-level function take existing arguments *or* an `frp` dictionary, which is user provided and supercedes the existing arguments. Your existing build_frate() function is a nice interface between `uvb` beam objects and a dictionary of `frp` profiles, so this connection can be made in the run script. I think this would help solve the issue of different operating modes, b/c now there are only two: give me an frp you've made, or have me make a simple one given your choice of kwargs.

Obviously we've got four people here with four slightly differing opinions, but I do think that maintaining this function as-written may be difficult for people other than AEW. lmk your thoughts aaron, thanks!

hera_cal/frf.py
```diff
@@ -642,22 +1001,101 @@ def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
...
+                else:
+                    if 'data' not in filter_kwargs:
+                        self.pre_filter_resid[k] = self.data[k]
+                    else:
+                        self.pre_filter_resid[k] = filter_kwargs['data'][k]
+                    if 'flags' not in filter_kwargs:
+                        self.pre_filter_resid_flags[k] = self.flags[k]
+                    else:
+                        self.pre_filter_flags[k] = filter_kwargs['flags'][k]
+                if center_before_filtering:
```
nkern, line 1054: why center before filtering? shouldn't the FRP be such that it is symmetric about the main lobe anyways?

hera_cal/frf.py
```diff
@@ -616,10 +956,29 @@ def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
...
+            default is 1 -> takes a long time. We recommend setting this to be larger.
         verbose: bool, optional, lots of outputs!
+        dfr: float, optional.
+            spacing of fringe-rate grid used to perform binning and percentile calc
+            in units of mHz.
+            default is None -> set to 1 / (dtime * Ntime) of uvd.
+        nfr: float, optional.
+            number of points on fringe-rate grid to perform binning and percentile calc.
+            default is None -> set to uvd.Ntimes.
+        pre_filter_modes_between_lobe_minimum_and_zero: bool, optional
```
nkern, line 975: if this is the case, isn't the answer to just change the FRP such that it has zero power near frate ~ 0?

aewallwi: This can be done if we are using the DFT basis to perform our fringe-rate filter. However, it will not work if we are using functions that are not localized in fringe-rate space including both DPSS modes and DFT modes with side-lobes induced by missing data and finite bandpass.

hera_cal/frf.py
aewallwi: If we want to use cleaning or DPSS fitting (to deal with missing data) which is, in some cases, only currently supports filter windows centered at zero, then we will want to center the filtering region at zero fringe-rate. In the case of DPSS fitting, non-zero window centers are supported but filters centered at zero tend to be better behaved.

hera_cal/abscal.py
aewallwi: Sorry about this, this was actually part of #757 but I wanted to include it here since I'm currently running the pipeline with this branch and need the bug fixed here too. #757 will be merged before this PR. I've added a dependency note to the PR description.

aewallwi: This is no longer a new addition with the merger of #757

hera_cal/chunker.py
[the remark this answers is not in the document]
aewallwi: That makes sense. I can expose read_kwargs.

hera_cal/tests/test_frf.py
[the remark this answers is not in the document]
aewallwi: Looks like sharing was restricted to berkeley emails 😂 I've updated this link and tested it in incognito mode. Now it does work.

aewallwi (commented):
Thanks for the ~final round of comments @jsdillon and @nkern. I think I've addressed everything now.

hera_cal/frf.py
```diff
@@ -1277,7 +1280,7 @@ def load_tophat_frfilter_and_write(datafile_list, baseline_list=None, calfile_li
         baseline_antennas = np.unique(baseline_antennas).astype(int)
         if calfile_list is not None:
             cals = io.HERACal(calfile_list)
-            cals.read(antenna_nums=baseline_antennas, frequencies=freqs)
+            cals.read(antenna_nums=baseline_antennas, frequencies=freqs, )
```
jsdillon, line 1283: I assume this is left over from something?

hera_cal/frf.py
jsdillon: Other than this thread, I'm fine with this PR. Happy to find a time to chat this afternoon to figure out a resolution to makes the three of us happy, @aewallwi @nkern

aewallwi: > perhaps I'm echoing the sentiment here: we have methods for applying an FIR filter given a fringe-rate profile (FRP) in frf.py already. currently, if you provide kwargs (e.g. frate_standoff, max_frate_coeffs) then it constructs the FRP and uses the existing code to apply the filter. part of the new additions is to generate an FRP given a beam, but I think we will likely want to filter with any arbitrary FRP in the near future, not just the ones generated by the function here. I think this is possibly simpler if you just have this high-level function take existing arguments _or_ an `frp` dictionary, which is user provided and supercedes the existing arguments. Your existing build_frate() function is a nice interface between `uvb` beam objects and a dictionary of `frp` profiles, so this connection can be made in the run script. I think this would help solve the issue of different operating modes, b/c now there are only two: give me an frp you've made, or have me make a simple one given your choice of kwargs.

@nkern I think I've addressed your concern here by providing an `frate_profiles` arg to `tophat_frfilter`. Let me know if this is good. 

Perhaps I'm not quite parsing what people are requesting here. Do people want this function to only take in frate_centers and hfrate_half_widths dicts (technically the core of what is required for rectangular fringe-rate filtering) and then have a helper function that determines these from the different possible options (uvbeams, max_frate_coeffs, frate_standoff etc..) and then have a more general fringe-rate filtering interface function that supports the tophat mode but also has hooks to support applying multiplicative (in fringe-rate) fir filters?

hera_cal/frf.py
aewallwi: yep. I've removed this.

hera_cal/frf.py
nkern: I see, so if certain modes require this than shouldn't this be hard-coded? perhaps we can for now just assume that CLEAN and DPSS only act as high-pass filters, such that this phasing should always be applied in these cases?

hera_cal/frf.py
```diff
@@ -642,30 +1008,110 @@ def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
...
             if mode != 'clean':
                 filter_kwargs['suppression_factors'] = [tol]
             else:
                 filter_kwargs['tol'] = tol
-            self.fourier_filter(keys=[k], filter_centers=[center_frates[k]], filter_half_widths=[width_frates[k]],
-                                mode=mode, x=self.times * 3.6 * 24.,
-                                data=self.data, flags=self.flags, wgts=wgts,
-                                ax='time', cache=filter_cache, skip_wgt=skip_wgt, verbose=verbose, **filter_kwargs)
+
+            if pre_filter_modes_between_lobe_minimum_and_zero:
```
nkern, line 1044: this seems like something that should really be dealt with as a separate call to this function, or is there some benefit (other than convenience) that this offers?

hera_cal/frf.py
nkern: Yeah, this function just has a bunch of different functionalities in its kwargs its kind of hard to use. Perhaps as you suggest it would be easier for this function to just take frate_centers and frate_half_widths and other kwargs.

Then we have a separate function that maps the variety of kwargs here to the frate_centers and frate_half_widths, which would be a massive multi-pronged function but at least we are upfront that this function is complex, and are making the end-user interface (tophat_frfilter) a little cleaner. effectively, revert this function almost back to its state on master (except for a few added features like center_frates) and push all the new guts to a new "usecase_determination_function". here is an example of what I mean

```
def tophat_frfilter(self, keys=None, ..., frate_centers=None, frate_half_widths=None, ...., **filter_kwargs):
    ...


def select_tophat_frates(case, ...):
    """
    Select fringe-rates to filter given a filtering case

    Parameters
    ------------
    case : str
        Filtering case ['frate_coeffs', 'uvbeam', 'fr_profiles']
    """
    if case == 'frate_coeffs':
        if max_frate_coeffs is None:
            center_frates, width_frates = sky_frates(uvd=self.hd, keys=keys, frate_standoff=frate_standoff,
                                                     frate_width_multiplier=frate_width_multiplier, min_frate_half_width=min_frate_half_width)
        else:
            width_frates = {k: np.max([max_frate_coeffs[0] * self.blvecs[k[:2]][0] + max_frate_coeffs[1], 0.0]) for k in keys}
            center_frates = {k: 0.0 for k in keys}

    elif case == 'uvbeam':
        ...

    elif case == 'fr_profiles':
        ...

    else:
        raise ValueError("did not recognize {}".format(case))
    
    return fc, fhw

# for xtalk filter mode as we've done previously we'd do
fc, fhw = select_tophat_frates('frate_coeffs', max_frate_coeffs=[0.01, 0.2], ...)

# for UVBeam inspired, we'd do
fc, fhw = select_tophat_frates('uvbeam', uvb=uvb, percentile_low=0.05, percentile_hi=0.95)

# etc..
```

lmk your thoughts

hera_cal/frf.py
aewallwi: I think it would make sense if the fringe rate limit containers were computed outside of this function. Perhaps this is another reason to build another layer of driver methods.

hera_cal/frf.py
aewallwi: I'm hardcoding this.

hera_cal/frf.py
```diff
@@ -648,35 +1075,200 @@ def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
...
+    def split_frates_and_tophat_frfilter(self, case, keys=None, wgts=None, mode='clean', uvb=None, percentile_low=5., percentile_high=95.,
+                                         frate_standoff=0.0, frate_width_multiplier=1.0, min_frate_half_width=0.025,
+                                         max_frate_coeffs=None, skip_wgt=0.1, tol=1e-9, cache_dir=None, read_cache=False,
+                                         write_cache=False, center_before_filtering=True, fr_freq_skip=1,
+                                         verbose=False, nfr=None, dfr=None, pre_filter_modes_between_lobe_minimum_and_zero=False,
+                                         frate_profiles=None, **filter_kwargs):
+        '''
+        A wrapper around VisClean.fourier_filter specifically for
+        filtering along the time axis with uniform fringe-rate weighting.
+        Parameters
```
nkern, line 1164: line break

hera_cal/frf.py
```diff
@@ -648,35 +1075,200 @@ def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
...
+            or if there is egregious cross-talk / ground pickup.
+            This arg is only used if beamfitsfile is not None.
+            Default is False.
+        frate_profiles: dict object, optional
+            Dictionary mapping antpairpol keys to fringe-rate profiles.
+            centered on a grid of length nfr spaced by dfr centered at 0.
+            default is None -> automatically calculate profiles using uvb.
+            User cannot provide both frate_profiles and uvb -- An error will be raised.
+        filter_kwargs: see fourier_filter for a full list of filter_specific arguments.
+        Returns
```
nkern, line 1243: You don't need a returns here, just 
````
Notes
------
Results are stored
```

hera_cal/frf.py
```diff
@@ -580,38 +913,128 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+
+        Returns
+        -------
+            frate_centers: dict object,
+                Dictionary with the center fringe-rate of each baseline in to_filter in units of mHz.
+                keys are antpairpols.
+            frate_half_widths: dict object
+                Dictionary with the half widths of each fringe-rate window around the frate_centers in units of mHz.
+                keys are antpairpols
+
```
nkern, line 985: no linebreak

nkern (commented):
some minor cosmetic clean-ups needed, and just a couple of comments on plumbing, but should be good to go shortly. thanks for your patience :)

hera_cal/frf.py
aewallwi: fixed.

hera_cal/frf.py
aewallwi: fixed.

hera_cal/frf.py
```diff
@@ -580,38 +913,128 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
             filt_flags[k] = f
             filt_nsamples[k] = eff_nsamples
 
-    def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
-                        frate_standoff=0.0, frate_width_multiplier=1.0, 
-                        max_frate_coeffs=None,
-                        min_frate_half_width=0.0, max_frate_half_width=np.inf,
+    def select_tophat_frates(self, case, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, min_frate_half_width=0.025,
```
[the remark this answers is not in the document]
aewallwi: technically no. I've made it module level (it does now require an extra blvecs arg and uvdata arg).

hera_cal/frf.py
```diff
@@ -580,38 +913,128 @@ def filter_data(self, data, frps, flags=None, nsamples=None,
...
+        verbose: bool, optional, lots of outputs!
+        frate_profiles: dict object, optional
+            Dictionary mapping antpairpol keys to fringe-rate profiles.
+            centered on a grid of length nfr spaced by dfr centered at 0.
+            only used if case == 'frate_profiles'
+        filter_kwargs: see fourier_filter for a full list of filter_specific arguments.
+
+        Returns
+        -------
+            frate_centers: dict object,
```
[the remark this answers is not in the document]
aewallwi: Fixed.

hera_cal/frf.py
aewallwi: fixed.

hera_cal/frf.py
```diff
@@ -648,35 +1075,200 @@ def tophat_frfilter(self, keys=None, wgts=None, mode='clean',
...
+            if 'data' in filter_kwargs:
+                input_data = filter_kwargs['data']
+            else:
+                input_data = self.data
+            input_data[k] *= phasor[:, None]
         if not mode == 'clean':
             if write_cache:
                 filter_cache = io.write_filter_cache_scratch(filter_cache, cache_dir, skip_keys=keys_before)
 
+    def split_frates_and_tophat_frfilter(self, case, keys=None, wgts=None, mode='clean', uvb=None, percentile_low=5., percentile_high=95.,
```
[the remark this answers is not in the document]
aewallwi: I've moved these two lines into `load_tophat_frfilter_and_write` which is the script level function that also deals with i/o.

hera_cal/frf.py
```diff
@@ -10,22 +10,165 @@
...
+    """
+    if keys is None:
+        keys = list(uvd.get_antpairpols())
+    if case == 'sky':
+        # if max_frate_coeffs is none and uvb is none, fringe-rate filter all modes that could be occupied by sky emission.
+        frate_centers, frate_half_widths = sky_frates(uvd, keys=keys, frate_standoff=frate_standoff,
+                                                      frate_width_multiplier=frate_width_multiplier, min_frate_half_width=min_frate_half_width)
+    elif case == 'max_frate_coeffs':
+        assert max_frate_coeffs is not None, "max_frate_coeffs must be provided if case='max_frate_coeffs'."
+        # if uvb is None and max_frate_coeffs is not None, use max_frate_coeffs.
```
nkern, line 148: can get rid of this comment b/c case is no longer selected implicitly

hera_cal/frf.py
```diff
@@ -10,22 +10,165 @@
...
+    unique_pols = set()
+    for bl in keys:
+        if bl[-1] not in unique_pols:
+            unique_pols.add(bl[-1])
+    reds = redcal.get_reds(antpos, pols=list(unique_pols), include_autos=True)
+    reds = redcal.filter_reds(reds, bls=keys)
+    return reds
+
+
+def select_tophat_frates(uvd, blvecs, case, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0, min_frate_half_width=0.025,
```
nkern, line 66: can we make this

```
select_tophat_frates(case, keys=None, blvecs=None, uvd=None, ...):
    ...
    elif case == 'max_frate_coeffs':
        keys = blvecs.keys()

```
i.e. case should be first argument, and uvd should be optional b/c its not strictly needed for the max_frate_coeeffs case

hera_cal/frf.py
```diff
@@ -10,22 +10,165 @@
...
+    min_frate_half_width: float, optional
+        minimum half-width of fringe-rate filter, regardless of baseline length in mHz.
+        Default is 0.025
+    keys: list of visibilities to filter in the (i,j,pol) format.
+      If None (the default), all visibilities are filtered.
+    max_frate_coeffs, 2-tuple float
+      Maximum fringe-rate coefficients for the model max_frate [mHz] = x1 * | EW_bl_len | [ m ] + x2."
+      These are used only if case == 'max_frate_coeffs'.
+    uvb: UVBeam object, optional
+     UVBeam object with model of the primary beam. Only used if case=='uvbeam'
```
nkern, line 102: indentation

hera_cal/frf.py
```diff
@@ -10,22 +10,165 @@
...
+        If case == 'uvbeam', then determine fringe rate centers and half widths
+        from histogram of main-beam wrt instantaneous sky fringe rates
+        If case == 'frate_profiles': then use user-supplied fringe-rate profiles
+        and compute cutoffs based on percentile_low and percentile_high
+    frate_standoff: float, optional
+        additional fringe-rate to add to min and max of computed fringe-rate bounds [mHz]
+        to add to analytic calculation of fringe-rate bounds for emission on the sky.
+        default = 0.0.
+    frate_width_multiplier: float, optional
+     fraction of horizon to fringe-rate filter.
```
nkern, line 91: indentation

hera_cal/frf.py
```diff
@@ -10,22 +10,165 @@
...
+        Dictionary with the center fringe-rate of each baseline in to_filter in units of mHz.
+        keys are antpairpols.
+    frate_half_widths: dict object
+        Dictionary with the half widths of each fringe-rate window around the frate_centers in units of mHz.
+        keys are antpairpols
+    """
+    if keys is None:
+        keys = list(uvd.get_antpairpols())
+    if case == 'sky':
+        # if max_frate_coeffs is none and uvb is none, fringe-rate filter all modes that could be occupied by sky emission.
```
nkern, line 143: same comment here

hera_cal/frf.py
```diff
@@ -10,22 +10,165 @@
...
+        frate_centers = {k: 0.0 for k in keys}
+    elif case == 'frate_profiles':
+        assert frate_profiles is not None, "frate_profiles must be provided if case='frate_profiles'"
+        frate_centers, frate_half_widths = get_fringe_rate_limits(uvd, nfr=nfr, dfr=dfr, percentile_low=percentile_low,
+                                                                  percentile_high=percentile_high, keys=keys, verbose=verbose, frate_standoff=frate_standoff,
+                                                                  fr_freq_skip=fr_freq_skip, frate_width_multiplier=frate_width_multiplier,
+                                                                  min_frate_half_width=min_frate_half_width, frate_profiles=frate_profiles)
+    elif case == 'uvbeam':
+        assert uvb is not None, "uvb must be provided iv case='uvbeam'."
+        # if uvb is not None, get fringe-rates from binning.
```
nkern, line 159: same here

hera_cal/frf.py
aewallwi: done.

hera_cal/frf.py
aewallwi: done.

hera_cal/frf.py
aewallwi: done.

hera_cal/frf.py
aewallwi: done.

hera_cal/frf.py
aewallwi: done.

hera_cal/frf.py
aewallwi: done.

nkern (approved):
looks good thanks

jsdillon (approved):

PR #739, opened 2021-09-08
Set lmin / lmax from integration times rather then file ordering

aewallwi:
Variables `lmin` and `lmax` in `lstbin.config_lst_bin_files` are meant to represent the lsts of the minimum and maximum integrations on each night. They are set by the first/last integrations on the first and last files on each night to be lst binned *but* the files are ordered by their names. If the names are not ascending by observation time, which can be the case when the files are labeled by LST instead of JD, then the min lst might be a late integration. If lst_stop is not set, this can cause significant chunks of data to be dropped in some cases. For example, if we are lst-binning files that start at some large radian LST (such as 6.2 say) and end at 0.08, lstmax will erroneously be be set to 6.2 and lstmin will be 0.08. If lst_stop is not provided but we provide and lst_start of 6.19, then lst_stop will be set to lmax=6.2 (in the below line), discarding most of the files from lstbinning.

https://github.com/HERA-Team/hera_cal/blob/a863658c8212fb0c60e2624900697b3fe316b40f/hera_cal/lstbin.py#L593

and this will cutoff all the LSTs after 0.0.

To fix this, I now set lmin to be the lst of the minimum observation time (not the first file) and lmax to the lst of the maximum observation time (not the last file).

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -555,9 +555,8 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
         # get times
         dlsts, dtimes, larrs, tarrs = io.get_file_times(dfs, filetype='uvh5')
 
-        filemaxtime = np.argmax([tarr.max() for tarr in tarrs]) # we want lmax to be the lst from the file with the maximum time actually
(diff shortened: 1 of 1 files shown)


jsdillon:
Closed in favor of #741

PR #741, opened 2021-09-10
Make config_lst_bin_files cover 24 hours of LST to avoid edge cases and assumptions about the first and last file

jsdillon:


nkern:
@jsdillon as I recall, the reason for leaving those files out is such that we don't run empty jobs. even if we happen to currently have a particular pipeline (i.e. makeflow) that runs those jobs trivially fast, I think its a design flaw for hera_cal to identify those LSTbin files as "runnable". e.g. someone else's pipeline may not run those jobs trivially fast, so we are kind of building a rigid structure that only fits our current pipeline, and that's not personally how I originally intended the script to behave.

Alternatively, can you elaborate why we can't we just address the edge cases that were raised by the issue? I'll keep looking 
 through this PR, and also at the issue to see if I can come up w/ an elegant solution. thanks

nkern:
It seems that the issue raised in #739 shouldn't be that hard to solve, unless I'm reading it wrong. I thought I added a kwarg to set the start of the lst grid regardless of what's in the files, which could be useful for example when LST binning data that wrap across 2pi. Does that not work here? 

I mean, another option if you want to keep the stuff you've added here is to create a 24-hour grid as you suggest, but to then determine which output LST bin files will be empty and then remove those lst bins from the grid.

jsdillon:
Just as a proof of concept, here a sample result with this branch: 

Here's |V(12, 13, 'ee')| after x-talk subtraction for each epoch and then LST-binned together:
image

And here's nsamples for the same visibilities:
image

hera_cal/lstbin.py
```diff
@@ -512,37 +510,34 @@ def lst_bin_arg_parser():
...
-        Default is lst_start + 2pi.
-    fixed_lst_start : type=bool, if True, LST grid starts at lst_start, regardless of LST of first data
-        record. Otherwise, LST grid starts at LST of first data record.
+        Default is first LST of the first file of the first night.
     ntimes_per_file : type=int, number of LST bins in a single output file
 
     Returns
     -------
-    lst_grid : float ndarray holding LST bin centers
+    lst_grid : float ndarray holding LST bin centers. Spans 2 pi radians.
```
nkern, line 532: does it make sense to only return lst_grid indices that are populated by the file_lsts? In other words, we should drop file_lsts that are empty, as well as lst_grid indices that are empty. that makes it closer to the behavior we had before, but I think also accomplishes what you want here, which is a simplification of the creation of the grid.

hera_cal/lstbin.py
```diff
@@ -552,70 +547,52 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
...
-        begin_lst = lmin
-
-    # get stopping LST for output binning
-    if lst_stop is None:
-        lst_stop = lmax
-    if lst_stop < begin_lst:
-        lst_stop += 2 * np.pi
+        all_lsts = [lst for larrs in lst_arrays for larr in larrs for lst in larr]
+        all_times = [time for tarrs in time_arrays for tarr in tarrs for time in tarr]
+        lst_start = all_lsts[np.argmin(all_times)]
```
nkern, line 559: why argmin all_times?

hera_cal/lstbin.py
```diff
@@ -552,70 +547,52 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
...
     lst_grid = make_lst_grid(dlst, begin_lst=begin_lst, verbose=verbose)
     dlst = np.median(np.diff(lst_grid))
 
-    # get starting and stopping indicies
-    start_index = np.argmin(np.abs(lst_grid - lst_start))
-    stop_index = np.argmin(np.abs(lst_grid - lst_stop))
+    # enforce that lst_arrays are in the same range as the lst_grid
+    for larrs in lst_arrays:
+        for larr in larrs:
+            while np.any(larr < np.min(lst_grid)):
```
nkern, line 569: i don't think you need this while statement

hera_cal/lstbin.py
```diff
@@ -552,70 +547,52 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
...
+                larr[larr < np.min(lst_grid)] += 2 * np.pi
+            while np.any(larr > np.max(lst_grid)):
```
nkern, line 571: here too

hera_cal/lstbin.py
```diff
@@ -552,70 +547,52 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
...
+                larr[larr > np.max(lst_grid)] -= 2 * np.pi
 
     # get number of output files
-    nfiles = int(np.ceil(float(stop_index - start_index) / ntimes_per_file))
+    nfiles = int(np.ceil(1.0 * len(lst_grid) / ntimes_per_file))
```
nkern, line 575: nitpicking but: you don't need the 1.0. int division results in float for python3+. if you want to make it python2 compatible I think the more readable thing is float(), rather than implicit type casting via float multiplication. but not really a big thing anyways

hera_cal/lstbin.py
```diff
@@ -552,70 +547,52 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, lst_
...
 
-    # get output file lsts
-    file_lsts = [lst_grid[start_index:stop_index][ntimes_per_file * i:ntimes_per_file * (i + 1)] for i in range(nfiles)]
+    # get output file lsts that are not empty
+    all_file_lsts = [lst_grid[ntimes_per_file * i:ntimes_per_file * (i + 1)] for i in range(nfiles)]
+    file_lsts = []
+    for i, f_lst in enumerate(all_file_lsts):
+        fmin = f_lst[0] - (dlst / 2 + atol)
+        fmax = f_lst[-1] + (dlst / 2 + atol)
+        for lst in [lst for larrs in lst_arrays for larr in larrs for lst in larr]:
```
nkern, line 583: just for readability, can you make this
```
# flattened lsts across days, nights, files
flat_lsts = [lst for...]
for lst in flat_lsts:
```

hera_cal/lstbin.py
jsdillon: Not at all. This is simply asserting that the starting LST should correspond to the first JD on the first night. This method allows the days to come in in any order and uses their JDs (which don't have a phase wrap, unlike lst) to figure out which LST is "first."

hera_cal/lstbin.py
jsdillon: fair point. fixed.

hera_cal/lstbin.py
jsdillon: Done.

hera_cal/lstbin.py
jsdillon: Ok, I've cut the lst_grid to only include lsts that are in the file_lsts

hera_cal/lstbin.py
jsdillon: It's probably not necessary, though I do believe that there are pathological values of `lst_start` that would make this happen. I'd prefer to keep it... definitely not hurting anything.

jsdillon:
Thanks @nkern for the review. It's ready for another look, I think.

hera_cal/lstbin.py
nkern: Consider the following example: let's say I want to LST bin two (or more) nights from either end of the year (or perhaps more realistically, across different seasons). Let's say these are JDs 2458100.2-.6 and 2458400.2-.6 (almost a year separated). These corresponds to JDs of 6.19 - 2.43 radians and 5.07 - 1.31 radians, respectively. This argmin will grad 2458100.2 as the starting LST, so the LST grid will be 6.19 - 12.47 radians, but there is data before that 6.19 radians coming from the second day, which will then get reflected up to 11.3 radians. This could not be a problem, because those LSTs will be binned and then re-reflected down to 5.07 radians, but I wonder if these files may actually just be dropped? Or will this not happen?

nkern: also this may pose an issue for binning the first (and last) bin of the grid. before we used lst_start and lst_end to specify an LST grid that spans across and beyond the data entirely (assuming we were binning data that did not span exactly 24 hours). with a odd reflection like what I describe above, its possible that the first and last LST bin of the grid will not grab the data that would have fallen in the bin on the other side, if you know what I mean

hera_cal/lstbin.py
nkern: did this not get pushed up, or am I just seeing a stale diff? it looks the same

jsdillon: Uh... it's there. https://github.com/HERA-Team/hera_cal/blob/304763303f1de08957ebb7a0c167c8fca950ef7c/hera_cal/lstbin.py#L577-L583

hera_cal/lstbin.py
jsdillon: Regarding the first issue, of trying to bin together two days from different seasons, this edge case can be easily addressed by manually providing an `lst_start`. I'm trying to do the most sensible thing where no `lst_start` is provided and I think this is it. The only alternative is making the `lst_start` parameter mandatory, but I feel like that's more API-breaking than this current PR. 

The current code that drops files has logic borrowed by LST-bin files that makes sure that no data gets dropped. I'm not seeing how we'd lose an integration.

nkern: right I forgot that lst_start is still a parameter, okay sounds good thanks!

Issue #744, opened 2021-10-13
Faulty standard deviation calculation in LST-binning


matyasmolnar:
I think I have found an issue when the standard deviations are calculated in `lst_bin`, see [here](https://github.com/HERA-Team/hera_cal/blob/7829a309519cef386d0053c33673b78e80901dde/hera_cal/lstbin.py#L380).

The `real_std` and `imag_std` are calculated from the `d` variable, which has been zero-padded for nans (all flagged indices have been nan'd at this stage) just above `here` with `d[~isfinite] = 0.0` (that is if `median = False` as the `if median:` branch above does not affect `d`).

The nan indices are therefore not discarded in `np.nanstd` and the zeros are contributing to the resulting standard deviation.

A quick fix would be to have, e.g.:
```
d_w = d.copy()
isfinite = np.isfinite(d_w)
d_w[~isfinite] = 0.0
n[~isfinite] = 0.0

norm = np.sum(n, axis=0).clip(1e-99, np.inf)
real_avg.append(np.sum(d_w.real * n, axis=0) / norm)
imag_avg.append(np.sum(d_w.imag * n, axis=0) / norm)
```
which then does not modify d
alternatively, the weighted average could be done with `np.ma.average` with `weights=n` and `mask=np.isnan(d)` .

jsdillon:
Fixed via #745

PR #745, opened 2021-10-13
nan consistency in lst-binning

matyasmolnar:
Quick fix to issue #744 

[edit by nk]:
to fix this we use nansum to ignore nans in the weighted mean, which preserves nan for downstream nanstd estimator


nkern:
Great catch! thanks @matyasmolnar

nkern (approved):

matyasmolnar:
@nkern I've just added a couple of lines to filter out (the many) warnings raised when an all nan row is encountered in the `np.nanstd` statistics [here](https://github.com/HERA-Team/hera_cal/blob/7829a309519cef386d0053c33673b78e80901dde/hera_cal/lstbin.py#L380)

nkern:
@matyasmolnar looks great, but can we keep the nanned data as 1.0? there are good reasons to keep fake data as 1 not zero, mainly in case we want to divide two datasets and not incur divide by zero errors. after that reversion I think this is good to go!

hera_cal/utils.py
```diff
@@ -1318,7 +1318,10 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
             # take the weighted average
             wsum = np.sum(w, axis=0).clip(1e-10, np.inf)  # this is the normalization
             davg = np.sum(d * w, axis=0) / wsum  # weighted average
-            fmax = np.max(f, axis=2)             # collapse along freq: marks any fully flagged integrations
+            # set flagged data to 1.0+0.0j
```
nkern, line 1321: @jsdillon @aewallwi we are proposing to change the behavior of `red_average` such that pixels that are fully flagged are set to 1.0 instead of 0.0. this is more consistent with what the LST binner does. can you comment on any possible ramifications of this?

In theory, any downstream block that honors flags should not be affected. the reason to move to 1.0 instead of 0.0 is that it is 1. consistent with lst binner and 2. plays nicer with division

hera_cal/tests/test_utils.py
```diff
@@ -611,6 +611,9 @@ def test_red_average():
     d = np.asarray([hd.get_data(bl + ('ee',)) for bl in reds[0]])
     w = np.asarray([(~hd.get_flags(bl + ('ee',))).astype(float) for bl in reds[0]])
     davg = np.sum(d * w, axis=0) / np.sum(w, axis=0).clip(1e-10, np.inf)
+    # set flagged data to 1.0+0.0j
```
nkern, line 614: the consequence of changing that line in `utils.red_average` is a need to update this unit test, which seems fairly isolated to this specific functionality and wouldn't affect other parts of hera_cal.

jsdillon:
I don't feel too strongly about this, but the convention in `redcal` is that flagged visibilities are 0 and flagged gains are 1. 

That said, I don't think there will be any downstream consequences...

hera_cal/utils.py
aewallwi: I think this should be fine. I approve.

jsdillon:
Are we waiting on anyone's approval here?

matyasmolnar:
Just waiting on approval from @aewallwi I believe

aewallwi (approved):

PR #749, opened 2021-12-17
Ensure that autocorrelations remain real

jsdillon:
This PR on `pyuvdata` https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1110 is currently breaking hera_cal.'s tests. I've tracked this down to a few sources:

- Some test files had small but non-zero imaginary parts of their autocorrelations. I've rewritten all the relevant tests files to eliminate this.
- `apply_cal` used to divide out one antenna's gain and then the other's. This could create non-zero imaginary parts due to round-off errors. I've added in special cases for this.
- Some unit tests involved multiplying aucorrelations by complex numbers. Since will generally raise an error in pyuvdata now, unless we explicitly silence it, I've just modified those tests.

However, there's still three tests that are failing with this branch and that `pyuvdata` branch, all in `vis_clean`. I *think* what's going on is that delay/FR-filtering autocorrelations is creating non-zero imaginary comments, though I'm not sure. @aewallwi I'd like you to look into it and see what we need to do to get hera_cal compatible with pyuvdata's new checks.


bhazelton (approved):
Looks fine to me, but I'm not an expert on hera_cal.

PR #751, opened 2021-12-22
Begin preparing hera_cal for BDA with updates to hera_cal.io

### LINKED ISSUE #509: Enable up/down sampling of BDA data in hera_cal.io
The current plan for H3C is to perform the calibration at a common cadence. This requires up/down sampling of data. Ideally, this will be done under the hood in a HERAData object before building DataContainers.

This requires this issue to be done: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues/428

jsdillon:
This PR beings the process of preparing the pipeline for baseline-dependent averaging, starting with tools in `hera_cal.io`. 

The main changes are:
- Enabling upsampling and downsampling as a property of a hera_data object that will be applied every time it does a read.
- Changing how the `HERAData.times` and `HERAData.lsts` are calculated, removing `np.unique` and instead attempting to figure out the "base" set of times (i.e. those without BDA).
- Better handling of single integration files.
- Enabling new selects on `HERAData.read()` (`lsts`, `lst_range`, and `time_range`). The most important of these is probably `time_range` which is a much safer way to perform partial i/o on a BDA file and will be crucial for updating tasks that are chunked in time, like `redcal` and `abscal`.
- An overhaul of `partial_time_io()` to enable the new selects and BDA better.
- Lots more unit tests, including on BDA files created with https://github.com/HERA-Team/baseline_dependent_averaging/blob/main/scripts/apply_bda.py 

In the course of writing this PR, I discovered a small bug in `pyuvdata`. This PR will not pass tests until https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1122 is approved and merged in.

This PR closes #509.


hera_cal/io.py
```diff
@@ -312,6 +323,12 @@ def __init__(self, input_data, filetype='uvh5', **read_kwargs):
         else:
             raise NotImplementedError('Filetype ' + self.filetype + ' has not been implemented.')
 
+        # save longest and shortest integration times in the file for later use in up/downsampling
+        # if available, these will be used instead of the ones in self.integration_time during partial I/O
+        if self.integration_time is not None:
+            self.longest_integration = np.max(self.integration_time)
+            self.shortest_integration = np.min(self.integration_time)
```
nkern, line 330: should this be defined as None if integration_time is None?

jsdillon: We could. I used `if hasattr(self, 'shortest_integration'):` instead. Do you have a strong preference?

nkern: no strong preference, it just seems better from a bookkeeping perspective

jsdillon: Done.

hera_cal/io.py
```diff
@@ -312,6 +323,14 @@ def __init__(self, input_data, filetype='uvh5', **read_kwargs):
         else:
             raise NotImplementedError('Filetype ' + self.filetype + ' has not been implemented.')
 
+        # save longest and shortest integration times in the file for later use in up/downsampling
+        # if available, these will be used instead of the ones in self.integration_time during partial I/O
+        self.longest_integration = None
+        self.longest_integration = None
+        if self.integration_time is not None:
+            self.longest_integration = np.max(self.integration_time)
```
nkern, line 331: the only way this may break is if the integration_time is updated downstream to track coherent averaging. Currently we do this by adding to `nsamples` but I vaguely recall a conversation years ago about possibly changing integration_time?

nkern (approved):
this looks good to me overall. minor comments.

I'm not as familiar with the partial_time_io func so I'll let aew review that.

aewallwi (approved):
Looks good to me. Feel free to merge.

hera_cal/io.py
```diff
@@ -256,13 +257,17 @@ class HERAData(UVData):
     # times_by+bl: dictionary mapping antpairs to times (JD). Also includes all reverse pairs.
```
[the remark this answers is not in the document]

PR #752, opened 2021-12-30
Bda redcal

jsdillon:
This PR, which follows from (and requires) #751, enables redcal with BDA data via either upsampling or downsampling to a common temporal resolution. This is enabled by passing through upsample/downsample options to the command line and by using `time_range` instead of `times` for partial i/o. Also, this PR improves the handling of Delta t calculations on single integration files and files where some baselines have a single integration because of BDA.

The good news is that this appears to be working quite well on a BDA file I created with H4C data from 8 integrations (BDAed to 8 or 4 or 2 or 1) starting at 2459122.30030:
Screen Shot 2021-12-30 at 10 07 35 AM
Screen Shot 2021-12-30 at 10 08 25 AM

Based on these results, I wonder whether we should consider redcal on BDA data that's been downsampled to a common resolution, rather than upsampled. We'll get some decoherence, but I'm not sure that actually matters as long as all redundant baselines decohere the same and there's no fine temporal structure in the gains.


hera_cal/noise.py
```diff
@@ -34,6 +35,37 @@ def interleaved_noise_variance_estimate(vis, kernel=[[1, -2, 1], [-2, 4, -2], [1
     return variance
 
 
+def infer_dt(times_by_bl, bl, default_dt=None):
+    '''Attemps to infer dt for a baseline from times by bl. If len(times_by_bl[bl])
+    
+    Arguments:
```
aewallwi, line 41: Maybe consider adopting the numpy format which is prevalent in many docstrings in `hera_cal`

hera_cal/noise.py
```diff
@@ -34,6 +35,37 @@ def interleaved_noise_variance_estimate(vis, kernel=[[1, -2, 1], [-2, 4, -2], [1
     return variance
 
 
+def infer_dt(times_by_bl, bl, default_dt=None):
+    '''Attemps to infer dt for a baseline from times by bl. If len(times_by_bl[bl])
```
aewallwi, line 39: Explicitely state what dt is.

hera_cal/noise.py
```diff
@@ -34,6 +35,37 @@ def interleaved_noise_variance_estimate(vis, kernel=[[1, -2, 1], [-2, 4, -2], [1
...
+    
+    Arguments:
+        times_by_bl: dictionary mapping antenna pair tuples to 1D time arrays (usually in JD)
+        bl: antpair tuple e.g. (0, 1) or basebline tuple e.g. (0, 1, 'ee'). Polarization ignored.
+        default_dt: default value to return if times cannot be infered. Default None raises ValueError.
```
aewallwi, line 44: mention units.

hera_cal/noise.py
aewallwi: I can see this going either way though since `redcal.py` does not use this convention. Up to you.

hera_cal/noise.py
```diff
@@ -51,8 +83,7 @@ def predict_noise_variance_from_autos(bl, data, dt=None, df=None, nsamples=None)
         Noise variance predicted on baseline bl in units of data squared
```
aewallwi, line 83: Maybe mention the default value for uniform time samples just so people easily understand where things come from.

aewallwi (changes requested):
Would it be possible to include a unittest that runs redcal on a bda file? I don't seem to see any here.

hera_cal/noise.py
jsdillon: Done

hera_cal/noise.py
jsdillon: I'm going to keep this consistent within this module. If we want to make the whole repo one style, that feels like a separate PR.

hera_cal/noise.py
jsdillon: done.

hera_cal/noise.py
jsdillon: Done.

hera_cal/noise.py
aewallwi: 👍

aewallwi (approved):
This all looks good. I approve!

PR #753, opened 2022-01-08
fix floor division.

aewallwi:
numpy 1.22 and above throw value errors when you try to perform floor division with complex numbers. I've modified the implementation in `datacontainer.py` to perform floor division by taking the ratio of the complex numbers and then separately rounding down the real and imaginary components.

### DIFF
--- a/hera_cal/datacontainer.py
+++ b/hera_cal/datacontainer.py
@@ -61,7 +61,7 @@ class DataContainer:
             self._pols = set([k[-1] for k in self._data.keys()])
 
             # placeholders for metadata (or get them from data, if possible)
-            for attr in ['ants', 'data_ants', 'antpos', 'data_antpos', 
+            for attr in ['ants', 'data_ants', 'antpos', 'data_antpos',
                          'freqs', 'times', 'lsts', 'times_by_bl', 'lsts_by_bl']:
                 if hasattr(data, attr):
                     setattr(self, attr, getattr(data, attr))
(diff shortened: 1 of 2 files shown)


hera_cal/datacontainer.py
```diff
@@ -324,14 +324,22 @@ def __floordiv__(self, D):
             # iterate over D keys
             for i, k in enumerate(D.keys()):
                 if self.__contains__(k):
-                    newD[k] = self.__getitem__(k) // D[k]
+                    if not (np.any(np.iscomplex(self.__getitem__(k))) or np.any(np.iscomplex(D[k]))):
```
jsdillon, line 327: Instead of checking if every element is not complex, I wonder if it would be better to use `np.iscomplexobj()` which just checks the datatype.

hera_cal/datacontainer.py
```diff
@@ -324,14 +324,22 @@ def __floordiv__(self, D):
...
+                        div = self.__getitem__(k) / D[k]
+                        newD[k] = np.real(div).astype(int) + 1j * np.imag(div).astype(int)
 
             return DataContainer(newD)
 
         else:
             newD = copy.deepcopy(self)
             for k in newD.keys():
-                newD[k] = newD[k] // D
+                if not (np.any(np.iscomplex(newD[k])) or np.any(np.iscomplex(D))):
```
jsdillon, line 338: Same comment.

jsdillon (changes requested):
Just one question that might make the code slightly faster. Otherwise, looks good.

jsdillon (approved):

hera_cal/datacontainer.py
aewallwi: It looks like `iscomplexobj` doesn't work on the dictionary level. Here is a test in numpy 1.22.

```
>>> x = dict({0:np.ones(10, dtype=complex)})
>>> x
{0: array([1.+0.j, 1.+0.j, 1.+0.j, 1.+0.j, 1.+0.j, 1.+0.j, 1.+0.j, 1.+0.j,
       1.+0.j, 1.+0.j])}
>>> np.iscomplexobj(x)
False
```
I've changed the `np.any(np.iscomplex())` bits to `np.iscomplexobj()` at the numpy array level but we still iterate through the dictionary keys.

hera_cal/datacontainer.py
aewallwi: See above.

Issue #541, opened 2019-10-17
Update and Modularize the LST-binner


jsdillon:
For on-site LST-binning, we need to update and modularize the LST-binner to enable the following:

- Take in some notion of blessed files/antennas/channels ranges from the observing team
- Temporary saving of medians and MADs for a given stack 
- Comparison of modified z-scores (This is basically #258) to find: 
    - Bad freqs/times over all baselines on a given night
    - Bad antennas over all freqs/times on a given night
    - Bad baselines over all freqs/times on a given night
    - Bad nights
- Ability to take flags from sums (from XRFI, MAD-clipping, etc.) and apply them to the diffs.

See this rough design document, discussed in the 10/16/19 analysis meeting: https://docs.google.com/document/d/1Cf8q88TVHlCCuO0357AsoYc-RVddMhJtzsgR-Yatc94/edit#

Issue #258, opened 2018-04-17
LST-Metrics: find bad days


jsdillon:
We need a way to figure out if specific days are systematically flagged as bad by the MAD clipping. Likely the easiest way to do this is compare LST vs. day, freq vs. day, and baseline vs. day and, folding over the other two axes, count the number of times a day gets flagged by the MAD clipping. This can then feedback into further analysis.

This is likely a post-IDR2.1 feature.

PR #754, opened 2022-01-10
fix min_N criterion to sigma clipping

matyasmolnar:
This PR ensures that there are at least min_N unflagged data points when sigma clipping - the current implementation only looks at the axis size rather than the number of unflagged data points.

The [sigma_clip](https://github.com/HERA-Team/hera_cal/blob/0339da21d4e050f035abde1762c8b4990831709a/hera_cal/lstbin.py#L965) function only looks at the array shape to determine whether to clip or not (with `if array.shape[axis] < min_N:`) - this does not consider flagged data points, so even if the array has `array.shape[axis] > min_N` some of the freq slices may have fewer than `min_N` data points. This PR corrects this and sets clip flag slices to `False` if there are fewer than `min_N` unflagged data points for each freq slice.

This effect is negligible for `min_N = 5` but as `min_N` increases this becomes significant.


hera_cal/lstbin.py
```diff
@@ -345,6 +345,10 @@ def lst_bin(data_list, lst_list, flags_list=None, nsamples_list=None, dlst=None,
                 # get real + imag flags
                 clip_flags = real_f + imag_f
 
+                # apply min_N condition as sigma_clip only checks axis size, not number of flags
+                sc_min_N = np.logical_not(f).sum(axis=0) < min_N
```
jsdillon, line 349: Wouldn't it be more sensible to put this into the `sigma_clip` function?

jsdillon:
Thanks @matyasmolnar for bringing this up.

Your PR seems reasonable, though it does raise a question. Say for example, you're binning together 8 points and `min_N` is 5 and `sigma_clip=True`. If 4 of the 8 points were already flagged, should the LST-binner just flag everything because there weren't even enough points to identify outliers? 

@nkern, I'm curious what you think.

hera_cal/lstbin.py
matyasmolnar: @jsdillon yes we could put it in, but then for this particular scenario where we clip both Re and Im components, `sc_min_N = np.logical_not(f).sum(axis=0) < min_N` would be repeated so leaving it outside the `sigma_clip` is a bit faster - if `sigma_clip` is used elsewhere, however, it would make sense to put it in the function.

matyasmolnar:
@jsdillon yes good question. For the LST-binning functions/script, `min_N` always refers to the minimum number of points needed for sigma clipping. I don't think that there is a step that completely flags a slice if a minimum number of points is not reached - a separate parameter could be introduced for this? separate; because sigma clipping could further flag some data points, this minimum parameter may not be equal to `min_N`?

hera_cal/lstbin.py
jsdillon: Is this step an appreciable fraction of the run-time of the function?

jsdillon:
> @jsdillon yes good question. For the LST-binning functions/script, `min_N` always refers to the minimum number of points needed for sigma clipping. I don't think that there is a step that completely flags a slice if a minimum number of points is not reached - a separate parameter could be introduced for this? separate; because sigma clipping could further flag some data points, this minimum parameter may not be equal to `min_N`?

In my example, if `min_N` is 5 and there are 8 data points and 2 prior flags, and sigma clipping adds 2 more flags, I don't think this should trigger flagging everything. I think the min_N is a minimum prior to sigma clipping, not afterwards.

Anyway, this is a design question that I think @nkern should weigh in on.

hera_cal/lstbin.py
matyasmolnar: Just looked at timings and it hardly makes a difference - I can include it in the `sigma_clip` function if you want? I also tested out the `sigma_clip` function and I think it has various problems:

- It modifies the input array `here` which means that there may be a difference if Re or Im is clipped first - this should be removed
- The `sigma_clip` function as it stands only works with 2D arrays where axis=0 is specified - anything else and it fails, so these arguments are a bit misleading. As we're only using it here for LST-binning we could make it simpler?

hera_cal/lstbin.py
```diff
@@ -1000,22 +1002,24 @@ def sigma_clip(array, flags=None, sigma=4.0, axis=0, min_N=4):
...
     # get MAD! * 1.482579
-    scale = np.nanmedian(np.abs(array - location), axis=axis) * 1.482579
+    scale = np.nanmedian(np.abs(array_ - location), axis=axis) * 1.482579
 
     # get clipped data
-    clip = np.abs(array - location) / scale > sigma
+    clip = np.abs(array_ - location) / scale > sigma
 
-    # set clipped data to nan and set clipped flags to True
-    array[clip] *= np.nan
```
matyasmolnar, line 1018: I've removed this as it modifies the input array, which means that the Re and Im clipping may see different data arrays, meaning that the ordering matters. We don't want this, therefore `array[clip] *= np.nan` is removed.

hera_cal/lstbin.py
```diff
@@ -1000,22 +1002,24 @@ def sigma_clip(array, flags=None, sigma=4.0, axis=0, min_N=4):
     # create empty clip_flags array
     clip_flags = np.zeros_like(array, np.bool)
 
+    # make copy of data array as to not replace original array
+    array_ = array.copy()
```
matyasmolnar, line 1006: We have to make a copy as to not change input array.

matyasmolnar:
@jsdillon @nkern I added another commit which removes the line that modifies the input array. I've explained why this is required above and in line comments.

As I mentioned above, the `sigma_clip` function doesn't really work for arrays that are not 2D and for which `axis=0` is specified. I've left it as is, but if you wanted to have a more flexible function then please see this `[mad_clip](https://github.com/matyasmolnar/robstat/blob/dbe5b6fd8b26d8ee85e33562d70315dd805103de/robstat/stdstat.py#L12)` function that I wrote. I think it is fine for now though and serves the current purpose.

I will distribute a memo very shortly on a different clipping approach that is based on Mahalanobis distances - this is computationally expensive but should perform better than the current MAD-clipping.

hera_cal/lstbin.py
```diff
@@ -1000,22 +1002,24 @@ def sigma_clip(array, flags=None, sigma=4.0, axis=0, min_N=4):
...
-    location = np.nanmedian(array, axis=axis)
+    location = np.nanmedian(array_, axis=axis)
 
     # get MAD! * 1.482579
-    scale = np.nanmedian(np.abs(array - location), axis=axis) * 1.482579
+    scale = np.nanmedian(np.abs(array_ - location), axis=axis) * 1.482579
 
     # get clipped data
-    clip = np.abs(array - location) / scale > sigma
+    clip = np.abs(array_ - location) / scale > sigma
```
jsdillon, line 1020: Can't we just return clip instead of doing `clip_flags[clip] = True`?

jsdillon:
Hi Matyas,

I think this is reasonable, though I'd like to get @nkern's opinion on the design choice above (should having too few samples mean no sigma clipping, or completely flagging the bin?).

I'm not familiar with the Mahalanobis distance, so I look forward to your memo on that.

Just to put things on your radar, if you're interested in a deeper dive into LST-binning, there's been this idea floating around for a long time that we want to keep track of the statistics of what the LST-binner flags due to sigma clipping so as to identify bad times/freqs/days/baselines. That's discussed a bit in some old, unresolved issues #541 and #258.

hera_cal/lstbin.py
matyasmolnar: `if flags is not None` then `clip != clip_flags` so could leave as is?

matyasmolnar:
Ok great, thanks Josh.

As for #541 and #258 I actually ended up writing some code a while ago to track the MAD-clipping that occurs during LST-binning - see https://github.com/matyasmolnar/hera_cal/blob/jd_tracking/hera_cal/lstbin.py - this involves pickling a DataContainer that contains all this tracked info, which is not hugely aesthetic... very happy to further work on this.

nkern:
hi @matyasmolnar thanks for all of your efforts on this. I will try to take a look at this today. I'll list here a few of my thoughts at the onset.

I remember that I chose to not look at the flags and just look at the axis size, which would allow some cases where N < min_N but as you state this only really matters for large min_N. I think my thinking here was that we did *not* want to be inconsistent on sigma clipping between frequency bins--this could lead to significant spectral structure! You could make an argument for sigma clipping some specific LST bins while not clipping others (within one LST chunk being looked at by the binner), but I think we almost certainly don't want to perform inconsistent sigma clipping wrt frequency.

You may have already addressed this (I'll take a look at the code shortly) but if not, lets try to think about how we can make the sigma clipping more flexible to flags, as this PR tries to implement, while retaining the functionality of an ON/OFF switch for all frequency bins at a time (rather than determining if each freq bin needs clipping based on min_N). lmk if you have any thoughts on this, thanks!

hera_cal/lstbin.py
nkern: yes sigma_clip purposely operates in place to save memory, and I think you've pointed out something that we actually didn't consider too closely (clipping re or im first) but you'll notice that if re or im is flagged, the other is automatically flagged (for obvious reasons) so I don't actually know if the order makes a difference. correct me if you think this is incorrect, but its a nice spot regardless.

nkern: about the sigma_clip function args: I agree that its possible it is fairly brittle, and if you've noticed that setting axis!=0 makes it fail, then we should either fix it or get rid of the argument and hardcode axis=0. Generally I'd say we do the former, but b/c this function is very specific to the LST binner (e.g. I wouldn't use hera_cal.lstbin.sigma_clip for any generic sigma clipping routine in my analysis), I think doing the latter is a good solution: easier and less bug-prone.

hera_cal/lstbin.py
nkern: let's re-think if this is really necessary, as stated above. memory performance for the LST binner is generally crucial, so saving memory wherever possible is usually helpful

hera_cal/lstbin.py
nkern: I don' think it does: you can have a complex number as 1 + 1j*nan. b/c we do `sigma_clip(data.real, ...` and `sigma_clip(data.imag)` I don't think this is actually a concern

nkern:
@matyasmolnar could you also just help me understand what the motivation here is? are you trying to perform sigma_clipping for large min_N? are you interested in the case of sigma_clipping some freq bins and not others? if I get a sense for what you are interested in I might be able to help you get there with the other constraints we've set for the LST binner, thanks :)

hera_cal/lstbin.py
matyasmolnar: I think this is necessary if we still want `flags` as an input, as we don't want to change the original input array, as this would impact sigma clipping for the Im component if Re is done first

Since the `flags` arg is not used in the routine (because `d` already has nan entries for flagged points), we could get rid of the `flags` arg and the `array` here would not have to be copied

hera_cal/lstbin.py
matyasmolnar: The way that `sigma_clip` is currently implemented with the function modifying the data array in place, the ordering does make a difference here. If a point is clipped for the Re component and subsequently flagged, then that point is not included in the Im sigma clipping calculations, which would change the Im MAD. We need both Re and Im clipping routines to be done on the same data array.

matyasmolnar: Yes `sigma_clip` is brittle - I'll change it to get rid of the `axis` arg and also the `flag` arg (see below comment).

hera_cal/lstbin.py
matyasmolnar: I'm not sure I follow? If `array[clip] *= np.nan` is left in the `sigma_clip` function, then the data array is modified in place and the second sigma_clip call will see a different data array.

matyasmolnar:
@nkern thanks for having a look! I've replied to your comment and pushed another commit to make `sigma_clip` simpler and more memory efficient (but therefore very specific to the LST-binning done). This does, however, also involve remove a few lines from the sigma_clip tests, which I hope you will find acceptable.

I've spent a while looking at the LST-binning code and when I was reviewing the sigma clipping, I found that the current implementation had a few issues, notably:
 - `min_N` was only looking at the axis dimension, and from what I understand we do in fact want this to be the minimum number of unflagged data points considered. Currently, even if there are only e.g. 3 unflagged data points, sigma clipping still occurs for that LST-bin/frequency. I.e. `min_N` currently isn't doing anything.
 - ordering of clipping for Re and Im components mattered since `sigma_clip` modified the input data array in place, which is problematic for reasons explained above.
 
 This PR just attempts to fixes these issues. I'm very happy to have a Zoom to better explain this PR if required.

hera_cal/lstbin.py
nkern: thanks for updating the axis arg.

about the ordering, here is how this looks on master

```
# clip real
real_f = sigma_clip(d.real, sigma=sigma, min_N=min_N, axis=0)

# clip imag
imag_f = sigma_clip(d.imag, sigma=sigma, min_N=min_N, axis=0)

# clip flags
clip_flags = real_f + imag_f
```
in this case, even though d.real and d.imag are updated in place, I don't think that changes their respective flagging. notice that we don't input flags into `sigma_clip`.

lmk if i'm missing something here, thanks

matyasmolnar: Ah, I see.. yes you're right.

I thought for some reason that `array[clip] *= np.nan` in `sigma_clip` was `nan`'ing both components of the `d` array, but in fact, it only operates on the specified component...

Just tried it for
```
a = np.ones(10, dtype=complex)
a.real[5] *= np.nan
a
```
and you do get
```
array([ 1.+0.j,  1.+0.j,  1.+0.j,  1.+0.j,  1.+0.j, nan+0.j,  1.+0.j, 1.+0.j,  1.+0.j,  1.+0.j])
```

so if Re is done first and some Re points are clipped, the data array is modified for Re only but the Im points that were originally unflagged before `sigma_clip(d.real, ...)` are still all there for `sigma_clip(d.imag, ...)`

apologies for this, understood now

hera_cal/lstbin.py
matyasmolnar: Ah yes, sorry just got my head around this. Thanks for explaining!

hera_cal/lstbin.py
nkern: no worries! its always refreshing to get new eyes on the code--you've certainly caught a fair share of bugs here :).

so from a code perspective, could we revert changes back to master that attempted to address this?

matyasmolnar: @nkern from a code perspective, I think the PR is still good to go, as the `array[clip] *= np.nan` in `sigma_clip` was unnecessary since `d[clip_flags] *= np.nan` is done on the outside after the clip flags are found. What do you think?

nkern: so yes, I agree that b/c we do `array[clip_flags] *= np.nan` in the lst binner that technically the inplace operation in `sigma_clip` is irrelevant, so I'm fine w/ that.

I see you deprecated the `axis` kwarg, which I'm fine with, however, in deprecating the `flags` input I think you removed a functionality that (while we currently do not actually use) could be useful in the future. namely, in preventing the `location` variable  in `sigma_clip` from being influenced by known bad nights, for example.
now b/c the flags operation edits `array` in place, I think we might as well just keep the in-place operations in `sigma_clip` for consistency.

nkern: ah you know what I see this is actually done in the lst binner before the `sig_clip` if statement, so nvm, I'm actually fine w/ you deprecating both the inplace operations and `flags` and `axis` kwargs in `sip_clip`, thanks!

nkern:
@matyasmolnar I think this looks good! I'm fine with the changes and even though I have some reservations about applying sigma_clip to some freqs and not others (not exactly sure if that will produce spectral structure or not), this will only be a concern when min_N is much larger than what we traditionally use in the lst binner anyways, and we can always just go back and flag pixels that don't meet the min_N criteria after the fact.

nkern (approved):

jsdillon (changes requested):
Glad you two converged. I have one request before we pull this in: 

I'd like to add an option (the default can be `False`) where if the number of unflagged visibilities going into a sigma clipping step is less than `min_N` instead of no sigma clipping getting applied, instead the entire visibility for all JDs gets flagged. Basically, I think the user should get to say "if there are so few unflagged JDs at this integration and channel that we can't reliably perform sigma clipping, let's just throw out the whole pixel in the waterfall."  Something like `flag_below_min_N`.

nkern:
@jsdillon I was going to suggest the same thing, but I thought it would be easy enough to do that after the fact ourselves without having to expose another kwarg, but if you think its warranted then I'm fine w/ doing that.

@matyasmolnar if you add that kwarg to sigma_clip, can you also propagate it to `lst_bin` function and its argparser as well? lmk if that makes sense, thanks!

matyasmolnar:
@jsdillon @nkern 

I have now added a `flag_below_min_N` kwarg but I left the code for this outside of the `sigma_clip` function, for efficiency (it would be repeated twice if put inside `sigma_clip` and flags would also have to be propagated [or `np.isnan` used]). Note that the `flag_below_min_N` can only be activated if `sig_clip` is True - hope this is fine.

Let me know if all okay!

hera_cal/lstbin.py
```diff
@@ -336,6 +337,12 @@ def lst_bin(data_list, lst_list, flags_list=None, nsamples_list=None, dlst=None,
 
             # sigma clip if desired
             if sig_clip:
+                # flag frequency slices with fewer than min_N data points
+                if flag_below_min_N:
+                    below_min_N_freqs = np.where(f.sum(axis=0) < min_N)[0]
+                    if below_min_N_freqs.size > 0:
+                        d[:, below_min_N_freqs] *= np.nan
```
jsdillon, line 344: Shouldn't you also be modifying `f` like `f[:, below_min_N_freqs] = True`?

jsdillon (commented):
Looking good! Thanks so much for adding this feature. Just a couple of minor things.

hera_cal/lstbin.py
matyasmolnar: Oh yes, I'll add this now

hera_cal/lstbin.py
```diff
@@ -608,8 +616,8 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, verb
 
 def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_per_file=60,
                   file_ext="{type}.{time:7.5f}.uvh5", outdir=None, overwrite=False, history='', lst_start=None,
-                  atol=1e-6, sig_clip=True, sigma=5.0, min_N=5, rephase=False, output_file_select=None,
-                  Nbls_to_load=None, ignore_flags=False, average_redundant_baselines=False,
+                  atol=1e-6, sig_clip=True, sigma=5.0, min_N=5, flag_below_min_N=False, rephase=False,
```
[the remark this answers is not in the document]
matyasmolnar: In the `lst_bin_files` doctstring, other kwargs like `sigma`, `min_N` etc. are not included? I would have thought that `bin_kwargs` summarises them? Have left `flag_below_min_N ` out for this reason, but can include if you wish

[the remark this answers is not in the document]
jsdillon: Ah, I see. Nevermind, it's fine as is.

jsdillon (approved):

matyasmolnar:
Hi @jsdillon @nkern 

On revision, looks like `flag_below_min_N` would serve the same purpose as `flag_thresh`? Still worth keeping or should I remove?

jsdillon:
They are similar, but not the same. I think we should keep both.

nkern (approved):

jsdillon:
@matyasmolnar I think we're good here. Shall I merge?

matyasmolnar:
> @matyasmolnar I think we're good here. Shall I merge?

Yes, all good to merge!

PR #755, opened 2022-01-14
make sure that noise auto correlations are real in vis_clean tests.

aewallwi:
This PR fixes lines in several `vis_clean` unittests where unphysical imaginary autocorrelations were generated as test data. For the autos, only the real parts of this noise are retained.


bhazelton (approved):
Looks good, thanks for fixing this!

PR #757, opened 2022-01-31
Fix match_times bug

aewallwi:
As written, match_times does not successfully apply phase wraps if a model file contains LST=0 cross somewhere in the middle. This PR fixes this.


aewallwi:
> 

Done.

PR #758, opened 2022-02-02
Address edge case where model file can have some (but not all) lsts that are less then model_lsts[0]

aewallwi:


aewallwi (commented):
A beautifully written piece of code 🥇 🥇 . I'd approve it if I could 💯 💯 .

jsdillon (approved):

PR #759, opened 2022-02-08
fix another lst matching bug.

aewallwi:
Fix bug where the `data_lsts` in `abscal.match_times` can end up being non-monotonically increasing which will break the model matching routine.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -2029,7 +2029,8 @@ def match_times(datafile, modelfiles, filetype='uvh5', atol=1e-5):
 
     # shift data relative to model if needed
     data_lsts[data_lsts < model_starts[0]] += 2 * np.pi
-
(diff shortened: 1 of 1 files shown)


hera_cal/abscal.py
```diff
@@ -2029,7 +2029,8 @@ def match_times(datafile, modelfiles, filetype='uvh5', atol=1e-5):
 
     # shift data relative to model if needed
     data_lsts[data_lsts < model_starts[0]] += 2 * np.pi
-
+    # make sure monotonically increasing.
+    data_lsts[data_lsts < data_lsts[0]] += 2 * np.pi
```
jsdillon, line 2033: Doesn't np.unwrap do what you want here?

aewallwi: it does. I've changed this.

jsdillon (approved):

PR #760, opened 2022-02-15
Use `self.data_ants` instead of `self.ants` in ReflectionFitter

jsdillon:
Following from #691 and #692, this PR now updates the cable reflection fitter to loop through only antennas in the data, rather than antennas in the array. This is useful for running on down-selected data and for running and late-H4C (and onwards) data where the full array ends up in `HERAData.ants` but not in `HERAData.data_ants`.


nkern (approved):
looks good!

PR #761, opened 2022-02-17
Streamline I/O in `auto_reflection_run()`

jsdillon:
I've been experimenting with making reflection fitting part of a daily pipeline for a re-analysis of H1C IDR3 that I'm exploring. Given the paucity of evidence for substantial time variability in the reflections, I'm trying to produce a single reflection fit for the whole day. I tried implementing this with `auto_reflection_run.py` and ran into two issues, which this PR address:

1.  This function produced one calfits file per `dly_range` provided. If I'm fitting dozens of reflections, that produces a lot of files that I don't really want. If I want to go back and reconstruct all the individual reflections, that's what the npzs are good for. So this PR implements a `write_each_calfits` option, which is default True (so no change in the default behavior) but when False will only write a final combined calfits file. If one turns on `write_npz`, that's still done per delay range.

2. When using `time_avg=True` its a bit silly to produce a calfits file with the full-temporal resolution when a single-integration calfits file will do (especially considering that I'm now handling this case explicitly in `apply_cal`--see #756). In my case, since I was doing a whole day, each calfits file was 11 GB! And there were dozens of them! So now, when `time_avg` is true, RF.times is reduced to a length-1 vector so that gains are not broadcast.

### DIFF
--- a/hera_cal/reflections.py
+++ b/hera_cal/reflections.py
@@ -1446,13 +1446,13 @@ def auto_reflection_run(data, dly_ranges, output_fname, filetype='uvh5', input_c
         The CLEAN min_dly should always be less than the lower boundary of dly_range.
     """
     # dly_ranges type check
-    if isinstance(dly_ranges, (str, np.str)):
+    if isinstance(dly_ranges, str):
         dly_ranges = [ast.literal_eval(dly_ranges)]
     if isinstance(dly_ranges, tuple):
         dly_ranges = [dly_ranges]
     if isinstance(dly_ranges, list):
         for i, dlyr in enumerate(dly_ranges):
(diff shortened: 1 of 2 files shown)


nkern:
@jsdillon this looks fine I think but it I do vaguely recall a reason for writing the full time resolution out even with time-average, but I forget now why that was the case. you've almost nearly made this backwards compatible (by allowing write_calfits=True by default)--could you make this fully backwards compatible by just allowing an option to compress the time length to 1, which is set to False by default? I don't think i have any other concerns other than this, thanks

jsdillon:
Thanks @nkern. I've made that parameter optional and this is ready for another review.

nkern (approved):

PR #762, opened 2022-02-23
Allow argparser to set max_contiguous_edge_flags in delay_filter_run.py in DPSS mode

jsdillon:
For reasons I don't understand, `delay_filter_run.py` ignored `max_contiguous_edge_flags` from the argparser and just set it to 1 in `dpss_leastsq` mode. This opens that up, but shouldn't change the default behavior, since the default value is 1.


aewallwi (approved):
looks good too me.

PR #763, opened 2022-03-03
Option is use DPSS filters in smoothcal 2D

tyler-a-cox:
Minor edits to `smoothcal` will allow the option to run `smoothcal.time_freq_2D_filter` using 2D DPSS filters rather than the traditional CLEAN method. The DPSS smoothing should improve smoothing of calibration solutions at the band edges.


tyler-a-cox:
Here's a plot of the `DPSS` and `CLEAN` `smoothcal` median gain amplitude plotted against the `abscal` median gain amplitude for two antennas on JD 2458187 (similar in style to Figure 3 in `smooth_cal_inspect_2458187.ipynb`). The `DPSS` method does appear to better match the amplitude of the `abscal` gains, particularly at the band edges.

Screen Shot 2022-03-21 at 12 23 22 PM

hera_cal/smooth_cal.py
```diff
@@ -67,6 +68,84 @@ def single_iterative_fft_dly(gains, wgts, freqs, conv_crit=1e-5, maxiter=100):
...
+        weights: ndarray of shape (Ntimes, Nfreqs) of calibration flags
+        design_matrix: ndarray of shape (Ntimes * Nfreqs, N_frequency_vectors * N_time_vectors)
+            obtained from hera_cal.smooth_cal.dpss_filters
+
+    Returns:
+        S: ndarray of shape (N_frequency_vectors * N_time_vectors, )
+    """
+    cmat = (design_matrix.T * weights.reshape(-1)) @ design_matrix
+
+    if np.linalg.cond(cmat) >= 1e9:
```
jsdillon, line 139: No magic numbers. This should be a default kwarg.

jsdillon: Also, computing the condition number can be expensive. Is doing this check significantly increasing runtime? Why not just always do pinv?

hera_cal/smooth_cal.py
```diff
@@ -67,6 +68,84 @@ def single_iterative_fft_dly(gains, wgts, freqs, conv_crit=1e-5, maxiter=100):
...
+        warnings.warn("Warning! Poorly conditioned matrix")
+        smat = np.linalg.pinv(cmat) @ design_matrix.T * weights.reshape(-1)
```
jsdillon, line 141: Whatever condition number cutoff you do, it should be the one used in `np.linalg.pinv(cmat, rcond=max_rcond)`

hera_cal/smooth_cal.py
```diff
@@ -67,6 +68,84 @@ def single_iterative_fft_dly(gains, wgts, freqs, conv_crit=1e-5, maxiter=100):
...
+        design_matrix: ndarray of shape (Ntimes * Nfreqs, N_frequency_vectors * N_time_vectors)
+            obtained from hera_cal.smooth_cal.dpss_filters
+
+    Returns:
+        S: ndarray of shape (N_frequency_vectors * N_time_vectors, )
+    """
+    cmat = (design_matrix.T * weights.reshape(-1)) @ design_matrix
+
+    if np.linalg.cond(cmat) >= 1e9:
+        warnings.warn("Warning! Poorly conditioned matrix")
```
jsdillon, line 140: I'd suggest printing out what the condition number is as part of the warning. Also, mention that you're doing a pinv and what rcond_max is. In general, warnings should be as informative as possible!

hera_cal/smooth_cal.py
```diff
@@ -67,6 +68,84 @@ def single_iterative_fft_dly(gains, wgts, freqs, conv_crit=1e-5, maxiter=100):
...
+            Note that freq_scale is in MHz while freqs is in Hz.
+        time_scale: time scale in seconds. Defined analogously to freq_scale.
+            Note that time_scale is in seconds, times is in days.
+        eigenval_cutoff: sinc_matrix eigenvalue cutoff to use for included dpss modes.
+            Only used when the filtering method is 'DPSS'
+        return_1D_filters: if True, return the 1D filters that make up the 2D DPSS filter.
+            Default is False
+
+    Returns:
+        filters: DPSS filtering vectors, ndarray of size (Ntimes * Nfreqs, N_frequency_vectors * N_time_vectors)
```
jsdillon, line 90: Explain how the output changes if return_1D_filters is True

hera_cal/smooth_cal.py
```diff
@@ -182,15 +263,28 @@ def time_freq_2D_filter(gains, wgts, freqs, times, freq_scale=10.0, time_scale=1
...
+        method: Algorithm used to smooth calibration solutions. Either 'CLEAN' or 'DPSS':
+            'CLEAN': uses the CLEAN algorithm to
+            'DPSS': uses discrete prolate spheroidal sequences (DPSS) to filter calibration solutions
+        design_matrix: Matrix of DPSS filters calculated using smooth_cal.dpss_filters. If matrix is not provided,
+            one will be calculated using smooth_cal.dpss_filters and the time and frequency scale. Only used
+            when the method used is 'DPSS'
+        sol_matrix: Solution matrix obtained from performing linear least squares with 2D DPSS vectors
+        eigenval_cutoff: sinc_matrix eigenvalue cutoffs to use for included dpss modes.
+            Only used when the filtering method is 'DPSS'
+        skip_flagged_edges : bool, optional
```
jsdillon, line 278: Do we generally want to use this when DPSS is true? If so, then perhaps this should be default True (but still ignored if CLEAN is True)

hera_cal/smooth_cal.py
```diff
@@ -182,15 +263,28 @@ def time_freq_2D_filter(gains, wgts, freqs, times, freq_scale=10.0, time_scale=1
                     and fringe rate = 0
             'plus': produce a separable calibration solution by only keeping modes with 0 delay,
                     0 fringe rate, or both
-        window: window function for filtering applied to the filtered axis.
+        window: window function for filtering applied to the filtered axis. Only used with when method used is 'CLEAN'.
             See aipy.dsp.gen_window for options.
         maxiter: Maximum number of iterations for aipy.deconv.clean to converge.
+        method: Algorithm used to smooth calibration solutions. Either 'CLEAN' or 'DPSS':
+            'CLEAN': uses the CLEAN algorithm to
```
jsdillon, line 270: Unfinished sentence

hera_cal/smooth_cal.py
```diff
@@ -182,15 +263,28 @@ def time_freq_2D_filter(gains, wgts, freqs, times, freq_scale=10.0, time_scale=1
                     and fringe rate = 0
             'plus': produce a separable calibration solution by only keeping modes with 0 delay,
```
jsdillon, line 264: Explain that filter_mode `plus` is not implemented for DPSS.

scripts/smooth_cal_run.py
```diff
@@ -25,7 +25,7 @@
                              chan_blacklists=a.chan_blacklists, pick_refant=a.pick_refant, freq_threshold=a.freq_threshold,
                              time_threshold=a.time_threshold, ant_threshold=a.ant_threshold, verbose=a.verbose)
     cs.time_freq_2D_filter(freq_scale=a.freq_scale, time_scale=a.time_scale, tol=a.tol,
-                           filter_mode=a.filter_mode, window=a.window, maxiter=a.maxiter, **win_kwargs)
+                           filter_mode=a.filter_mode, window=a.window, maxiter=a.maxiter, method=a.method, **win_kwargs)
```
jsdillon, line 28: You are missing eigenval_cutoff and skip_flagged_edges

hera_cal/smooth_cal.py
```diff
@@ -682,21 +824,72 @@ def time_freq_2D_filter(self, freq_scale=10.0, time_scale=1800.0, tol=1e-09,
...
+                Only used when the filtering method is 'DPSS'
+            method: Algorithm used to smooth calibration solutions. Either 'CLEAN' or 'DPSS':
+                'CLEAN': uses the CLEAN algorithm to
+                'DPSS': uses discrete prolate spheroidal sequences to filter calibration solutions
+            skip_flagged_edges : if True, do not filter over flagged edge times (filter over sub-region)
+                default is False, only used when method='DPSS'
             win_kwargs : any keyword arguments for the window function selection in aipy.dsp.gen_window.
                 Currently, the only window that takes a kwarg is the tukey window with a alpha=0.5 default
         '''
+        if method == 'DPSS':
```
jsdillon, line 840: Could you avoid doing all this by just doing what you do for sol_matrix (`sol_matrix = info['sol_matrix'] if np.allclose(wgts_grid, wgts_prev) else None`) for the design matrix too?

hera_cal/smooth_cal.py
```diff
@@ -667,8 +808,9 @@ def freq_filter(self, filter_scale=10.0, tol=1e-09, window='tukey', skip_wgt=0.1
                     self.flag_grids[ant][i, :] = np.ones_like(self.flag_grids[ant][i, :])
         self.rephase_to_refant(warn=False)
 
-    def time_freq_2D_filter(self, freq_scale=10.0, time_scale=1800.0, tol=1e-09,
-                            filter_mode='rect', window='tukey', maxiter=100, **win_kwargs):
+    def time_freq_2D_filter(self, freq_scale=10.0, time_scale=1800.0, tol=1e-09, filter_mode='rect',
```
jsdillon, line 811: In general, this function doesn't have enough inline comments to explain what's going on.

hera_cal/tests/test_smooth_cal.py
```diff
@@ -144,6 +180,7 @@ def test_time_freq_2D_filter(self):
         np.testing.assert_array_almost_equal(ff, np.ones((100, 100), dtype=complex))
         ff, info = smooth_cal.time_freq_2D_filter(gains, wgts, freqs, times, filter_mode='plus')
         np.testing.assert_array_almost_equal(ff, np.ones((100, 100), dtype=complex))
+        ff, info = smooth_cal.time_freq_2D_filter(gains, wgts, freqs, times, method='DPSS')
```
jsdillon, line 183: Can you add another `np.testing.assert_array_almost_equal(ff, np.ones((100, 100), dtype=complex))` after DPSS?

jsdillon (changes requested):
Nice work @tyler-a-cox! 

I've got a handful of suggested changes, but this looks pretty close. I'd in addition to the spectral results you showed (which look great!), I'd love to see temporal results and 2D results in amplitude and phase. Kinda like figures 3 and 4 in this notebook: https://nbviewer.org/github/HERA-Team/H1C_IDR3_Notebooks/blob/main/smooth_cal_inspect/smooth_cal_inspect_2458098.ipynb

hera_cal/smooth_cal.py
aewallwi: > No magic numbers. This should be a default kwarg.

"magic numbers" makes me think of this because there are numbers and wizards

image

aewallwi (changes requested):
Looks good for the most part. A few minor changes requested.

jsdillon:
OMG, I totally forgot about that book. A classic!

I'm actually more reminded of https://www.youtube.com/watch?v=8BqnN72OlqA

hera_cal/smooth_cal.py
tyler-a-cox: I updated this to just perform the pinv instead of checking the condition number. This function is a slightly modified version of a function in `uvtools` which is where the `1e9` came from. We might want to make some updates there as well

hera_cal/smooth_cal.py
tyler-a-cox: See comment above

hera_cal/smooth_cal.py
tyler-a-cox: Yeah, I think in most cases we probably would want to. I set the default to True

hera_cal/smooth_cal.py
tyler-a-cox: Fixed this

hera_cal/smooth_cal.py
tyler-a-cox: Added this

scripts/smooth_cal_run.py
tyler-a-cox: Added these

hera_cal/smooth_cal.py
```diff
@@ -67,6 +68,84 @@ def single_iterative_fft_dly(gains, wgts, freqs, conv_crit=1e-5, maxiter=100):
     return np.sum(taus)
 
 
+def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1e-9, return_1D_filters=False):
```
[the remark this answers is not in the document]
tyler-a-cox: Sounds good! I opened up an issue in `uvtools`. I can work on moving this over later.

hera_cal/smooth_cal.py
```diff
@@ -182,15 +263,28 @@ def time_freq_2D_filter(gains, wgts, freqs, times, freq_scale=10.0, time_scale=1
...
         win_kwargs : any keyword arguments for the window function selection in aipy.dsp.gen_window.
             Currently, the only window that takes a kwarg is the tukey window with a alpha=0.5 default.
 
     Returns:
         filtered: filtered gains, ndarray of shape=(Ntimes,Nfreqs)
-        info: dictionary of metadata from aipy.deconv.clean
+        info: dictionary of metadata from aipy.deconv.clean or DPSS, depending on the filter method
+            chosen.
     '''
     assert AIPY, "You need aipy to use this function"
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed this,

hera_cal/smooth_cal.py
tyler-a-cox: I added more comments and simplified the function. Let me know if you'd like for me to add more!

hera_cal/smooth_cal.py
tyler-a-cox: I decided to simplify the function a little. If you're using `skip_flagged_edges`, your weights/gain grid may come out to be different sizes depending on the flagging pattern, which would mean you would need to recalculate the DPSS vectors to fit the new size of the gain grid. Calculating the DPSS filters can start to take a while as you get larger gain grid sizes, so I was using this as a way of keeping track of the grid sizes we've already calculated the DPSS vectors for. There might be a way to better optimize this in the future, but for now, I've just removed it for simplicity and readability.

hera_cal/smooth_cal.py
tyler-a-cox: I ended up removing this when I simplified the `CalbrationSmoother.time_freq_2d_filter` function

tyler-a-cox:
Here are a couple of plots that roughly match Figure 3 and Figure 4 in https://nbviewer.org/github/HERA-Team/H1C_IDR3_Notebooks/blob/main/smooth_cal_inspect/smooth_cal_inspect_2458187.ipynb. I only DPSS `smoothcal` on 20 files in `JD2458187` so that I could run the test quickly in a notebook. If you'd like to see the results on a whole night's worth of data let me know. They do seem to watch well though!

figure_3
figure_4

hera_cal/tests/test_smooth_cal.py
tyler-a-cox: Added this

hera_cal/smooth_cal.py
tyler-a-cox: Decided to simplify this function and just perform the pinv

jsdillon (approved):
This looks good to me! Nice work @tyler-a-cox. I'm excited to try this out. If @aewallwi is good to go, let's merge!

aewallwi (approved):
Thanks for all the great work on this @tyler-a-cox . I'm happy to see this merged now.

PR #764, opened 2022-03-05
Make tophat_frfilter use baseline east-west magnitude, treating EW and WE baselines the same

jsdillon:
Right now tophat_frfilter breaks when using `max_frate_coeffs` when the baselines have negative x components. Adding the absolute value makes the function identical to what was done in the preprocessing pipelines: 
https://github.com/HERA-Team/hera_pipelines/blob/8c6e343f3bbfab6c17a4cec488d15af386b7bec0/pipelines/h1c/idr3/v2/preprocess/preprocess_data.py#L913

I also fixed the indentation in the docstring to make it more readable.


aewallwi (approved):

aewallwi:
looks good to me.

PR #768, opened 2022-03-17
Allow weights to be set externally in tophat_frfilter

jsdillon:
Despite what the documentation said, wgts were always ~self.flags, regardless of what was passed in.


jsdillon:
This was addressed in #736, so I'm closing this in favor of that

PR #770, opened 2022-03-31
script to transfer calibration flags.

aewallwi:
Command line script to transfer calibration flags


aewallwi:
I've decided that this script is no longer necessary and have closed this request.

PR #771, opened 2022-04-02
Model calibration

aewallwi:
Add driver function and bash script for model based calibration.

PR #772, opened 2022-04-04
One dimensional freq filter

aewallwi:
Command line driver scripts for frequency domain smoothcal using dpss modes. Depends on #763 and #771 .


tyler-a-cox (approved):
This looks good to me! We should probably just remove the mode option "DPSS" at some point and make it match `uvtools` like you did with "dpss_leastsq", but this works for me.

PR #765, opened 2022-03-05
Allow explicitly setting of minimum and maximum FRF half-widths, regardless of baseline length or mode.

jsdillon:
This PR enables explicit setting of minimum FRs half-widths when also using max_frate_coeffs. It also enables explicitly setting maximum FR half-widths in either mode in order to match functionality in the H1C IDR2/3 preprocessing pipeline:
https://github.com/HERA-Team/hera_pipelines/blob/8c6e343f3bbfab6c17a4cec488d15af386b7bec0/pipelines/h1c/idr3/v2/preprocess/preprocess_data.py#L914

This PR depends on #764

PR #773, opened 2022-04-06
Restore feature to allow a global maximum fringe rate filter half width

jsdillon:
This PR restores a feature from #765 that got lost in the shuffle of #764


aewallwi (approved):
Looks good to me. I've approved.

PR #774, opened 2022-04-06
Remove python 3.7 from unit testing and add 3.10

jsdillon:
astropy doesn't support 3.7 anymore, so pyuvdata doesn't either

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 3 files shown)


bhazelton:
I'd suggest adding python 3.10. Note that you'll need to put it in quotes (as "3.10") in your yaml or it will try to grab python 3.1 instead.

jsdillon:
> I'd suggest adding python 3.10. Note that you'll need to put it in quotes (as "3.10") in your yaml or it will try to grab python 3.1 instead.

Done, @bhazelton

bhazelton (approved):
Looks good to me!

PR #775, opened 2022-04-07
Lower tolerance in redcal + abscal test

jsdillon:
Something is causing tests to just barely fail on some python versions. Maybe an updated library? Probably not worth digging into, so I'll just lower the threshold a bit.


nkern:
@jsdillon don't want to create work for ppl, but we have gotten burned by changing unit-tests before. I'm fine changing the tolerances slightly for now, but could we at least document here in this thread which python versions, which tests failed and by how much?

bhazelton:
@nkern It looks like it's failing on python 3.7, 3.8, 3.9 on linux but not on mac. See the checks on #774

bhazelton (approved):
Looks reasonable.

jsdillon:
Good news, @nkern. If I give redcal and abscal more iterations and tighter convergence criteria, I can get back to the previous level of accuracy.

So this PR has been undone by #774

PR #776, opened 2022-04-07
make sure that kwargs are passed.

aewallwi:
ensure that some kwargs that were being dropped by frf updates are in fact passed correctly.


jsdillon:
Tests are failing...

aewallwi:
should be fixed now.

jsdillon:
Looks like there are some merge conflicts.

Also, to be clear, are we eliminating these arguments or just grouping them in `**kwargs`?

aewallwi:
> 

That's correct, we are moving explicit args into kwargs when possible (to avoid confusion). I think all of the conflicts should be taken care of now.

jsdillon (approved):
Looks good to me.

PR #777, opened 2022-04-07
Fix deprecated numpy types

jsdillon:
hera_cal tests are getting a bajillion warnings for numpy types that are deprecated aliases to built-in types. This PR hopefully gets them all.

In general, we've got a huge backlog of warnings that should be looked into, but this seemed like an easy place to start.


hera_cal/reflections.py
```diff
@@ -1287,14 +1287,14 @@ def L(x, amp, dly, phs, clean_data, clean_model, clean_wgts, dly_range, freqs, f
...
-        clean_wgts = (~clean_flags).astype(np.float)
+        clean_wgts = (~clean_flags).astype(float)
     edge_flags = np.isclose(np.mean(clean_wgts, axis=0), 0.0)
     edge_flags[np.argmin(edge_flags):-np.argmin(edge_flags[::-1])] = False
     if fix_amp and fix_dly and fix_phs:
         raise ValueError("Can't hold amp, dly and phs fixed")
 
     # create delay range
-    if isinstance(dly_range, (int, np.int, float, np.float)):
+    if isinstance(dly_range, (int, int, float, float)):
```
r-pascua, line 1297: I'm not sure if this is what you want to do? Above, I see something like `if isinstance(x, (int, np.integer, float, np.floating))`, which is what I imagine you're going for. One thing I'm wondering is whether you can get away with just doing `if isinstance(x, (int, float))`.

hera_cal/tempcal.py
```diff
@@ -81,10 +81,10 @@ def gains_from_autos(data, times, flags=None, smooth_frate=1.0, nl=1e-10,
     # edgeflag
     if edgeflag is not None:
         if flags is None:
-            flags = np.zeros_like(data, np.bool)
+            flags = np.zeros_like(data, bool)
         else:
             flags = copy.deepcopy(flags)
-        if isinstance(edgeflag, (int, np.int, float, np.float)):
+        if isinstance(edgeflag, (int, int, float, float)):
```
r-pascua, line 87: Same as other comment.

hera_cal/tests/test_flag_utils.py
```diff
@@ -21,7 +21,7 @@ def test_solar_flag():
     data, flags, antp, ant, f, t, l, p = io.load_vis(uvd, return_meta=True)
     # get solar altitude
     a = utils.get_sun_alt(2458043)
-    assert isinstance(a, (float, np.float, np.float64))
+    assert isinstance(a, (float, float, np.float64))
```
r-pascua, line 24: Essentially the same comment as above.

hera_cal/tests/test_io.py
```diff
@@ -1111,8 +1111,8 @@ def test_get_file_times():
 
     # test if fed as a str
     dlsts, dtimes, larrs, tarrs = io.get_file_times(filepaths[0], filetype='miriad')
-    assert isinstance(dlsts, (float, np.float))
-    assert isinstance(dtimes, (float, np.float))
+    assert isinstance(dlsts, (float, float))
+    assert isinstance(dtimes, (float, float))
```
r-pascua, line 1115: Again

hera_cal/tests/test_reflections.py
```diff
@@ -57,14 +57,14 @@ def noise(n, sig):
...
     # form cable gains
     if add_cable:
-        if isinstance(cdelay, (float, np.float, int, np.int)):
+        if isinstance(cdelay, (float, float, int, int)):
             cdelay = [cdelay]
-        if isinstance(camp, (float, np.float, int, np.int)):
+        if isinstance(camp, (float, float, int, int)):
             camp = [camp]
-        if isinstance(cphase, (float, np.float, int, np.int)):
+        if isinstance(cphase, (float, float, int, int)):
```
r-pascua, line 64: Here as well.

hera_cal/utils.py
```diff
@@ -674,7 +674,7 @@ def get_sun_alt(jds, latitude=-30.721526120689507, longitude=21.428303826863015)
     """
     # type check
     array = True
-    if isinstance(jds, (float, np.float, np.float64, int, np.int, np.int32)):
+    if isinstance(jds, (float, float, np.float64, int, int, np.int32)):
```
r-pascua, line 677: Also here.

hera_cal/vis_clean.py
```diff
@@ -771,12 +771,12 @@ def vis_clean(self, keys=None, x=None, data=None, flags=None, wgts=None,
...
             keys = data.keys()
         if wgts is None:
-            wgts = DataContainer(dict([(k, np.ones_like(flags[k], dtype=np.float)) for k in keys]))
+            wgts = DataContainer(dict([(k, np.ones_like(flags[k], dtype=float)) for k in keys]))
         # make sure flagged channels have zero weight, regardless of what user supplied.
         wgts = DataContainer(dict([(k, (~flags[k]).astype(float) * wgts[k]) for k in keys]))
         # convert max_frate to DataContainer
         if max_frate is not None:
-            if isinstance(max_frate, (int, np.integer, float, np.float)):
+            if isinstance(max_frate, (int, np.integer, float, float)):
```
r-pascua, line 779: and here

hera_cal/vis_clean.py
```diff
@@ -1106,7 +1106,7 @@ def fourier_filter(self, filter_centers, filter_half_widths, mode='clean',
                     xp[1] = xp[1][spw_slice]
                     if not isinstance(zeropad, (list, tuple)) or not len(zeropad) == 2:
                         raise ValueError("zeropad must be a 2-tuple or 2-list of integers")
-                    if not (isinstance(zeropad[0], (int, np.int)) and isinstance(zeropad[0], (int, np.int))):
+                    if not (isinstance(zeropad[0], (int, int)) and isinstance(zeropad[0], (int, int))):
```
r-pascua, line 1109: Here as well. Looks like the second item should index the first element of `zeropad`, too. You could also do a comprehension: `if not all (isinstance(x, int) for x in zeropad)`.

r-pascua (changes requested):
For the most part, it looks like the changes are good. There are some spots where things should be changed, though&mdash;I left comments pointing these out. These all come up when checking the `type` of an object, so it's probably worthwhile to test whether `isinstance(x, (int, float))` gives the same result as `isinstance(x, (int, np.integer, float, np.floating))` for all scalar `x`, since you can simplify the statements if this is true.

hera_cal/reflections.py
jsdillon: Good catch. We want np.floating and np.integer to catch other float/int types like float32 or whatever. Fixed.

hera_cal/tempcal.py
jsdillon: Fixed.

hera_cal/tests/test_flag_utils.py
jsdillon: Fixed.

hera_cal/tests/test_reflections.py
jsdillon: fixed

hera_cal/tests/test_io.py
jsdillon: fixed

hera_cal/utils.py
jsdillon: fixed

hera_cal/vis_clean.py
jsdillon: fixed

hera_cal/vis_clean.py
jsdillon: Good catch. Fixed.

r-pascua (approved):
Looks like everything has been addressed! Thanks!

PR #778, opened 2022-04-08
Expand options for weights in load_tophat_frfilter_and_write and tophat_frfilter_run

jsdillon:
This PR adds two new options to load_tophat_frfilter_and_write and the associated script that let the user better control the weights that go FRF. These allow weighting proportional to nsamples (not just flags) and allow certain LSTs to be assigned zero weight (useful to downweighting the galaxy).

These options allow us to reproduce the current functionality for xtalk subtraction in the preprocessing pipeline for H1C IDR3 with DPSS filtering.


hera_cal/frf.py
```diff
@@ -1289,7 +1293,7 @@ def tophat_frfilter_argparser(mode='clean'):
 def load_tophat_frfilter_and_write(datafile_list, case, baseline_list=None, calfile_list=None,
                                    Nbls_per_load=None, spw_range=None, cache_dir=None,
                                    read_cache=False, write_cache=False, external_flags=None,
-                                   factorize_flags=False, time_thresh=0.05,
+                                   factorize_flags=False, time_thresh=0.05, wgt_by_nsample=False, excluded_lsts=[],
```
r-pascua, line 1296: If the `excluded_lsts` parameter is updated anywhere in the function, then I think it's better to set the default to `None` and have something like `excluded_lsts = excluded_lsts or []` in the code. (Unless the optional parameter is a cache, default values should be immutable objects. You won't have any issues if the code doesn't modify the parameter, but it can lead to some interesting bugs if the code does modify it.)

hera_cal/tests/test_frf.py
```diff
@@ -580,6 +580,50 @@ def test_load_tophat_frfilter_and_write(self, tmpdir):
...
+
+        wgts = datacontainer.DataContainer({k: (~frfil.flags[k]).astype(float) for k in frfil.flags})
+        for k in wgts:
+            wgts[k][0:5, :] = 0
+            wgts[k][-5:, :] = 0
+        frate_centers, frate_half_widths = frf.select_tophat_frates(case='sky', uvd=frfil.hd, blvecs=frfil.blvecs)
+        frfil.tophat_frfilter(keys=[(53, 54, 'ee')], tol=1e-4, verbose=True, frate_centers=frate_centers, frate_half_widths=frate_half_widths, wgts=wgts)
+        np.testing.assert_almost_equal(d[(53, 54, 'ee')], frfil.clean_resid[(53, 54, 'ee')], decimal=5)
+        np.testing.assert_array_equal(f[(53, 54, 'ee')], frfil.flags[(53, 54, 'ee')])
+
```
r-pascua, line 626: This looks like it should be its own test. I'm not sure what the testing philosophy is for this package, so feel free to ignore this comment, but I think it's best to make tests as granular as possible.

r-pascua (approved):
Everything looks good to me! I just left a few comments regarding coding practice things, but I'll leave it up to you to decide whether to make any changes.

hera_cal/frf.py
jsdillon: Done.

hera_cal/tests/test_frf.py
jsdillon: This test already has a gazillion tests in it, so I'll just leave it for some future reorg.

PR #779, opened 2022-04-09
Only perform sigma clipping on real part of autocorrelations

jsdillon:
In the process of validation, I discovered that a huge fraction of the nsamples and were getting thrown away. I'm not 100% sure, but my theory is that any autocorrelations whose imaginary component is not identically 0 were often getting tossed by the sigma clipping (also this was done with an older version of pyuvdata before real autos were more strictly enforced). 

image

Regardless, we should not be separately sigma-clipping the imaginary parts of the autos. This PR fixes that.


jsdillon:
I'm re-running with this PR for epoch 1 and it looks like it fixes things. I still think the LST-binner could seriously use some overhaul and optimization, but this certainly helps things.

Before:
image

After:
image

nkern (approved):
looks good, thanks

PR #781, opened 2022-04-11
Make default skip_flagged_edges self-consistent in smooth_cal

jsdillon:
This PR makes the default skip_flagged_edges `True`, both in the argparser and script, so that the default value in `time_freq_2D_filter()` matches the default value in the script.


tyler-a-cox (approved):
This looks good to me! Thanks for catching that Josh

PR #782, opened 2022-04-20
Improve the runtime and memory usage efficiency of smoothcal with DPSS modes

tyler-a-cox:
Smoothcal with DPSS modes seems to perform better than `CLEAN`, particularly at the band edges but takes much longer to run than `CLEAN` (8 hours vs 1 hour) and uses more memory because it performs a linear fit using a large design matrix [(Nfreqs x Ntimes) x (N_freq_dpss_vec x N_time_dpss_vec)] which is the Kronecker product of two 1D DPSS vectors. The key to speeding up smoothcal and reducing memory usage is by never building the design matrix in memory. Instead, you can use the time and frequency 1D DPSS vectors and `np.einsum` to calculate `X^T W X` without explicitly storing `X` in memory. You can also use the properties of Kronecker products to simplify `X^T W y` reducing the total number of computations.

A simple test case shows that the updated version of smoothcal with DPSS modes running on a full night's worth of data for one antenna runs faster than `CLEAN`.

Screen Shot 2022-04-20 at 5 04 38 AM

The fit seems to still be good for this updated version:

linear_fit

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -104,7 +104,7 @@ def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1
         if Nmax < x0.shape[0]:
             Nw = 2 * x0.shape[0] * fw * np.diff(x0)[0]
             Nterms = int(min(10 * Nw, x0.shape[0]))
-            windows, eigvals = uvtools.dspec.windows.dpss(x0.shape[0], Nw // 2, Nterms, return_ratios=True)
+            windows, eigvals = uvtools.dspec.windows.dpss(x0.shape[0], Nw / 2, Nterms, return_ratios=True)
             windows = windows[eigvals > eigenval_cutoff].T
             dpss_windows.append(windows)
 
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -85,52 +85,85 @@ def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1
             Only used when the filtering method is 'DPSS'
 
     Returns:
-        filters: DPSS filtering vectors, ndarray of size (Ntimes * Nfreqs, N_frequency_vectors * N_time_vectors)
+        time_filters: DPSS filtering vectors along the time axis, ndarray of size (Ntimes, N_time_vectors)
+        freq_filters: DPSS filtering vectors along the frequency axis, (Nfreqs, N_freq_vectors)
     """
     delay_scale = (freq_scale * 1e6) ** -1  # Puts it in seconds
     fringe_scale = (time_scale) ** -1  # fringe scale in Hz
     time_in_seconds = (times - times.min()) * 60 * 60 * 24  # time array in seconds
 
-    freq_filters, nfreq_comps = uvtools.dspec.dpss_operator(
-        freqs,
-        filter_centers=[0],
-        filter_half_widths=[delay_scale],
-        eigenval_cutoff=[eigenval_cutoff],
-    )  # DPSS filters along the frequency-axis
-    freq_filters = freq_filters.real.astype(np.float32)
-
+    # Generate DPSS filters along the time-axis
     time_filters, ntime_comps = uvtools.dspec.dpss_operator(
         time_in_seconds,
         filter_centers=[0],
         filter_half_widths=[fringe_scale],
         eigenval_cutoff=[eigenval_cutoff],
-    )  # DPSS filters along the time-axis
-    time_filters = time_filters.real.astype(np.float32)
+    )
+    time_filters = time_filters.astype(np.float64)
 
-    filters = np.reshape(
-        time_filters[:, None, :, None] * freq_filters[None, :, None, :],
-        (times.shape[0] * freqs.shape[0], ntime_comps[0] * nfreq_comps[0]),
-    )  # 2D time-frequency DPSS vectors
-    return filters
+    # Generate DPSS filters along the frequency-axis
+    freq_filters, nfreq_comps = uvtools.dspec.dpss_operator(
+        freqs,
+        filter_centers=[0],
+        filter_half_widths=[delay_scale],
+        eigenval_cutoff=[eigenval_cutoff],
+    )
+    freq_filters = freq_filters.astype(np.float64)
+
+    return time_filters, freq_filters
 
 
-def fit_solution_matrix(weights, design_matrix):
+def solve_2D_DPSS(gains, weights, time_filters, freq_filters, XTXinv=None):
     """
-    Calculate the linear least squares solution matrix
-    from a design matrix, A and a weights matrix W
-    S = [A^T W A]^{-1} A^T W
+    Filters gain solutions by solving the weighted linear least squares problem
+    for a design matrix that can be factored into two kronecker products in the following way
+
+         y = X b = (A outer B) b
+
+    where A and B are time and frequency DPSS vector matrices. More memory and computationally
+    efficient than computing the design matrix from the time and frequency filters for
+    most input matrix sizes.
 
     Arguments:
+        gains: ndarray of shape=(Ntimes,Nfreqs) of complex calibration solutions to filter
         weights: ndarray of shape (Ntimes, Nfreqs) of calibration flags
-        design_matrix: ndarray of shape (Ntimes * Nfreqs, N_frequency_vectors * N_time_vectors)
+        time_filters: DPSS filtering vectors along the time axis, ndarray of size (Ntimes, N_time_vectors)
             obtained from hera_cal.smooth_cal.dpss_filters
+        freq_filters: DPSS filtering vectors along the frequency axis, (Nfreqs, N_freq_vectors)
+            obtained from hera_cal.smooth_cal.dpss_filters
+        XTXinv: Matrix of (X^T W X)^{-1} for the input DPSS filters and weights. Useful for filtering
+            many gain grids with similar flagging patterns. np.ndarray of shape (N_time_vectors * N_freq_vectors,
+            N_time_vectors * N_freq_vectors)
 
     Returns:
-        S: ndarray of shape (N_frequency_vectors * N_time_vectors, Ntimes * Nfreqs)
+        filtered: filtered gain grids from least squares fit, np.ndarray with the
+            same shape as the input gain grid
+        info: dictionary containing fit components and XTXinv
     """
-    cmat = (design_matrix.T * weights.reshape(-1)) @ design_matrix
-    smat = np.linalg.pinv(cmat) @ design_matrix.T * weights.reshape(-1)
-    return smat
+    # If (X^T W X)^-1 is not precalculated, calculate it
+    if XTXinv is None:
+        # Use einsum to calculate (X^T W X) in a memory efficient way
+        XTX = np.einsum(
+            "ij,kl,jl,jm,ln->ikmn", np.transpose(time_filters), np.transpose(freq_filters),
+            weights, time_filters, freq_filters, optimize=True
+        )
+        ncomps = time_filters.shape[1] * freq_filters.shape[1]
+        XTX = np.reshape(XTX, ncomps, ncomps)
+        XTXinv = np.linalg.pinv(XTX)
+
+    # Calculate X^T W y using the property (A \otimes B) vec(y) = (A Y B)
+    XTWy = np.ravel(np.transpose(time_filters) @ (gains * weights) @ freq_filters)
+
+    # Calculate beta
+    beta = np.reshape(XTWy @ XTXinv, (time_filters.shape[0], freq_filters.shape[1]))
+
+    # Produce an estimate of the filtered gains
+    filtered = time_filters @ beta @ np.transpose(freq_filters)
+
+    # Dictionary for storing fitting information
+    info = {"beta": beta, "XTXinv": XTXinv}
+
+    return filtered, info
 
 
 def freq_filter(gains, wgts, freqs, filter_scale=10.0, skip_wgt=0.1,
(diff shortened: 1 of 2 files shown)


jsdillon:
@tyler-a-cox Thanks so much! I'll take a look at the code. Since the relevant comparison is the slow and fast DPSS smooth_cal, rather than fast DPSS smooth_cal vs. CLEAN, can you show that comparison?

hera_cal/smooth_cal.py
```diff
@@ -68,7 +68,7 @@ def single_iterative_fft_dly(gains, wgts, freqs, conv_crit=1e-5, maxiter=100):
     return np.sum(taus)
 
 
-def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1e-8):
+def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1e-9, Nmax=2000):
```
jsdillon, line 71: How did you pick this default?

hera_cal/smooth_cal.py
```diff
@@ -82,55 +82,96 @@ def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1
         time_scale: time scale in seconds. Defined analogously to freq_scale.
             Note that time_scale is in seconds, times is in days.
         eigenval_cutoff: sinc_matrix eigenvalue cutoff to use for included dpss modes.
-            Only used when the filtering method is 'DPSS'
```
jsdillon, line 85: Why delete this line? It might seem redundant with the advanced user, but I think it's better to be explicit.

hera_cal/smooth_cal.py
jsdillon: Also, I'd suggest piping this parameter through to the command line argparser.

hera_cal/smooth_cal.py
```diff
@@ -256,10 +297,13 @@ def time_freq_2D_filter(gains, wgts, freqs, times, freq_scale=10.0, time_scale=1
...
-            one will be calculated using smooth_cal.dpss_filters and the time and frequency scale. Only used
-            when the method used is 'DPSS'
-        sol_matrix: Solution matrix obtained from performing linear least squares with 2D DPSS vectors
+        design_matrix: Tuple of 2 1D DPSS filters, one for the time axis and one for the frequency axis
+            that form the least squares design matrix, X, when the outer product of the two is taken.
+            If design_matrix is not provided, one will be calculated using smooth_cal.dpss_filters and the
+            time and frequency scale. Only used when the method is 'DPSS'
+        sol_matrix: Matrix operation of (X^T W X)^{-1} where X are the set of 2D DPSS vectors and W is the
+            wgts grid. Useful when filtering many gain grids where wgts grid is the same between two or more
+            gain solutions. Only used when the filterign method is 'DPSS'
```
jsdillon, line 306: typo `filterign`

hera_cal/tests/test_smooth_cal.py
```diff
@@ -42,27 +42,31 @@ def test_dpss_filters(self):
...
+        # test approximate estimator of the eigenvalues
+        time_filters, freq_filters = smooth_cal.dpss_filters(
+            times=times, freqs=freqs, time_scale=1.01 / time_scale, freq_scale=1.01 / freq_scale, Nmax=45
+        )
 
-    def test_fit_solution_matrix(self):
-        design_matrix = np.random.uniform(0, 1, size=(100, 10))
-        weights = np.random.uniform(0, 1, size=(10, 10))
-        sol_matrix = smooth_cal.fit_solution_matrix(weights, design_matrix)
+    def test_solve_2D_DPSS(self):
```
jsdillon, line 60: This test could use more assets/checks.

hera_cal/tests/test_smooth_cal.py
```diff
@@ -42,27 +42,31 @@ def test_dpss_filters(self):
...
         )
-        assert design_matrix.shape[0] == int(freqs.shape[0] * times.shape[0])
+        assert time_filters.shape[0] == times.shape[0]
+        assert freq_filters.shape[0] == freqs.shape[0]
         # test that error is thrown when filtering scale is too fine
         assert pytest.raises(ValueError, smooth_cal.dpss_filters, times=times, freqs=freqs,
                              time_scale=0.99 / time_scale, freq_scale=1.01 / freq_scale)
         assert pytest.raises(ValueError, smooth_cal.dpss_filters, times=times, freqs=freqs,
                              time_scale=1.01 / time_scale, freq_scale=0.99 / freq_scale)
+        # test approximate estimator of the eigenvalues
```
jsdillon, line 55: Should probably add some tests/asserts on `time_filters` and `freq_filters` like those above

hera_cal/smooth_cal.py
```diff
@@ -825,20 +870,12 @@ def time_freq_2D_filter(self, freq_scale=10.0, time_scale=1800.0, tol=1e-09, fil
...
-                edges_old = None
-
-        else:
-            design_matrix = None
+        # Keep track of the size of the weights grid after removing flagged edges
+        if skip_flagged_edges:
+            edges_old = None
 
         # Create default variables to be passed in to time_freq_2D_filter
+        design_matrix = None
```
jsdillon, line 878: I'd suggest explicitly initializing info as `{'sol_matrix': None, 'design_matrix': None}` so that the later calls are a bit less dependant on edges vs. edges_old and wgts_grid vs. wgts_old.

hera_cal/smooth_cal.py
```diff
@@ -82,55 +82,96 @@ def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1
...
-        time_filters[:, None, :, None] * freq_filters[None, :, None, :],
-        (times.shape[0] * freqs.shape[0], ntime_comps[0] * nfreq_comps[0]),
-    )  # 2D time-frequency DPSS vectors
-    return filters
-
-
-def fit_solution_matrix(weights, design_matrix):
+    dpss_windows = []
+
+    for x0, fw in zip([time_in_seconds, freqs], [fringe_scale, delay_scale]):
```
jsdillon, line 101: I'd like to see some more comments explaining what you're doing here.

hera_cal/smooth_cal.py
```diff
@@ -82,55 +82,96 @@ def dpss_filters(freqs, times, freq_scale=10, time_scale=1800, eigenval_cutoff=1
...
+
+        if Nmax < x0.shape[0]:
+            Nw = 2 * x0.shape[0] * fw * np.diff(x0)[0]
+            Nterms = int(min(10 * Nw, x0.shape[0]))
```
jsdillon, line 105: Where does 10 come from?

hera_cal/smooth_cal.py
```diff
@@ -256,10 +297,13 @@ def time_freq_2D_filter(gains, wgts, freqs, times, freq_scale=10.0, time_scale=1
         method: Algorithm used to smooth calibration solutions. Either 'CLEAN' or 'DPSS':
             'CLEAN': uses the CLEAN algorithm to smooth calibration solutions
             'DPSS': uses discrete prolate spheroidal sequences (DPSS) to filter calibration solutions
-        design_matrix: Matrix of DPSS filters calculated using smooth_cal.dpss_filters. If matrix is not provided,
-            one will be calculated using smooth_cal.dpss_filters and the time and frequency scale. Only used
-            when the method used is 'DPSS'
-        sol_matrix: Solution matrix obtained from performing linear least squares with 2D DPSS vectors
+        design_matrix: Tuple of 2 1D DPSS filters, one for the time axis and one for the frequency axis
```
jsdillon, line 300: Isn't it a bit confusing to call this the design_matrix? Is there something more precise we can call this pair of filters?

hera_cal/smooth_cal.py
```diff
@@ -298,16 +342,16 @@ def time_freq_2D_filter(gains, wgts, freqs, times, freq_scale=10.0, time_scale=1
...
-            if sol_matrix is None:
-                sol_matrix = fit_solution_matrix(wout, design_matrix)
+            time_filters, freq_filters = design_matrix
 
-            # Filter gains
-            filtered = design_matrix @ (sol_matrix @ gout.reshape(-1))
-            filtered = filtered.reshape(gout.shape)
+            # Filter gain solutions
+            filtered, fit_info = solve_2D_DPSS(
+                gains=gout, weights=wout, time_filters=time_filters, freq_filters=freq_filters, XTXinv=sol_matrix
```
jsdillon, line 353: Likewise, I think we should decide whether we want to call this the sol_matrix or XTXinv. Let's be consistent throughout the code. I wouldn't worry about breaking interfaces, your previous PR is too new.

jsdillon (changes requested):
Left a handful of comments, mostly documentation, comments, and nomenclature.

tyler-a-cox:
Thanks for the comments! Here's the comparison between slow vs fast DPSS. I also included a speed test for the two functions when the design matrix and solution matrix are already calculated.

Screen Shot 2022-04-20 at 5 36 33 PM

jsdillon:
Nice! That's quite an improvement. 

I actually meant the comparison of the DPSS-smoothed calibration solutions before and after. Like, are these the same to machine precision?

tyler-a-cox:
Oh sorry, I thought you were looking for runtime. They do seem to be within machine precision. Here's a plot of the absolute difference between the calibration solutions for the fast and slow DPSS:
Screen Shot 2022-04-20 at 9 10 17 PM

jsdillon:
Nice! That’s very persuasive.

I think the rest of my comments should be easy to deal with.

hera_cal/smooth_cal.py
tyler-a-cox: That was mostly chosen for performance reasons. After about N > 2000, calculating the DPSS vectors starts to take a while if you're generating the full set of N DPSS vectors. Also, as you start going to higher N, the approximation for the number of non-zero eigenvalues, Nz = 2 * N * fw * dt, gets fairly accurate. It's a somewhat arbitrary choice, but it does seem to be a good transition point between runtime and the accuracy of the estimation of the number of eigenfunctions we want to keep.

hera_cal/smooth_cal.py
tyler-a-cox: Fixed this

hera_cal/smooth_cal.py
tyler-a-cox: Fixed this

hera_cal/smooth_cal.py
tyler-a-cox: I meant to add that as an argument. I'll add more comments to this function, but the idea behind this speed-up is that when we generate these vectors, we generate the whole set (N vectors of size N) which takes a while to compute for large N.  Instead, what I'm attempting to do is estimate the number of vectors we need so that we don't have to compute all of the DPSS vectors, but just enough to where the minimum eigenvalue is at least as small as your eigenvalue cutoff. When the number of grid points in your DPSS vectors is large (N > 1000) you can estimate the number of DPSS vectors with close to unity eigenvalues using the approximation Nw = 2 * N * fw * dt. The `10 * Nw` was just providing a buffer so that we're sure to include enough vectors to meet the eigenvalue cut-off. Since the fall-off in the eigenvalues from 1 to 0 happens quickly, the value of `10` provided more than enough vectors for standard smoothcal inputs. There may be a better way to do this though

tyler-a-cox: I actually found a paper that gives an upper bound of the number of eigenvalues greater than some minimum eigenvalue so the buffer as an input argument won't be important.

hera_cal/smooth_cal.py
jsdillon: Gotcha. I'd make a comment in the docstring about why that's the default and what happens if the user raises/lowers it. Since you don't think the user is likely to change it (and it doesn't change the final answer, only the runtime, correct?) perhaps it doesn't need to be piped through to the command line.

hera_cal/smooth_cal.py
jsdillon: Feel free to cite the paper in the comments.

hera_cal/smooth_cal.py
tyler-a-cox: I'm considering completely removing it as a parameter and using this speed-up as the default. It improves the runtime by quite a bit and the results end up being the same regardless of how many data points are used.

jsdillon: As we discussed, magic numbers in the code aren't great. It's better to have it as a parameter, provide a reasonable default value, and warn the user in the docstring that changing the value is only for advanced users.

tyler-a-cox: Sorry, I meant that I'm considering completely removing `Nmax` from this function. The code within the `if` and `else` blocks end up giving the same results. The only difference between the two is that there is a minor indexing error in `uvtools.dspec.dpss_operator` that I'm going to put an issue in for. Here's a comparison between the two methods.

Screen Shot 2022-04-21 at 12 25 42 PM

hera_cal/smooth_cal.py
tyler-a-cox: Yes, I just switched over to `dpss_vectors` insteads

hera_cal/smooth_cal.py
tyler-a-cox: I cited two of the papers I used in the comments

hera_cal/smooth_cal.py
tyler-a-cox: I added more comments to this function

hera_cal/tests/test_smooth_cal.py
tyler-a-cox: I removed this test because I removed the `Nmax` keyword argument from this function, but added some additional tests

hera_cal/tests/test_smooth_cal.py
tyler-a-cox: Added more tests to this function

jsdillon (approved):
Looks great! Very nice work @tyler-a-cox

PR #783, opened 2022-04-20
Allow uvfits metadata only reads

bhazelton:
This was inspired by needing to fix a test to be robust against a antenna numbering fix in pyuvdata

If you don't want me do make this larger change to the package, I can just use pyuvdata functionality to find out what the antpairs are. But since metadata only reads of uvfits files are fully supported in pyuvdata, I figured I might as well make that work in this package.

see https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1148 for details

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -523,18 +523,13 @@ class HERAData(UVData):
             self.read = super().read  # re-define self.read so UVData can call self.read recursively for lists of files
             # load data
             try:
-                if self.filetype == 'uvh5':
-                    super().read(self.filepaths, file_type='uvh5', axis=axis, bls=bls, polarizations=polarizations,
+                if self.filetype in ['uvh5', 'uvfits']:
+                    super().read(self.filepaths, file_type=self.filetype, axis=axis, bls=bls, polarizations=polarizations,
                                  times=times, time_range=time_range, lsts=lsts, lst_range=lst_range, frequencies=frequencies, 
                                  freq_chans=freq_chans, read_data=read_data, run_check=run_check, check_extra=check_extra,
                                  run_check_acceptability=run_check_acceptability, **kwargs)
-                elif self.filetype == 'uvfits':
-                    super().read(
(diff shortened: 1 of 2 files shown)


hera_cal/io.py
```diff
@@ -297,7 +297,7 @@ def __init__(self, input_data, upsample=False, downsample=False, filetype='uvh5'
         self.filetype = filetype
 
         # load metadata from file
-        if self.filetype == 'uvh5':
+        if self.filetype in ['uvh5', 'uvfits']:
```
jsdillon, line 300: The docstring for `__init__` should be changed to reflect this new capability.

hera_cal/io.py
```diff
@@ -528,6 +528,11 @@ def read(self, bls=None, polarizations=None, times=None, time_range=None, lsts=N
                                  times=times, time_range=time_range, lsts=lsts, lst_range=lst_range, frequencies=frequencies, 
                                  freq_chans=freq_chans, read_data=read_data, run_check=run_check, check_extra=check_extra,
                                  run_check_acceptability=run_check_acceptability, **kwargs)
+                elif self.filetype == 'uvfits':
+                    super().read(self.filepaths, file_type='uvfits', axis=axis, bls=bls, polarizations=polarizations, times=times,
```
jsdillon: This line can be merged with the equivalent the above read for uvh5. Then we can make `self.unphase_to_drift()` an addition interior `if`.

jsdillon (changes requested):
Thanks @bhazelton for adding this new functionality. 

This PR looks basically fine, just a few minor changes requested.

From a hera_cal perspective, I'd suggest changing the PR title to reflect that they key change here is enabling metadata reads for uvfits files (just to make it more searchable in the future). That fact that it fixes a test is a nice added side benefit.

jsdillon (approved):
Assuming tests pass, I approve.

bhazelton:
@jsdillon the codecov project level check fails here because I reduced the number of lines of code, so the denominator went down.

jsdillon:
I consider that passing tests.

PR #784, opened 2022-04-22
Enable weighting of blacklisted times, etc. in smooth_cal and FRF larger than 0

jsdillon:
DPSS struggles to interpolate over blacklisted regions and is especially bad at extrapolating to edge times with 0 weight. In some cases, flagged region simply leads to flagging because the region of filtering is taken to be the contiguous region with non-zero weights.

This PR enables blacklisting that downweights but doesn't necessarily downweight all the way to zero. It also harmonizes the nomenclature for this idea between `smooth_cal` and `frf`.


tyler-a-cox (approved):
This looks good to me. Thanks for adding this Josh!

PR #785, opened 2022-04-28
Made hera_cal default to not run file checks in io.py

AaronParsons:
I think we are losing a lot of project time (and field run time) rechecking files. Default for hera_cal should be to not run these checks on data and cal files. Here is a shot at it.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -135,7 +135,10 @@ class HERACal(UVCal):
 
         if self.filepaths is not None:
             # load data
-            self.read_calfits(self.filepaths[0])
+            self.read_calfits(self.filepaths[0],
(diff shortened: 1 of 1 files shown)


jsdillon:
I'm not opposed to changing this default @AaronParsons, but my sense has always been that the time running checks was tiny compared to the I/O time. Do you have any evidence that it's significant?

AaronParsons:
Maybe @aewallwi can comment. We were waiting a ridiculously long time to read in HERACal files (~10 MB each?) on lustre. Definitely not I/O bound.

jsdillon:
ARP and I did a little test and found that this is not the leading order slowdown.l

PR #786, opened 2022-04-29
Fix uvfits test to be robust against antenna numbering fix in pyuvdata

bhazelton:
This fixes a line I missed in #783.


jsdillon (approved):
Looks fine to me.

PR #789, opened 2022-05-10
Avoid reading data to speed up `lstbin.gen_bl_nightly_dicts` and fix major memory leak

jsdillon:
This PR remove an unnecessary read from `lstbin.gen_bl_nightly_dicts()` that was causing a whole file to be loaded into memory from every night being lst-binned. Not only was this a major memory leak, but I don't think the read is necessary and all relevant information can be gotten from uvdata metadata. This PR does that, and streamlines some tasks using `redcal.filter_reds()`. 

Here's a demonstration of running `gen_bl_nightly_dicts` on 18 nights of validation data. Note both the reduction in time (from over 13 minutes to 42 seconds), the reduction in increased memory usage (from ~80 GB to basically 0) and that both techniques give precisely the same answer:

image


aewallwi (approved):
Looks good to me.

PR #790, opened 2022-05-19
Added a read_hera_hdf5 function to hera_cal.io, which extracts data f…

AaronParsons:
…rom an hdf5 file ~100x faster than previous interfaces.

I'm tagging a lot of reviewers here so we can start a conversation about supporting an interface to access HERA data faster. The function in this branch is a proof of concept that we can get a huge amount of speedup for file reads.


AaronParsons:
image

AaronParsons:
A relevant link about hdf5 read speeds: https://stackoverflow.com/questions/41856463/increasing-reading-speed-for-h5py

AaronParsons:
image
Speedup for reading whole files

AaronParsons:
image
And here's another comparison with a short list of baselines

AaronParsons:
Another relevant hdf5 speed-related post: https://stackoverflow.com/questions/21766145/h5py-correct-way-to-slice-array-datasets

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+    bl2ind = {}
+    inds = {}
+    # Read file metadata to size up arrays and sort times
+    for filename in filenames:
+        if verbose:
+            print(f'Reading header of {filename}')
+        with h5py.File(filename, 'r') as f:
+            h = f['/Header']
+            if check:
+                assert int(h['Nspws'][()]) == 1  # not a hera file
```
jsdillon, line 904: Can you have these checks print some useful error info?

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+                   dtype=np.complex128, verbose=False):
+    '''A ~100x faster interface to getting data out of HERA HDF5 files. Only concatenates
+    along time axis. Assumes that if antenna_numbers header array stays the same,
+    the baseline order in the files stays the same. Puts times in ascending order,
+    but does not check that files are contiguous.
+
+    Arguments:
+        filenames: list of files to read
+        bls: list of (ant_1, ant_2, [polstr]) tuples to read out of files
+        pols: list of pol strings to read out of files
```
jsdillon, line 878: Can you indicate what happens to this parameter if the bls have `polstr` provided?

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+        full_read_thresh (0.01): fractional threshold for reading whole file
```
jsdillon, line 879: docstring doesn't match default kwarg

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+                                 instead of baseline by baseline.
+        read_data (bool, True): read visdata
+        read_flags (bool, False): read flags
+        read_nsamples (bool, False): read nsamples
+        check (bool, False): run sanity checks to make sure files match.
+        dtype (np.complex128): numpy datatype for output complex-valued arrays
+        verbose: print some progress messages.
+
+    Returns:
+        rv: dictionary with keys 'info' (metadata), 'visdata' (dictionary of waterfalls
```
jsdillon, line 889: In most of hera_cal, `visdata` is just called `data`. Any objections to calling it that?

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+            if check:
+                assert int(h['Nspws'][()]) == 1  # not a hera file
+            if len(times) == 0:
+                info['freqs'] = h['freq_array'][()]
+                pol_array = h['polarization_array'][()]
+                npols = pol_array.size
+                pol_indices = {POL_NUM2STR_DICT[n]: cnt for cnt, n in enumerate(pol_array)}
+                info['pols'] = list(pol_indices.keys())
+            elif check:
+                assert int(h['Nfreqs'][()]) == info['freqs'].size
```
jsdillon, line 912: Same comment as above

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+    for filename in filenames:
+        inds = bl2ind[filename]
+        if verbose:
+            print(f'Reading data from {filename}')
+        with h5py.File(filename, 'r') as f:
+            h = f['/Header']
+            ntimes = int(h['Ntimes'][()])
+            nbls = int(h['Nblts'][()]) // ntimes
+            if check:
+                assert np.allclose(h['time_array'][:ntimes], times[t:t + ntimes])
```
jsdillon, line 973: Same comment as above

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
             yield self.read(times=times[i:i + Nints])
 
 
+def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
+                   read_data=True, read_flags=False, read_nsamples=False, check=False,
+                   dtype=np.complex128, verbose=False):
+    '''A ~100x faster interface to getting data out of HERA HDF5 files. Only concatenates
+    along time axis. Assumes that if antenna_numbers header array stays the same,
```
jsdillon, line 871: `antenna_numbers` is out of date, you look at `ant_1_array` and `ant_2_array now, right?

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+            _hash = hash((ant1_array.tobytes(), ant2_array.tobytes()))
+            # map baselines to array indices for each unique antenna order
+            if _hash not in inds:
+                inds[_hash] = {(i, j): slice(n * ntimes, (n + 1) * ntimes)
+                               for n, (i, j) in enumerate(zip(ant1_array,
+                                                              ant2_array))}
+                if 'bls' not in info:
+                    info['bls'] = set(inds[_hash].keys())
+                else:
+                    info['bls'].intersection_update(set(inds[_hash].keys()))
```
jsdillon, line 926: Why are we doing an intersection update? This seems very strange. Shouldn't this code error if the ant_1_array and ant_2_array don't match from file to file?

This seems like it could silently drop baselines not common to all files.

hera_cal/io.py
jsdillon, line 889: Also, can you list what's in the metadata and what formats to expect in there?

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+            ant1_array = h['ant_1_array'][::ntimes]
+            ant2_array = h['ant_2_array'][::ntimes]
+            _hash = hash((ant1_array.tobytes(), ant2_array.tobytes()))
+            # map baselines to array indices for each unique antenna order
+            if _hash not in inds:
+                inds[_hash] = {(i, j): slice(n * ntimes, (n + 1) * ntimes)
+                               for n, (i, j) in enumerate(zip(ant1_array,
+                                                              ant2_array))}
+                if 'bls' not in info:
+                    info['bls'] = set(inds[_hash].keys())
```
jsdillon, line 924: In pyuvdata and sometimes also in hera_cal, length-2 tuples are called `antpairs` rather than `bls` and `bls` is used for length-3 tuples that include polarizations. We're not super consistent about this though, so if you really want to call these `bls` I won't stand in your way.

jsdillon: Alternatively, you can do `info['bls'] =  set(bl + (p,) for bl in bls for p in pols)` at the end of this.

hera_cal/io.py
jsdillon: And say whether the metadata is relevant to the whole file or to the selection that was performed? For example, if you ask for one baseline, will info know about all the other baselines?

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+            if len(times) == 0:
+                info['freqs'] = h['freq_array'][()]
+                pol_array = h['polarization_array'][()]
+                npols = pol_array.size
+                pol_indices = {POL_NUM2STR_DICT[n]: cnt for cnt, n in enumerate(pol_array)}
+                info['pols'] = list(pol_indices.keys())
+            elif check:
+                assert int(h['Nfreqs'][()]) == info['freqs'].size
+            ntimes = int(h['Ntimes'][()])
+            times.append((h['time_array'][:ntimes], filename))
```
jsdillon, line 914: I'd suggest also pulling out the lst_array, which is generally quite useful.

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+    info['times'] = times
+
+    # preallocate buffers
+    rv = {}
+    if read_data:
+        rv['visdata'] = {bl: np.empty((times.size, info['freqs'].size), dtype=dtype) for bl in bls}
+    if read_flags:
+        rv['flags'] = {bl: np.empty((times.size, info['freqs'].size), dtype=bool) for bl in bls}
+    if read_nsamples:
+        rv['nsamples'] = {bl: np.empty((times.size, info['freqs'].size), dtype=np.float32) for bl in bls}
```
jsdillon, line 958: Do you want to cast these as DataContainers? I feel like that would make this more interoperable with the rest of `hera_cal` without adding much overhead.

jsdillon: Also, as DataContainers know to expect metadata as class attributes, namely `'ants', 'data_ants', 'antpos', 'data_antpos', 'freqs', 'times', 'lsts', 'times_by_bl', 'lsts_by_bl'`, so perhaps a lot of the metadata currently stored in the `info` object could be stored instead in these objects. There might be some duplication of metadata as a result, but I think that'd be trivial in terms of extra memory and runtime, no?

hera_cal/io.py
jsdillon: Also, I kind of feel like this would slightly easier to use if it returned 3 or 4 arguments and just stuck `None` if you didn't ask for nsamples or flags or whatever. That's a bit more similar to how `HERAData.read()` and `HERACal.read()` work.

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
...
+            if full_read and verbose:
+                print('Reading full file')
+            for key, data in rv.items():
+                d = f['/Data'][key]  # data not read yet
+                if full_read:
+                    d = d[()]  # reads data
+                # handle HERA's raw (int) and calibrated (complex) file formats
+                if key == 'visdata' and not np.iscomplexobj(d):
+                    # Support polarization-transposed arrays
+                    if d.shape[-1] == info['freqs'].size:
```
jsdillon, line 985: I think you can save a few lines of code if you put this if inside the for loop

jsdillon (changes requested):
Very nice stuff @AaronParsons!

I've got a few functional questions about edge cases, a few thoughts about making APIs more similar to the rest of `hera_cal`, and some documentation/cometic suggestions.

hera_cal/io.py
AaronParsons: This is how it was originally, but then I added options for reading data/flags/nsamples in any subset, and I didn't want a case where positional returns got misinterpreted.

I agree most of hera_cal calls it "data", but pyuvdata (and the file itself) calls it "visdata", and I think of this function as a thin stand-in to pyuvdata and would like it to stay as close to the file format as possible

hera_cal/io.py
AaronParsons: I purposely unrolled to not do if checks inside the for loop, which is slower

hera_cal/io.py
AaronParsons: I don't think it's correct to think of this function as a stand-in for DataContainers or any other hera_cal object. It should be thought of as a stand-in for pyuvdata. The goal should be to get the data into a generic format (dicts are fine) as fast as possible. Then HERAData can take care of the rest.

That said, I agree that "info" is light on the metadata it reads out and should be expanded as far as it can without hurting performance. Is this a complete list of commonly used file metadata?

hera_cal/io.py
AaronParsons: I don't see why this would need to error. If baselines legitimately change from file to file, then the only thing "read all" could mean is to return the baselines which appear in all files. If you provide specific baselines, and they are not in all the files, it will error out in that case.

hera_cal/io.py
```diff
@@ -864,6 +864,147 @@ def iterate_over_times(self, Nints=1, times=None):
             yield self.read(times=times[i:i + Nints])
 
 
+def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
+                   read_data=True, read_flags=False, read_nsamples=False, check=False,
+                   dtype=np.complex128, verbose=False):
+    '''A ~100x faster interface to getting data out of HERA HDF5 files. Only concatenates
```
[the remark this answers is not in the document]
AaronParsons: Removed statement.

hera_cal/io.py
AaronParsons: Done

hera_cal/io.py
AaronParsons: Done

hera_cal/io.py
AaronParsons: Fixed

hera_cal/io.py
AaronParsons: Done

hera_cal/io.py
AaronParsons: Done

hera_cal/io.py
AaronParsons: Removed

hera_cal/io.py
AaronParsons: Done

hera_cal/io.py
AaronParsons: Done

hera_cal/io.py
AaronParsons: Leaving this all as stands. bls purposely not overwritten until after all files are scanned

hera_cal/io.py
AaronParsons: I added antpos, but lst is not in the files, so i'm not computing it

hera_cal/io.py
AaronParsons: Meh, I'll do 'data'.

hera_cal/io.py
AaronParsons: actually, lst_array is not in the files, so i'm not returning it

hera_cal/io.py
jsdillon: In that case, can you clarify that in the docstring?

jsdillon (approved):

hera_cal/io.py
jsdillon: Ok, then I think perhaps the next thing to do (and this is probably on me) is to write a wrapper around this function that puts things into a format that might be closer to a drop-in replacement for HERAData in the code.

PR #793, opened 2022-05-21
Increased filtering options for vis_clean, delay_filter, and tophat_frfilter

jsdillon:
In theory, I haven't changed any defaults.

EDIT: After consulting with @aewallwi, we've decided to harmonize the defaults to what was the default for DPSS wherever possible.


aewallwi (approved):
This looks good to me.

PR #794, opened 2022-05-23
fixed gp_interp1d predict shape update in sklearn

nkern:
fixes a bug where sklearn.GP.predict changed their output shape for GP.predict.


nkern:
when I pulled from  main locally I didn't get everything for some reason. starting over on a new branch

PR #795, opened 2022-05-23
Median nightly delays

aewallwi:
Depens on #772

### DIFF
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -1578,7 +1578,7 @@ def redundantly_calibrate(data, reds, freqs=None, times_by_bl=None, fc_conv_crit
         times_by_bl = data.times_by_bl
 
     # perform firstcal
-    rv['fc_meta'], rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit, 
+    rv['fc_meta'], rv['g_firstcal'] = rc.firstcal(data, freqs, maxiter=fc_maxiter, conv_crit=fc_conv_crit,
                                                   fc_min_vis_per_ant=fc_min_vis_per_ant)
     rv['gf_firstcal'] = {ant: np.zeros_like(g, dtype=bool) for ant, g in rv['g_firstcal'].items()}
 
@@ -1736,7 +1736,7 @@ def redcal_iteration(hd, nInt_to_load=None, pol_mode='2pol', bl_error_tol=1.0, e
                 cal = redundantly_calibrate(data, reds, freqs=hd.freqs[fSlice], times_by_bl=hd.times_by_bl,
                                             fc_conv_crit=fc_conv_crit, fc_maxiter=fc_maxiter,
                                             oc_conv_crit=oc_conv_crit, oc_maxiter=oc_maxiter,
-                                            check_every=check_every, check_after=check_after, 
+                                            check_every=check_every, check_after=check_after,
                                             max_dims=max_dims, gain=gain, fc_min_vis_per_ant=fc_min_vis_per_ant)
                 expand_omni_sol(cal, filter_reds(all_reds, pols=pols), data, nsamples)
 
@@ -1956,6 +1956,54 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
 
     return cal
 
+def build_median_firstcal_delays(redcal_meta_file_list, output_ext='.redcal_meta.median_delay.hdf5',
+                                 output_replace='.redcal_meta.hdf5'):
+    """
+    Find the median delay and polarity for list of firstcal meta files.
+    and write them out with that median.
+
+    Parameters
+    ----------
+    redcal_meta_file_list: list of str
+        list of file names containing redcal meta data. All files should have same number of times.
+
+    """
+    recal_metas = []
+    for meta_file in recal_meta_file_list:
+        recal_metas.append(read_redcal_meta(meta_file))
+    # build list of delays and polarity flips for each antenna.
+    nfiles = len(redcal_metas)
+    ntimes = [len(rcm[4]) for rcm in redcal_metas]
+    delays = {ant: np.zeros(np.sum(ntimes)) for ant in recal_meta_list[0][1]['dlys']}
+    polarity_flips = copy.deepcopy(delays)
+    nt = 0
+    for filenum, (meta_filename, fc_meta, omni_meta, freqs, times, lsts, antpos, history) in enumerate(recal_metas):
+        for ant in fc_meta['dlys']:
+            tslice = slice(nt, nt + len(times))
+            delays[ant][tslice] = fc_meta['dlys'][ant]
+            polarity_flips[tslice] = fc_meta['polarity_flips'][ant]
+        nt += len(times)
+    # compute medians.
+    for ant in delays:
+        delays[ant] = np.nanmedian(delays[ant])
+        polarity_flips[ant] = np.nanmedian(polarity_flips[ant])
+
+    # update metadata and write out.
+    for filenum, (meta_filename, fc_meta, omni_meta, freqs, times, lsts, antpos, history) in enumerate(recal_metas):
+        for ant in fc_meta['dlys']:
+            fc_meta['dlys'][ant][:] = delays[ant]
+            fc_meta['polarity_flips'][:] = polarity_flips[ant]
+        save_redcal_meta(meta_filename.replace(output_replace, output_ext), fc_meta, omni_meta, freqs, times, lsts, antpos,
+                         history + '\nTook nightly time median on delay and polarity flips.')
+
+def build_median_firstcal_delays_argparser():
+    """Arg praser for build_median_firstcal_delays
+    """
+    ap = argparse.ArgumentParser(description="Compute nightly median of firstcal delays derived from per-observation firstcal meta files. Update the files and write out with medians.")
+    ap.add_argument("redcal_meta_file_list", type=str, nargs="+", help="List of reccal meta file names.")
+    ap.add_argument("--output_ext", type=str, default='.redcal_meta.median_delay.hdf5', help="File extensionf of output files. This will overwrite output_replace in original file names to drive output filenames.")
+    ap.add_argument("--output_replace", type=str, default=".redcal_meta.hdf5", help="File extension in non median files to replace in each file name with output_ext.")
+    return ap
 
 def redcal_argparser():
     '''Arg parser for commandline operation of redcal_run'''
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -1956,6 +1956,7 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
 
     return cal
 
+
 def nightly_median_firstcal_delays(redcal_meta_file_list, output_ext='.redcal_meta.median_delay.hdf5',
                                  output_replace='.redcal_meta.hdf5'):
     """
@@ -1970,12 +1971,12 @@ def nightly_median_firstcal_delays(redcal_meta_file_list, output_ext='.redcal_me
     """
     recal_metas = []
     for meta_file in recal_meta_file_list:
-        recal_metas.append(read_redcal_meta(meta_file))
+        recal_metas.append((meta_file, ) + read_redcal_meta(meta_file))
     # build list of delays and polarity flips for each antenna.
     nfiles = len(redcal_metas)
     ntimes = [len(rcm[4]) for rcm in redcal_metas]
     delays = {ant: np.zeros(np.sum(ntimes)) for ant in recal_meta_list[0][1]['dlys']}
-    polarity_flips = copy.deepcopy(delays)
+    polarity_flips = copy.deepcopy(delays).astype(bool)
     nt = 0
     for filenum, (meta_filename, fc_meta, omni_meta, freqs, times, lsts, antpos, history) in enumerate(recal_metas):
         for ant in fc_meta['dlys']:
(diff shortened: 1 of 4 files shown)


aewallwi:
The problem that I am trying to fix arises from transient misidentified antenna polarities. I found that using `remove_degen_gains` with a median delay and polarity over the night does not seem to fix the problem. What does seem to fix the problem is just mulitplying each antenna gain by (-1) ** abs(median_polarity - per_time_polarity). Here are some examples:

Original phases
image

Using `remove_degen_gains` where `degen_gains` is set to the median delay and polarity.
image

Here is just multiplying by (-1) ** abs(median_polarity - per_time_polarity)
image

And finally, here is using `remove_degen_gains` only to fix delays and then using a separate step of (-1) ** abs(median_polarity - per_time_polarity) to fix the polarities 

image

jsdillon:
I'm surprised that last step does literally nothing... Have you checked multiple antennas? What's your reference antenna?

aewallwi:
> I'm surprised that last step does literally nothing... Have you checked multiple antennas? What's your reference antenna?
> 
> We saw jumps like these in H1C IDR2 before there was ever any polarity flipping, e.g. figure 6 of https://arxiv.org/pdf/2108.02263.pdf

The last plot had a bug. This is actually what happens when you try to fix the delays without paying attention to polarity flips and then flip polarities as a separate step.


image

aewallwi:
I've updated this PR so that the default recipe for updating first-cal is to both switch the polarities and to update the offsets and delays in the degenerate subspace.

hera_cal/redcal.py
```diff
@@ -1957,6 +1963,123 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
...
+        for ant in fc_meta['dlys']:
+            tslice = slice(nt, nt + len(times))
+            delays[ant][tslice] = fc_meta['dlys'][ant]
+            polarity_flips[ant][tslice] = fc_meta['polarity_flips'][ant]
+            offsets[ant][tslice] = fc_meta['offsets'][ant]
+        nt += len(times)
+    # compute medians.
+    for ant in delays:
+        delays[ant] = np.nanmedian(delays[ant])
+        polarity_flips[ant] = bool(np.nanmedian(polarity_flips[ant]))
```
jsdillon, line 1998: If all the polarity flips are `np.nan` for a given antenna, this will turn into `True` though I think you want it to be `False` by default. Probably doesn't matter but you might want an a check on `not np.any(np.isfinite())`

hera_cal/redcal.py
```diff
@@ -1957,6 +1963,123 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
...
+    replace_degens: bool, optional
+        If True, replace degenerate portion of calibration solution with solution in redcal_meta_file.
+    fix_polarities: bool, optional
+        If True, fix polarities specified in old_redcal_meta_file.
+    clobber: bool, optional
+        overwrite output calfits.
+    """
+    hc = HERACal(redcal_file)
+    # get redundant calibrator.
+    # get reds
```
jsdillon, line 2035: I think these comments got misplaced

hera_cal/redcal.py
```diff
@@ -1957,6 +1963,123 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
...
+        path to calfits file containing redcal solution to update.
+    redcal_meta_file:
+        path to redcal meta file generated with io.save_redcal_meta()  with delays to be used to update degeneracy.
+    old_redcal_meta_file:
+        path to redcal meta file with old polarity flips.
+    output_file: str, optional
+        path to output file. Optional to write output to.
+    replace_degens: bool, optional
+        If True, replace degenerate portion of calibration solution with solution in redcal_meta_file.
+    fix_polarities: bool, optional
```
jsdillon, line 2028: If you have this option, it's weird that `old_redcal_meta_file` is a required parameter. Perhaps you want to put a way to allow the user to pass None for `old_redcal_meta_file` if this is False?

hera_cal/redcal.py
```diff
@@ -1957,6 +1963,123 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
...
+    fc_meta, omni_meta, freqs, times, lsts, antpos, history = read_redcal_meta(redcal_meta_file)
+    fc_meta_old = read_redcal_meta(old_redcal_meta_file)[0]
+    reds = get_reds(antpos, pols=[pol.replace('J', '').replace('j', '') for pol in hc.pols])
+    rc = RedundantCalibrator(reds)
+    firstcal_gains = {}
+
+    if not dont_replace_polarities:
+        for ant in fc_meta['polarity_flips']:
+            gains[ant] *= (-1. + 0j) ** np.abs(fc_meta['polarity_flips'][ant][:, None] - fc_meta_old['polarity_flips'][ant][:, None])
+    ntimes = hc.Ntimes
```
jsdillon, line 2046: ntimes isn't used

hera_cal/redcal.py
```diff
@@ -1957,6 +1963,123 @@ def redcal_run(input_data, filetype='uvh5', firstcal_ext='.first.calfits', omnic
...
+    if not dont_replace_degens:
+        for ant in gains:
+            dly_factor = np.exp(2j * np.pi * freqs[None, :] * fc_meta['dlys'][ant][:, None]) * np.exp(1j * fc_meta['offsets'][ant][:, None])
```
jsdillon, line 2049: a bit weird to call it a dly_factor when it includes an offset (which can be summed inside the exponent)

hera_cal/redcal.py
```diff
@@ -1986,7 +2109,7 @@ def redcal_argparser():
                              Default None loads all integrations.")
     redcal_opts.add_argument("--upsample", default=False, action="store_true", help="Upsample BDA files to the highest temporal resolution.")
     redcal_opts.add_argument("--downsample", default=False, action="store_true", help="Downsample BDA files to the highest temporal resolution.")
-    redcal_opts.add_argument("--pol_mode", type=str, default='2pol', help="polarization mode of redundancies. Can be '1pol', '2pol', '4pol', or '4pol_minV'. See recal.get_reds documentation.")
+    redcal_opts.add_argument("--pol_mode", type=str, default='2pol', help="polarization mode of redundancies. Can be '1pol', '2pol', '4pol', or '4pol_minV'. See redcal.get_reds documentation.")
```
jsdillon, line 2112: lol, thanks

hera_cal/tests/test_redcal.py
```diff
@@ -1499,6 +1499,55 @@ def test_redcal_iteration(self):
...
+        # now replace the redcal degeneracies.
+        omnifiles = [dfile.replace('.uvh5', '.omni.calfits') for dfile in data_files]
+        for omni_file, meta_file, meta_file_old in zip(omnifiles, meta_files_median, meta_files):
+            om.update_redcal_phase_degeneracy(omni_file, meta_file, meta_file_old,
+                                              output_file=omni_file.replace('.calfits', '.medphase.calfits'), clobber=True)
+
+        # check that files have been created
+        for ofile in omnifiles:
+            assert os.path.exists(omni_file.replace('.calfits', '.medphase.calfits'))
+
```
jsdillon, line 1550: Is there any other test we can do? Like, does doing this make the calibration solutions smoother in time? If so, can we quantify that and make an assertion about it?

jsdillon (changes requested):
Looking pretty good. Just a few minor tweaks.

Also, not that the PR is almost done, can you add some detail to the PR description and show some plots of this procedure in action on real data?

hera_cal/tests/test_redcal.py
aewallwi: I think it's overkill to show whether or not the calibration is smoother. Anecdotally, I find that the differences in the solutions after replacing the degeneracies are quite small. I will document this above.

aewallwi:
I am unable to reproduce the macOS unittest failures on a mac. I have tried to create a fresh conda environment and still no luck. Should we remove the macOS tests?

jsdillon:
Maybe @bhazelton or @plaplant have ideas about how to debug testing failures that appear only to happen in the CI but can't be reproduced locally...

plaplant:
@aewallwi @jsdillon I ran the tests locally on my (M1-based) Mac in a fresh environment, and I'm getting a test failure too, though on a different test than the CI failures. I've posted the full traceback at the bottom of my comment.

As a way to debug testing failures in CI that don't appear locally, I tend to ssh into the machine running the job, and poke around on the command line there. This way you're using the exact same environment and hardware that produced the error, so it's easier to figure out what's going wrong. I would **highly** (can't stress this enough) recommend against ignoring test failures or tuning off CI checks. It's better in the long run to get to the bottom of failing tests, and understand what's going wrong.

My local test failure:
```
__________________________________________________________________ Test_ReflectionFitter_XTalk.test_misc_svd_funcs ___________________________________________________________________

self = <hera_cal.tests.test_reflections.Test_ReflectionFitter_XTalk object at 0x12fb78640>

    def test_misc_svd_funcs(self):
        # setup RF object
        RF = reflections.ReflectionFitter(self.uvd)
        # add noise
        np.random.seed(0)
        Namp = 3e0
        for k in RF.data:
            RF.data += stats.norm.rvs(0, Namp, RF.Ntimes * RF.Nfreqs).reshape(RF.Ntimes, RF.Nfreqs) + 1j * stats.norm.rvs(0, Namp, RF.Ntimes * RF.Nfreqs).reshape(RF.Ntimes, RF.Nfreqs
)
        bl = (23, 24, 'ee')

        # fft data
        RF.fft_data(data=RF.data, window='blackmanharris', overwrite=True)

        # sparse sv decomp
        svd_wgts = RF.svd_weights(RF.dfft, RF.delays, min_dly=150, max_dly=500, side='both')
        RF.sv_decomp(RF.dfft, wgts=svd_wgts, keys=[bl], overwrite=True, Nkeep=None, sparse_svd=True)
        assert RF.umodes[bl].shape == (100, 98)
        assert RF.vmodes[bl].shape == (98, 128)

>       RF.sv_decomp(RF.dfft, wgts=svd_wgts, keys=[bl], overwrite=True, Nkeep=10, sparse_svd=True)

hera_cal/tests/test_reflections.py:429:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
hera_cal/reflections.py:589: in sv_decomp
    u, svals, v = sparse.linalg.svds(dfft[k] * wgts[k], k=Nk, which='LM')
../../../../miniforge3/envs/hera_cal_tests/lib/python3.10/site-packages/scipy/sparse/linalg/_eigen/_svds.py:350: in svds
    eigvals, eigvec = eigsh(XH_X, k=k, tol=tol ** 2, maxiter=maxiter,
../../../../miniforge3/envs/hera_cal_tests/lib/python3.10/site-packages/scipy/sparse/linalg/_eigen/arpack/arpack.py:1566: in eigsh
    ret = eigs(A, k, M=M, sigma=sigma, which=which, v0=v0,
../../../../miniforge3/envs/hera_cal_tests/lib/python3.10/site-packages/scipy/sparse/linalg/_eigen/arpack/arpack.py:1345: in eigs
    params.iterate()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <scipy.sparse.linalg._eigen.arpack.arpack._UnsymmetricArpackParams object at 0x2da3c35b0>

    def iterate(self):
        if self.tp in 'fd':
            self.ido, self.tol, self.resid, self.v, self.iparam, self.ipntr, self.info =\
                self._arpack_solver(self.ido, self.bmat, self.which, self.k,
                                    self.tol, self.resid, self.v, self.iparam,
                                    self.ipntr, self.workd, self.workl,
                                    self.info)
        else:
            self.ido, self.tol, self.resid, self.v, self.iparam, self.ipntr, self.info =\
                self._arpack_solver(self.ido, self.bmat, self.which, self.k,
                                    self.tol, self.resid, self.v, self.iparam,
                                    self.ipntr, self.workd, self.workl,
                                    self.rwork, self.info)

        xslice = slice(self.ipntr[0] - 1, self.ipntr[0] - 1 + self.n)
        yslice = slice(self.ipntr[1] - 1, self.ipntr[1] - 1 + self.n)
        if self.ido == -1:
            # initialization
            self.workd[yslice] = self.OP(self.workd[xslice])
        elif self.ido == 1:
            # compute y = Op*x
            if self.mode in (1, 2):
                self.workd[yslice] = self.OP(self.workd[xslice])
            else:
                Bxslice = slice(self.ipntr[2] - 1, self.ipntr[2] - 1 + self.n)
                self.workd[yslice] = self.OPa(self.workd[Bxslice])
        elif self.ido == 2:
            self.workd[yslice] = self.B(self.workd[xslice])
        elif self.ido == 3:
            raise ValueError("ARPACK requested user shifts.  Assure ISHIFT==0")
        else:
            self.converged = True

            if self.info == 0:
                pass
            elif self.info == 1:
                self._raise_no_convergence()
            else:
>               raise ArpackError(self.info, infodict=self.iterate_infodict)
E               scipy.sparse.linalg._eigen.arpack.arpack.ArpackError: ARPACK error 3: No shifts could be applied during a cycle of the Implicitly restarted Arnoldi iteration. One possibility is to increase the size of NCV relative to NEV.

../../../../miniforge3/envs/hera_cal_tests/lib/python3.10/site-packages/scipy/sparse/linalg/_eigen/arpack/arpack.py:757: ArpackError
```

jsdillon:
Thanks @plaplant. How do you ssh into github actions?

plaplant:
@jsdillon it looks like there isn't a built-in way to do it for Github Actions (unlike in CircleCI). I found this: https://github.com/marketplace/actions/debugging-with-ssh. Maybe that will work?

aewallwi:
> @jsdillon it looks like there isn't a built-in way to do it for Github Actions (unlike in CircleCI). I found this: https://github.com/marketplace/actions/debugging-with-ssh. Maybe that will work?

The test that is failing does not fail on the machines of actual users. The tests pass for linux which is really the OS that is going to be used by our analysis. I think this raises an interesting question. Do we really need to support multiple operating systems in the first place when Linux is the standard for our production analysis and the error here is not reproducible on other macOS machines?

PR #797, opened 2022-05-24
Use pyuvdata's future array shapes

jsdillon:
Following https://github.com/HERA-Team/hera_qm/pull/413 we can now start using pyuvdata's future array shapes.

For people using hera_cal.io as a wrapper around pyuvdata, this shouldn't change anything. However, in the relatively rare instances where users input pyvudata objects directly into hera_cal functions and classes, they'll need to call `.use_future_array_shapes()` on them first.

This turns out to be somewhat urgent, because that PR actually broke some hera_cal tests.


bhazelton (approved):
Looks good to me.

PR #798, opened 2022-05-25
Replace uvtools with hera_filters

tyler-a-cox:
The filtering code in `uvtools` was recently moved into its own repository, `hera_filters`. I replaced all instances of `uvtools` with `hera_filters`


aewallwi (changes requested):
Thanks for all the hard work getting this converted @tyler-a-cox . The only general comment I have is that for some reason the `HAVE_UVTOOLS` equivalent lines are now no longer covered by tests. Any idea why this is (in other words, how were the `HAVE_UVTOOLS` lines covered in the first place?) If this is too much of a dive into testing esoterica I'm in favor of just making `hera_filters` a dependancy of `hera_cal` and just dispensing of these lines -- much of the core functionality in `hera_cal` relies on `hera_filters` and the `hera_filters` dependencies are minimal (or at least smaller then other dependancies like `hera_qm`) Let me know if you are in favor of one path or the other.

tyler-a-cox:
I lost track of this one for a while. @aewallwi, I decided to make `hera_filters` a dependency, since it doesn't rely on any additional libraries `hera_cal` isn't already using. I did pin the numpy version for testing since one `redcal` test fails on linux as noted here https://github.com/HERA-Team/hera_cal/issues/812. We should probably figure out what's causing that failure, but for now this fixes the error.

aewallwi (approved):
Looks good too me.

tyler-a-cox:
@jsdillon, I don't think it's been added yet. I'll work on adding it this morning.

jsdillon:
Ok, let's do that. @steven-murray is the maestro for this

tyler-a-cox:
Great, I'll check in with him.

PR #804, opened 2022-06-08
Update to use setuptools-scm, remove current handling of version

jsdillon:


hera_cal/tests/test_utils.py
```diff
@@ -138,6 +138,17 @@ def test_filter_bls(self):
             utils.filter_bls(bls, min_bl_cut=1.5)
 
 
+class TestHistoryVersion(object):
```
steven-murray, line 141: don't use object

steven-murray (commented):
Looks good, one comment, plus delete `hera_cal/_version.py` and add to .gitignore. Oh, also review the README to check the installation instructions.

hera_cal/tests/test_utils.py
jsdillon: Done.

steven-murray (approved):
Great!

PR #814, opened 2022-07-06
fix bug where chunks can be added that are completely flagged.

aewallwi:
Fixes an edge case in `vis_clean.truncate_flagged_edges` in which chunks can be produced that are completely flagged if there is a discontinuity in the x-axis that falls within a completely flagged region.

Also add handling of case where a data waterfall is completely flagged in `vis_clean.vis_clean`.


tyler-a-cox (approved):
This looks good to me. Thanks for catching this!

PR #815, opened 2022-07-13
fix edge case in restore flagged edges but using flags instead of x-a…

aewallwi:
address edge case in `truncate_flagged_edges` and where there is a discontinuity in the x-axis within an completely flagged region.  If there is a discontinuity within a flagged region, then the current behavior is to keep that region in the truncated data which can preserve the edge flags which we need to remove. This PR fixes this by making sure that the flagged "edges" of such regions are the entire width of the region so they don't show up in the truncated data. This has necessitated some minor APi updates to the helper functions `restore_flagged_edges` as well.


tyler-a-cox (approved):
This looks good to me! Thanks for fixing this bug

PR #816, opened 2022-07-22
Add RedDataContainer

jsdillon:
This PR adds a subclass of `DataContainer` called `RedDataContainer` designed to store redundant data, like the visibility solutions that come out of `redcal`. It can be constructed from a `DataConatiner` or a dictionary, and needs to know about the redundancies, either from `reds` or `antpos`. Will make all entries in the underlying dictionary for a given unique baseline group point to the same place in memory.

### DIFF
--- a/hera_cal/datacontainer.py
+++ b/hera_cal/datacontainer.py
@@ -501,10 +501,11 @@ class DataContainer:
 
 
 class RedDataContainer(DataContainer):
+    '''Structure for containing redundant visibilities that can be accessed by any
+        one of the redundant baseline keys (or their conjugate).'''
 
     def __init__(self, data, reds=None, antpos=None, bl_error_tol=1.0):
-        '''Structure for containing redundant visibilities that can be accessed by any
-        one of the redundant baseline keys (or their conjugate).
+        '''Creates a RedDataContainer.
 
         Arguments:
             data: DataContainer or dictionary of visibilities, just as one would pass into DataContainer().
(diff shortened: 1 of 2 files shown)


hera_cal/datacontainer.py
```diff
@@ -498,3 +498,53 @@ def select_or_expand_times(self, new_times, in_place=True):
...
+            self.reds = reds
+        else:
+            from .redcal import get_reds
+            if antpos is not None:
+                self.reds = get_reds(antpos, pols=self.pols(), bl_error_tol=bl_error_tol)
+            elif hasattr(self, 'antpos') and self.antpos is not None:
+                self.reds = get_reds(self.antpos, pols=self.pols(), bl_error_tol=bl_error_tol)
+            else:
+                raise ValueError('Must provide reds, antpos, or have antpos available at data.antpos')
+
```
AaronParsons, line 536: Are we sure that the right way to do this is to make new keys, or do we want to map getitem to the appropriate baseline before handing it off to the super?

hera_cal/datacontainer.py
```diff
@@ -498,3 +498,53 @@ def select_or_expand_times(self, new_times, in_place=True):
...
+                raise ValueError(f'RedDataContainer can only be constructed with (at most) one baseline per group, \
+                                 but this data has the following redundant baselines: {bls_in_data}')
+            if len(bls_in_data) > 0:
+                # check if baseline is in self._data reversed; if so avoid making many copies by conjugating
+                if bls_in_data[0] in self._data:
+                    for bl in red:
+                        self._data[bl] = self._data[bls_in_data[0]]
+                elif reverse_bl(bls_in_data[0]) in self._data:
+                    for bl in red:
+                        self._data[reverse_bl(bl)] = self._data[reverse_bl(bls_in_data[0])]
```
AaronParsons, line 550: is the conjugation taken care of correctly here because _data has the conjugation applied? Does this make a copy of conjugated data? (and for that matter, does a DataContainer make a copy to hold conjugated data?)

AaronParsons (commented):
Happy to approve as is, but just wanted to check about whether it makes more sense to pass getitem through a mapping (to link baselines together), rather than linking data directly. Also had a question of whether conjugated data are stored in separate memory to regular data.

hera_cal/datacontainer.py
jsdillon: This is a good idea. I'll do that.

hera_cal/datacontainer.py
jsdillon: Ok, this is irrelevant because I've taken your other suggestion.

As a general matter, we do not store conjugated data. However, if you ask for a conjugated baseline, you'll get back a copy of that array (conjugated) rather than the original, so if you modify it in place it will not "stick" so to speak.

jsdillon:
Thanks @AaronParsons. This is ready for another look

hera_cal/datacontainer.py
AaronParsons: I'm still confused about the way you did it vs overwritting __getitem__() from DataContainer. Maybe we can discuss.

jsdillon: I did overwrite `__getitem__`

AaronParsons: Sorry, I think I was viewing a cached version of the PR from before vacation.

AaronParsons (approved):

hera_cal/datacontainer.py
jsdillon: No problem.

PR #819, opened 2022-07-30
add use_abs_amp_lincal to argparser

ele-rath:
Add use_abs_amp_lincal to the argparser, with default=True. Also modified scripts/post_redcal_abscal_run.py and hera_cal/abscal.py to include this variable

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -4005,7 +4005,7 @@ def post_redcal_abscal_argparser():
     a.add_argument("--phs_conv_crit", default=1e-6, type=float, help="convergence criterion for updates to iterative phase calibration that compares them to all 1.0s.")
     a.add_argument("--clobber", default=False, action="store_true", help="overwrites existing abscal calfits file at the output path")
     a.add_argument("--verbose", default=False, action="store_true", help="print calibration progress updates")
-    a.add_argument("--skip_abs_amp_lincal", defualt=False, action="store_true", help="finish calibration with an unbiased amplitude lincal step")
+    a.add_argument("--skip_abs_amp_lincal", default=False, action="store_true", help="finish calibration with an unbiased amplitude lincal step")
     a.add_argument("--write_delta_gains", default=False, action="store_true", help="Write degenerate abscal component of gains separately.")
     a.add_argument("--output_file_delta", type=str, default=None, help="Filename to write delta gains too.")
     args = a.parse_args()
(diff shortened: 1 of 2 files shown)


hera_cal/abscal.py
```diff
@@ -3490,7 +3490,7 @@ def post_redcal_abscal(model, data, data_wgts, rc_flags, edge_cut=0, tol=1.0, ke
 
 def post_redcal_abscal_run(data_file, redcal_file, model_files, raw_auto_file=None, data_is_redsol=False, model_is_redundant=False, output_file=None,
                            nInt_to_load=None, data_solar_horizon=90, model_solar_horizon=90, extrap_limit=.5, min_bl_cut=1.0, max_bl_cut=None,
-                           edge_cut=0, tol=1.0, phs_max_iter=100, phs_conv_crit=1e-6, refant=None, clobber=True, add_to_history='', verbose=True,
+                           edge_cut=0, tol=1.0, phs_max_iter=100, phs_conv_crit=1e-6, refant=None, clobber=True, add_to_history='', verbose=True, use_abs_amp_lincal=True,
```
jsdillon, line 3493: New parameters should be added to the docstring for the function.

hera_cal/abscal.py
```diff
@@ -3656,7 +3656,7 @@ def post_redcal_abscal_run(data_file, redcal_file, model_files, raw_auto_file=No
 
                             # run absolute calibration to get the gain updates
                             delta_gains = post_redcal_abscal(model, data, data_wgts, rc_flags_subset, edge_cut=edge_cut, tol=tol,
-                                                             phs_max_iter=phs_max_iter, phs_conv_crit=phs_conv_crit, verbose=verbose)
+                                                             phs_max_iter=phs_max_iter, phs_conv_crit=phs_conv_crit, verbose=verbose, use_abs_amp_lincal)
```
jsdillon, line 3659: This needs to be a keyword argument, so `use_abs_amp_lincal=use_abs_amp_lincal`

hera_cal/abscal.py
```diff
@@ -4004,6 +4004,7 @@ def post_redcal_abscal_argparser():
     a.add_argument("--phs_conv_crit", default=1e-6, type=float, help="convergence criterion for updates to iterative phase calibration that compares them to all 1.0s.")
     a.add_argument("--clobber", default=False, action="store_true", help="overwrites existing abscal calfits file at the output path")
     a.add_argument("--verbose", default=False, action="store_true", help="print calibration progress updates")
+    a.add_argument("--use_abs_amp_lincal", defualt=True, action="store_false", help="finish calibration with an unbiased amplitude lincal step")
```
jsdillon, line 4007: I'd call this `"--skip_abs_amp_lincal"` with default False and action store_true. Since the default behavior is to use abs_amp_lincal, you only want this commandline argument to be used when you want to do something other than the default. You'll also have to fix it in the run script too.

jsdillon (changes requested):

jsdillon (approved):

PR #820, opened 2022-08-02
Read metadata on init with `HERADataFastReader`

jsdillon:
This PR allows `HERADataFastReader` to read metadata on init by default, making them more like `HERAData` (which uses that read on init to enable partial I/O by baseline, time, etc.)

This should make drop-in replacement of `HERAData` with `HERADataFastReader` easier wherever possible/appropriate.


tyler-a-cox (approved):
This looks good to me, thanks @jsdillon

PR #822, opened 2022-08-16
Change default lst-binned visibilities from 1 to 0

jsdillon:
This PR changes the default value for flagged visibilities from 1+0j to 0+0j for the LST-binner. I also updated the redundant averager because having them have different defaults broke a unit test.

This was an issue for me when looking at delay-filtered and then delay-transformed visibilities coming out of the LST-binner using `vis_clean.fft_data()`. I was expecting 0s where there were flags, but I instead had to set them to 0s manually in order to perform this FFT without artifacts.

### DIFF
--- a/hera_cal/lstbin.py
+++ b/hera_cal/lstbin.py
@@ -410,9 +410,9 @@ def lst_bin(data_list, lst_list, flags_list=None, nsamples_list=None, dlst=None,
 
         # fill nans
         d_nan = np.isnan(d_avg)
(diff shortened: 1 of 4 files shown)


steven-murray:
@jsdillon do you still want to merge this? The new `lstbin_simple` module sets the default to `np.nan` (for data), zero for nsamples and, True for flags and `np.inf` for variance.

jsdillon:
No, we can close this.

PR #823, opened 2022-08-20
Minor redcal speed-ups and options to skip steps

jsdillon:
This PR allows skipping parts of redcal while using redundantly calibrate. It has a minor speedup in utils.


tyler-a-cox (approved):
This one seems pretty straightforward, so I'm happy to approve. Thanks, @jsdillon !

PR #824, opened 2022-08-31
Create new `RedSol` class for storing redundant solutions (gains and visibilities) with various useful methods

jsdillon:
This PR introduces a new class `hera_cal.redcal.RedSol()` which stores gains and redundant visibility solutions, the latter using the new `RedDataContainer`. This is meant to work as a drop-in replacement for the combined dictionary of gains and visibilities that could later be separated out using the length of keys.

This PR also changes some of the methods on RedundantCalibrator (omnical, lincal, logcal) to return RedSol objects.

This PR will enable future simplifications to `redundantly_calibrate()` and `expand_omni_sol()`, but I'm leaving that for later.


tyler-a-cox (approved):
This looks great! Excited to start using this soon. Thanks, @jsdillon!

PR #825, opened 2022-09-02
Run polarity flip detection algorithm one polarization at a time

jsdillon:
This PR fixes an issue in polarity flip detection when running firstcal with multiple polarizations simultaneously. It now runs each polarization independently and combines the results.


tyler-a-cox (approved):
Approved. Thanks @jsdillon !

PR #827, opened 2022-09-11
Add min_dim_size feature to redcal.filter_reds

jsdillon:
In working on redcal, I found a case where having just a few antennas in a new degeneracy dimension (the three in the northeast corner here) make redcal go slower and have a weird bistability in the convergence. 

image

And here's that apparent bistability in the phase. 

image

Eliminating those 3 antennas (167, 189, 191) helped a lot. Here's the phase of an antenna after: 
image

In theory, the degeneracies are fixed by firstcal and so shouldn't be allowed to have this kind of phase structure. I'm not 100% sure what's going on. One clue is that the convergence criterion is getting stuck at much higher values than elsewhere in the band at those frequencies:
image

Regardless, it seems like it's useful to enable to user to filter out dimensions that are small with `filter_reds`. This PR does that.


tyler-a-cox (approved):
Good catch, @jsdillon! Thanks for adding this, looks really useful. Happy to approve this.

PR #828, opened 2022-09-13
Delay slope calibration using RFI stations

jsdillon:
This PR is a more constrained version of what @AaronParsons developed [here](https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/fastcal_inspect.ipynb). There, a combination of stable RFI emitters at both ~FM and TV frequencies enabled per-antenna delay calibration as an alternative to firstcal.

In the meantime, some of those higher frequency stations were turned off, making this problem harder. However, I have found that FM alone is sufficient to solve for the degeneracies of firstcal, namely one per-polarization and per-dimension delay gradient across the array. 

Here's a set of measured per-baseline delays after redundant calibration:

image

And here's what I get after performing this calibration.

image

Basically, the delays are flattened to approximately the resolution that they are measured (0.1 ns).

### DIFF
--- a/ci/hera_cal_tests.yml
+++ b/ci/hera_cal_tests.yml
@@ -8,7 +8,7 @@ dependencies:
   - astropy>=5.0.4
   - h5py
   - hdf5plugin
-  - numpy>=1.20,<1.23
+  - numpy>=1.23
   - pytest
   - scipy
   - scikit-learn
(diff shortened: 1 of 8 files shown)


tyler-a-cox (approved):
This looks good to me! Thanks @jsdillon

PR #830, opened 2022-09-21
Add utility functions for `nucal`

tyler-a-cox:
This PR creates the nucal file, which contains utility functions for performing frequency redundancy calibration. Functions for finding groups of baselines which are radially redundant and filtering those baselines are included here.

### DIFF
--- a/hera_cal/nucal.py
+++ b/hera_cal/nucal.py
@@ -131,7 +131,7 @@ def get_u_bounds(radial_reds, antpos, freqs):
             
 
 def get_unique_orientations(
-    antpos, reds=None, pols=["nn"], min_ubl_per_orient=1, blvec_error_tol=1e-3, bl_error_tol=1.0,
+    antpos, reds, min_ubl_per_orient=1, blvec_error_tol=1e-3, bl_error_tol=1.0,
 ):
     """
     Sort baselines into groups with the same radial heading. These groups of baselines are potentially
@@ -143,8 +143,6 @@ def get_unique_orientations(
         Antenna positions in the form {ant_index: np.array([x,y,z])}.
     reds : list of lists
         List of lists of spatially redundant baselines in the array. Can be found using redcal.get_reds
-    pols : list, default=['nn']
-        A list of polarizations e.g. ['nn', 'ne', 'en', 'ee']
     min_ubl_per_orient : int, default=1
         Minimum number of baselines per unique orientation
     blvec_error_tol : float, default=1e-3
@@ -159,8 +157,8 @@ def get_unique_orientations(
     uors : list of lists of tuples
         List of list of tuples that are considered to be radially redundant
     """
-    if reds is None:
-        reds = redcal.get_reds(antpos, pols=pols, bl_error_tol=bl_error_tol)
+    # Get polarizations from reds
+    pols = list(set([rdgrp[0][-1] for rdgrp in reds]))
 
     _uors = {}
     for pol in pols:
--- a/hera_cal/nucal.py
+++ b/hera_cal/nucal.py
@@ -434,8 +434,6 @@ class FrequencyRedundancy:
     def filter_radial_groups(
         self,
         min_nbls=1,
-        pols=None,
-        ex_pols=None,
         min_bl_cut=None,
         max_bl_cut=None,
     ):
--- a/hera_cal/nucal.py
+++ b/hera_cal/nucal.py
@@ -97,6 +97,10 @@ def get_unique_orientations(
         The largest allowable difference between baselines in a redundant group
         (in the same units as antpos). Normally, this is up to 4x the largest antenna position error.
 
+    Returns:
+    -------
+    uors : list of lists of tuples
+        List of list of tuples that are considered to be radially redundant
     """
     if reds is None:
         reds = redcal.get_reds(antpos, pols=pols, bl_error_tol=bl_error_tol)
@@ -148,10 +152,29 @@ def get_unique_orientations(
 
 
 class RadialRedundantGroup:
-    """ """
+    """List-like object that holds tuples of baselines that are assumed to be
+    radially redundant (have the same heading). In addition to supporting list
+    like behavior, this object also gets the minimum and maximum u-mode magnitude
+    of the radially redundant group and filter the group based on a number of factors.
+    """
 
     def __init__(self, baselines, antpos, blvec=None, pol=None):
-        """ """
+        """
+        Create a RadialRedundantGroup object from a list baselines assumed
+        to be radially redundant
+
+        Parameters:
+        ----------
+        baselines : list of tuples
+            List
+        antpos : dict
+            Antenna positions in the form {ant_index: np.array([x,y,z])}.
+        blvec : np.ndarray
+            Normalized baseline vector for the radially redundant group. If one is not
+            provided, it will be estimated from the antenna positions
+        pol : str
+            Polarization of the baseline group
+        """
         _baselines = deepcopy(baselines)
 
         # Attach polarization and normalized vector to radial redundant group
@@ -193,15 +216,16 @@ class RadialRedundantGroup:
 
         Parameters:
         ----------
-            freqs: np.ndarray
-                Array of frequencies found in the data in units of Hz
+        freqs: np.ndarray
+            Array of frequencies found in the data in units of Hz
 
         Returns:
-            ubounds: tuple
-                Tuple of the magnitude minimum and maximum u-modes sampled by this baseline group
+        -------
+        ubounds: tuple
+            Tuple of the magnitude minimum and maximum u-modes sampled by this baseline group
         """
-        umin = freqs.min() / 2.998e8 * np.min(self.baseline_lengths)
-        umax = freqs.max() / 2.998e8 * np.max(self.baseline_lengths)
+        umin = freqs.min() / SPEED_OF_LIGHT * np.min(self.baseline_lengths)
+        umax = freqs.max() / SPEED_OF_LIGHT * np.max(self.baseline_lengths)
         return (umin, umax)
 
     def filter_group(
@@ -210,23 +234,44 @@ class RadialRedundantGroup:
         ex_bls=None,
         ants=None,
         ex_ants=None,
-        ubls=None,
-        ex_ubls=None,
         pols=None,
         ex_pols=None,
-        antpos=None,
         min_bl_cut=None,
         max_bl_cut=None,
     ):
-        """ """
+        """
+        Filter radially redundant group to include/exclude the specified bls, antennas. and polarizations.
+        Arguments are evaluated, in order of increasing precedence: (pols, ex_pols, bls, ex_bls, ants, ex_ants,
+        min_bl_cut, max_bl_cut).
+
+        Parameters:
+        ----------
+        bls : list of tuples, default=None
+            baselines to include. Baselines of the form (i,j,pol) include that specific
+            visibility.  Baselines of the form (i,j) are broadcast across all polarizations present in reds.
+        ex_bls : list of tuples, default=None
+            same as bls, but excludes baselines.
+        ants : list of tuples, default=None
+            antennas to include. Only baselines where both antenna indices are in ants
+            are included.  Antennas of the form (i,pol) include that antenna/pol. Antennas of the form i are
+            broadcast across all polarizations present in reds.
+        ex_ants : list of tuples, default=None
+            same as ants, but excludes antennas.
+        pols : list of strings
+            polarizations to include in reds. e.g. ['nn','ee','ne','en']
+        ex_pols : list of strings
+            same as pols, but excludes polarizations.
+        min_bl_cut:
+            Cut baselines in the radially redundant group with lengths less than min_bl_cut
+        max_bl_cut:
+            Cut baselines in the radially redundant group with lengths less than min_bl_cut
+        """
         _baselines = redcal.filter_reds(
             [self._baselines],
             bls=bls,
             ex_bls=ex_bls,
             ants=ants,
             ex_ants=ex_ants,
-            ubls=ubls,
-            ex_ubls=ex_ubls,
             pols=pols,
             ex_pols=ex_pols,
         )
@@ -262,7 +307,12 @@ class RadialRedundantGroup:
 
 
 class FrequencyRedundancy:
-    """ """
+    """List-like object that contains groups RadialRedundantGroup objects.
+    Functions similarly to the output of redcal.get_reds for frequency redundant
+    calibration. In addition to mimicking list functionality, this object also filters
+    radially redundant groups based on a number of factors, can get specific polarizations, and
+    radially redundant and spatially redundant groups by baseline key.
+    """
 
     def __init__(
         self, antpos, reds=None, blvec_error_tol=1e-9, pols=["nn"], bl_error_tol=1.0
@@ -290,11 +340,18 @@ class FrequencyRedundancy:
 
     def get_radial_group(self, key):
         """
-        Get baselines with the same heading as a given baseline
+        Get baselines with the same radal heading as a given baseline
 
         Parameters:
-            key: tuple
-                Basleine key of type (ant1, ant2, pol)
+        ----------
+        key : tuple
+            Baseline key of type (ant1, ant2, pol)
+
+        Returns:
+        -------
+        group : list of tuples
+            List of baseline tuples that have the same radial headings
+
         """
         # Identify headings
         for group_key in self._mapped_reds:
@@ -311,8 +368,13 @@ class FrequencyRedundancy:
         Get a list of baseline that are spatially redundant with the input baseline
 
         Parameters:
-            key: tuple
-                Baseline key with of type (ant1, ant2, pol)
+        ----------
+        key: tuple
+            Baseline key with of type (ant1, ant2, pol)
+        Returns:
+        -------
+        group: list of tuples
+            Return baseline tuples that are spatially redundant
         """
         for group_key in self._mapped_reds:
             if key == group_key or key in self._mapped_reds[group_key]:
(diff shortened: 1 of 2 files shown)


hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+            baselines to include. Baselines of the form (i,j,pol) include that specific
+            visibility.  Baselines of the form (i,j) are broadcast across all polarizations present in reds.
+        ex_bls : list of tuples, default=None
+            same as bls, but excludes baselines.
+        ants : list of tuples, default=None
+            antennas to include. Only baselines where both antenna indices are in ants
+            are included.  Antennas of the form (i,pol) include that antenna/pol. Antennas of the form i are
+            broadcast across all polarizations present in reds.
+        ex_ants : list of tuples, default=None
+            same as ants, but excludes antennas.
```
jsdillon, line 259: I think it's confusing to offer this option, because usually these objects will be constructed with unique baselines, in which case filtering by antennas is potentially misleading. If you really want this to be possible, I'd suggest adding a `**filter_kwargs` which passes into redcal.filter_reds and lets you do other options you don't explicitly call out here.

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+            pols=pols,
+            ex_pols=ex_pols,
+        )
+        if len(_baselines) == 0:
+            self._baselines = []
+            self.baseline_lengths = []
+        else:
+            new_bls = []
+            new_bls_lengths = []
+            for bls in _baselines[0]:
```
jsdillon, line 284: again, confusing to have this be `bls` and not `bl`

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+                index = self._baselines.index(bls)
```
jsdillon, line 285: doing all this index searching might be slower than recomputing baseline lengths from antpos, which honestly you probably want to store internally in the __init__

jsdillon: If lengths are stored via a dictionary, it'd be faster. That's another option.

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+        return iter(self._baselines)
+
+    def __len__(self):
+        """Return the length of the baselines list"""
+        return len(self._baselines)
+
+    def __getitem__(self, index):
+        """Get the baseline at the chosen index"""
+        return self._baselines[index]
+
```
jsdillon, line 307: I'd suggest supporting __setitem__, though there may be other list methods you also want to support (see https://www.w3schools.com/python/python_ref_list.asp)

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+            self._radial_groups.pop(_bad_group)
+
+    def __len__(self):
+        """Get number of frequency redundant groups"""
+        return len(self._radial_groups)
+
+    def __getitem__(self, index):
+        """Get RadialRedundantGroup object from list of unique orientations"""
+        return self._radial_groups[index]
+
```
jsdillon, line 470: again, I'd add a __setitem__ and possibly other list-like methods

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+            baselines to include. Baselines of the form (i,j,pol) include that specific
+            visibility.  Baselines of the form (i,j) are broadcast across all polarizations present in reds.
+        ex_bls : list of tuples, default=None
+            same as bls, but excludes baselines.
+        ants : list of tuples, default=None
+            antennas to include. Only baselines where both antenna indices are in ants
+            are included.  Antennas of the form (i,pol) include that antenna/pol. Antennas of the form i are
+            broadcast across all polarizations present in reds.
+        ex_ants : list of tuples, default=None
+            same as ants, but excludes antennas.
```
jsdillon, line 432: Same comment as before on `ants` and `ex_ants`

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+    def __len__(self):
+        """Return the length of the baselines list"""
+        return len(self._baselines)
+
+    def __getitem__(self, index):
+        """Get the baseline at the chosen index"""
+        return self._baselines[index]
+
+
+class FrequencyRedundancy:
```
jsdillon, line 309: Looking over both `FrequencyRedundancy` and `RadialRedundantGroup`, I'm really wondering whether you need both. Right now the only thing `RadialRedundantGroup` does that's special is `get_u_bounds.`

I'm wondering if you can make your life easier by just defining `FrequencyRedundancy` as a special list-like object and having the radially redundant groups under the hood just be lists? Your call.

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+
+        self._mapped_reds = {red[0]: red for red in reds}
+
+        self._radial_groups = get_unique_orientations(
+            antpos, reds=reds, pols=pols, blvec_error_tol=blvec_error_tol
+        )
+
+    def get_radial_group(self, key):
+        """
+        Get baselines with the same radal heading as a given baseline
```
jsdillon, line 343: typo "radal"

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+            Baseline key of type (ant1, ant2, pol)
+
+        Returns:
+        -------
+        group : list of tuples
+            List of baseline tuples that have the same radial headings
+
+        """
+        # Identify headings
+        for group_key in self._mapped_reds:
```
jsdillon, line 357: Not great that you have to do this loop every time you give it a single key. Seems to me like there should be a dictionary under the hood that figures this out once and for all.

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+        else:
+            raise KeyError(
+                f"Baseline {key} is not in the group of spatial redundancies"
+            )
+
+    def get_pol(self, pol):
+        """Get all radially redundant groups with a given polarization"""
+        for group in self:
+            if group.pol == pol:
+                yield group
```
jsdillon, line 401: Is there a reason this is an iterator? It's useful when you're doing loops, but it's a bit weird for a funciton called `get_pol` instead of `iterate_over_pols` or similar.

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+
+    def filter_radial_groups(
+        self,
+        min_nbls=4,
```
jsdillon, line 405: where does this default come from? Shouldn't it be 1 so it's minimally invasive?

jsdillon (changes requested):
Ok, left a bunch of comments and suggestions. I haven't looked at the unit tests yet but I figure I'll wait until the interfaces are more settled.

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+        ex_bls : list of tuples, default=None
+            same as bls, but excludes baselines.
+        ants : list of tuples, default=None
+            antennas to include. Only baselines where both antenna indices are in ants
+            are included.  Antennas of the form (i,pol) include that antenna/pol. Antennas of the form i are
+            broadcast across all polarizations present in reds.
+        ex_ants : list of tuples, default=None
+            same as ants, but excludes antennas.
+        pols : list of strings
+            polarizations to include in reds. e.g. ['nn','ee','ne','en']
```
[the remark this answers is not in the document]
tyler-a-cox: Good point. Removed this

hera_cal/nucal.py
tyler-a-cox: Fixed this typo

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+import numpy as np
+from copy import deepcopy
+import astropy.constants as const
+from collections import defaultdict
+from scipy.cluster.hierarchy import fclusterdata
+
+SPEED_OF_LIGHT = const.c.si.value
+
+
+def is_frequency_redundant(bl1, bl2, freqs, antpos, blvec_error_tol=1e-9):
```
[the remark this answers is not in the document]
tyler-a-cox: Added `is_same_orientation` function

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+        Largest allowable euclidean distance a unit baseline vector can be away from an existing
+        cluster to be considered a unique orientation. See "fclusterdata" for more details.
+
+    Returns:
+        Boolean value determining whether or not the baselines are frequency
+        redundant
+
+    """
+    # Split baselines in component antennas
+    _ant1, _ant2, _pol = bl1
```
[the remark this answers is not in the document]
tyler-a-cox: Renamed these

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+
+    # Check headings
+    norm_vec1 = blvec1 / np.linalg.norm(blvec1)
+    norm_vec2 = blvec2 / np.linalg.norm(blvec2)
+    clusters = fclusterdata(
+        np.array([norm_vec1, norm_vec2]), blvec_error_tol, criterion="distance"
+    )
+    cond4 = clusters[0] == clusters[1]
+
+    return (cond1 or cond2) and cond3 and cond4
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed this

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+        and blmag1 * freqs.max() >= blmag2 * freqs.min()
+    )
+
+    # Check polarization match
+    cond3 = _pol == pol
+
+    # Check headings
+    norm_vec1 = blvec1 / np.linalg.norm(blvec1)
+    norm_vec2 = blvec2 / np.linalg.norm(blvec2)
+    clusters = fclusterdata(
```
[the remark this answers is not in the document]
tyler-a-cox: Using np.linalg.norm now. I was using fclusterdata for consistency before but this should be fine.

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+
+    return (cond1 or cond2) and cond3 and cond4
+
+
+def get_unique_orientations(
+    antpos,
+    reds=None,
+    pols=["nn"],
+    min_ubl_per_orient=1,
+    blvec_error_tol=1e-9,
```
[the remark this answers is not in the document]
tyler-a-cox: Changed this

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+
+    Returns:
+    -------
+    uors : list of lists of tuples
+        List of list of tuples that are considered to be radially redundant
+    """
+    if reds is None:
+        reds = redcal.get_reds(antpos, pols=pols, bl_error_tol=bl_error_tol)
+    else:
+        reds = deepcopy(reds)
```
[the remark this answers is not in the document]
tyler-a-cox: I was doing this to prevent tuples in reds from being flipped when I'm flipping orientations later in the code. I don't actually think it'll be a problem so I removed it

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+    ----------
+    bl1 : tuple
+        Tuple of antenna indices and polarizations of the first baseline
+    bl2 : tuple
+        Tuple of antenna indices and polarizations of the second baseline
+    freqs : np.ndarray
+        Array of frequencies found in the data in units of Hz
+    antpos : dict
+        Antenna positions in the form {ant_index: np.array([x,y,z])}.
+    blvec_error_tol : float, default=1e-9
```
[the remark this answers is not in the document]
tyler-a-cox: Changed this

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+):
+    """
+    Sort baselines into groups with the same radial heading. These groups of baselines are potentially
+    frequency redundant in a similar way to redcal.get_reds does. Returns a list of RadialRedundantGroup objects
+
+    Parameters:
+    ----------
+    antpos : dict
+        Antenna positions in the form {ant_index: np.array([x,y,z])}.
+    reds : list of lists
```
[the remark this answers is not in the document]
tyler-a-cox: Changed this to accept antpos and reds and infer pols from the reds

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+    else:
+        reds = deepcopy(reds)
+
+    _uors = {}
+    for pol in pols:
+        ubl_pairs = [red[0] for red in reds if red[0][-1] == pol]
+
+        # Compute normalized baseline vectors
+        normalized_vecs = []
+        for bls in ubl_pairs:
```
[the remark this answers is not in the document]
tyler-a-cox: Changed this

hera_cal/nucal.py
tyler-a-cox: Changed this so it just returns the list of baseline groups that have a give polarization

hera_cal/nucal.py
tyler-a-cox: Set to 1

hera_cal/nucal.py
tyler-a-cox: Removed `RadialRedundantGroup` and added some of its functionality to `FrequencyRedundancy`

hera_cal/nucal.py
tyler-a-cox: I just decided to simplify this and only include filtering by baseline length and number of unique baselines per radially redundant group

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+        # Cluster orientations
+        clusters = fclusterdata(normalized_vecs, blvec_error_tol, criterion="distance")
+        uors = [[] for i in range(np.max(clusters))]
+
+        for cluster, bl in zip(clusters, ubl_pairs):
+            uors[cluster - 1].append(bl)
+
+        uors = sorted(uors, key=len, reverse=True)
+
+        # Find clusters with headings anti-parallel to others
```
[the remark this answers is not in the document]
tyler-a-cox: Changed this

hera_cal/nucal.py
```diff
@@ -0,0 +1,473 @@
...
+    uors : list of lists of tuples
+        List of list of tuples that are considered to be radially redundant
+    """
+    if reds is None:
+        reds = redcal.get_reds(antpos, pols=pols, bl_error_tol=bl_error_tol)
+    else:
+        reds = deepcopy(reds)
+
+    _uors = {}
+    for pol in pols:
```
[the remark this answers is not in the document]
tyler-a-cox: I may just try and optimize this later. This only takes slightly longer than `get_reds` on average anyway. I think I'd rather move on to the calibration code.

hera_cal/nucal.py
```diff
@@ -0,0 +1,464 @@
...
+        self._radial_groups[index] = value
+
+        # Add baseline lengths to length dictionary
+        for bl in value:
+            ant1, ant2, _ = bl
+            blmag = np.linalg.norm(self.antpos[ant2] - self.antpos[ant1])
+            self.baseline_lengths[bl] = blmag
+
+        # Reset baseline mapping to spectrally redundant groups
+        self._reset_mapping_dictionaries()
```
jsdillon, line 432: Having to do this every time you append is very slow. Can you instead just update the mapping dictionaries?

hera_cal/nucal.py
```diff
@@ -0,0 +1,464 @@
...
+        self._radial_groups.append(value)
+
+        # Add baseline lengths to length dictionary
+        for bl in value:
+            ant1, ant2, _ = bl
+            blmag = np.linalg.norm(self.antpos[ant2] - self.antpos[ant1])
+            self.baseline_lengths[bl] = blmag
+
+        # Reset baseline mapping to spectrally redundant groups
+        self._reset_mapping_dictionaries()
```
jsdillon, line 456: Same comment as before.

hera_cal/nucal.py
```diff
@@ -0,0 +1,464 @@
...
+
+    def append(self, value):
+        """
+        Append value to the end of _radial_groups. Also raises an error if a baseline with the same
```
jsdillon, line 436: I'd clarify what type `value` is supposed to be.

hera_cal/nucal.py
```diff
@@ -0,0 +1,464 @@
...
+            self.baseline_lengths[bl] = blmag
+
+        # Reset baseline mapping to spectrally redundant groups
+        self._reset_mapping_dictionaries()
+    
+    def __iter__(self):
+        """Iterates through the list of redundant groups"""
+        return iter(self._radial_groups)
+
+    def sort(self, key=None, reverse=True):
```
jsdillon, line 462: The key argument is just ignored. Is that intentional?

`key=(len if key is None else key)` might work

jsdillon (changes requested):
Looking a lot better! Just a handful of minor changes.

hera_cal/nucal.py
```diff
@@ -0,0 +1,464 @@
...
+        for group in uors:
+            _uors[group[0]] = group
+
+    # Convert lists to RadialRedundantGroup objects
+    uors = [_uors[key] for key in _uors if len(_uors[key]) >= min_ubl_per_orient]
+    return sorted(uors, key=len, reverse=True)
+
+
+class FrequencyRedundancy:
+    """List-like object that contains groups RadialRedundantGroup objects.
```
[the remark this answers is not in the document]
tyler-a-cox: Just changed this docstring

hera_cal/nucal.py
tyler-a-cox: Good point. I just replace this with appending to dictionaries directly

hera_cal/nucal.py
tyler-a-cox: Same as above

hera_cal/nucal.py
tyler-a-cox: Clarified this

hera_cal/nucal.py
tyler-a-cox: Yeah, good point. Just changed this to your suggestion

tyler-a-cox:
Thanks for the review, Josh! I think everything should be good to go now.

hera_cal/nucal.py
```diff
@@ -330,7 +330,7 @@ def get_pol(self, pol):
         """Get all radially redundant groups with a given polarization"""
         return [group for group in self if group[0][-1] == pol]
 
-    def filter_radial_groups(self, min_nbls=1, min_bl_cut=None, max_bl_cut=None):
+    def filter_radial_groups(self, min_nbls=1, min_bl_cut=None, max_bl_cut=None, min_u_cut=None, max_u_cut=None):
         """
         Filter each radially redundant group to include/exclude the baselines based on baseline length.
         Radially redundant groups can also be completely filtered based on the number of baselines in
```
jsdillon, line 336: The parameters in this docstring are out of date.

jsdillon (changes requested):
Just one docstring fix, then we're good to go.

hera_cal/nucal.py
tyler-a-cox: Removed those parameters for now. I was planning on adding that later, but I'll just make that part of another PR

jsdillon (approved):

PR #831, opened 2022-09-22
Speed up `expand_omni_sol`

jsdillon:
This PR speeds up expand_omni_sol by avoiding (if possible) a chi^2 calculation that's unnecessary under normal redcal operation (for example, when not using a min_bl_cut or max_bl_cut). It also adds a dictionary to a RedundantCalibrator object that prevents a lot of unnecessarily repeated list comprehension.

Here's the before:

image

(This takes 31.4 s on my laptop when not running inside snakeviz.)

Here's the after:

Screen Shot 2022-09-21 at 7 01 59 PM

(This takes 13.2 s on my laptop when not running inside snakeviz.)

### DIFF
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -950,7 +950,7 @@ class RedundantCalibrator:
                 even when allowing for an arbitrary number of phase slope degeneracies.
         """
 
-        self.reds = reds
+        self._set_reds(reds)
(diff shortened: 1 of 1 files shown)


tyler-a-cox (approved):
Okay, this seems pretty straightforward. Thanks for adding this!

PR #842, opened 2022-10-12
Make redcal_run_write_results write omnical visibilities with the same baselines for both pols

jsdillon:
This was breaking RTP when most of one polarization was flagged but not the other.

Now missing unique baseline visibilities are in the output file, but flagged (with zero nsamples).


jsdillon:
@r-pascua verbally approved this PR subject to tests passing

PR #844, opened 2022-10-16
Introduced RedSol.extend_ubls and extend_gains for expanding redundan…

AaronParsons:
…t solutions, as well as some infrastructure in RedDataContainer to allow redunant visibility sets to be expanded and rekeyed incrementally.


AaronParsons:
@jsdillon I've got the basics of this in place, but there's a lot of cleanup downstream of redundantly_calibrate that is necessary before we can deal with expand_omni_sol (hopefully to remove it). I've put lots of 'XXX' flags in the code. Can you take a look and then we can discuss where things are unclear? If we are moving away from the script interface, maybe we can just deprecate and eventually remove the rest.

jsdillon:
@AaronParsons This is ready for another look. 

I've piped the new interfaces through to the notebook, a draft of which is here: https://github.com/HERA-Team/hera_notebook_templates/blob/file_calibration_TT_phs_abscal/notebooks/file_calibration.ipynb

The result (on my mac) is that the notebook is about 20% faster. This comes from a large speedup in expanding the gains to bad antennas, as well as a small speedup from skipping logcal.

I've verified that the results are qualitatively very similar to before the update. I don't have a ton more time to work on this, though I plan on doing some more quantitative checks asap.

jsdillon:
Looks like the results are very, very similar. 

Here's chi^2 before:
image

And here's chi^2 after:
image

And here's another spot check of an unflagged antenna:
image

And of a flagged one:
image

PR #846, opened 2022-11-04
Lstbin improvements

steven-murray:
This PR adds a new LST-binner that is designed to have similar external behaviour to the original, but is simpler internally, and has better performance.

The main thing changed is that it is much more vectorized -- this helps both with memory management and speed. The algorithm is still roughly the same: 

```
configure_output_files()
all_baselines = get_all_baselines_from_all_inputs()
for output_file:
    output_data = (nlsts_in_file, nbls, npols, nfreqs)
    for bl_chunk in nchunks:
        input_data = np.zeros(ntimes, nbls_in_chunk, npols, nfreqs)
        for fl in input_fls:
            input_data[these_times, these_baselines, :, :] =  read_data_within_this_output_file()
        output_data[:, these_baselines, :, :] = do_lst_binning(input_data)
    write_out_file(output_data, output_file)
```
The `do_lst_binning` function is really very simple: just a `np.digitize` and then iterate through each digit and average.

One thing to note is that the `get_all_baselines_from_all_inputs()` part of the algorithm (called `get_all_unflagged_baselines` in the code) is fast custom-code that reads part of the header of each input file, and gathers ALL baselines present in any file that has data associated. Since this is done outside the "outfile" loop, it means that all output files get the same set of baselines, regardless of whether they have data for that file or not. Of course, if they have no data, they will be flagged appropriately.

The PR also fixes a huge problem in `io.write_vis` in which there is a `map(lambda: ...))` when trying to convert a `DataContainer` to a standard numpy array. This took upwards of 40min to complete, instead of the ~1sec it should take. I also wrote a new function that turns arrays of data into `UVdata` objects (though, this seemed like something that should have been solved before in a better way).


hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+        return fnc
+
+logger = logging.getLogger(__name__)
+
+@profile
+def simple_lst_bin(
+    data: np.ndarray,
+    data_lsts: np.ndarray,
+    baselines: list[tuple[int, int]],
+    lst_bin_edges: np.ndarray,
```
nkern, line 42: this doesn't specify whether this is left or right edges. it seems like it would be easier for the user to provide bin centers (as is the protocol for pyuvdata and UVData objects) and then when you got to run `np.digitize` you shift by width/2 to feed it bin edges. lmk what you think

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+            if np.any(tind):
+                tinds.append(tind)
+                time_arrays.append(tarr[tind])
+                all_lsts.append(lsts[tind])
+                file_list.append(meta)
+                cals.append(cal)
+    return tinds, time_arrays, all_lsts, file_list, cals
+
+
+def make_lst_grid(
```
nkern, line 719: why not load this from lstbin.py?

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+@profile
+def lst_average(
+    data: np.ndarray, nsamples: np.ndarray, flags: np.ndarray, 
+    flag_thresh: float = 0.7, median: bool = False,
+    mutable: bool = False,
+    sigma_clip_thresh: float = 0.0,
+    sigma_clip_min_N: int=4,
+) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    """
+    Compute statistics of a set of data over its first axis.
```
nkern, line 310: "zeroth axis"?

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+    at the same LST. However, this function is agnostic to the meaning of the first
+    axis. It just computes the mean, std, and nsamples over the first axis.
+
+    This function is meant to be used on a single element of a list returned by
+    :func:`simple_lst_bin`.
+
+    Parameters
+    ----------
+    data
+        The data to compute the statistics over. Shape ``(ntimes, nbl, nfreq, npol)``.
```
nkern, line 322: shouldn't this be nominally "(nnights, ntimes, nbl, nfreq, npol)"? or am I confused?

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+    Use this function to reduce lists of arrays with multiple time integrations per bin
+    (i.e. the output of :func:`simple_lst_bin`) to arrays of shape 
+    ``(nbl, nlst_bins, nfreq, npol)``. For example, compute the mean/std.
+
+    Parameters
+    ----------
+    data
+        The data to perform the reduction over. The length of the list is the number
+        of LST bins. Each array in the list should have shape 
+        ``(nbl, ntimes_per_lst, nfreq, npol)``.
```
nkern, line 230: (nbl, nsamples_per_lst, nfreq, npol)

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+        raise ValueError(
+            "lst_bin_edges must be monotonically increasing."
+        )
+
+    # Now ensure that all the observed LSTs are wrapped so they start above the first bin edges
+    grid_indices, data_lsts, lst_mask = get_lst_bins(data_lsts, lst_bin_edges)
+    lst_bin_centres = (lst_bin_edges[1:] + lst_bin_edges[:-1])/2
+
+    # TODO: check whether this creates a data copy. Don't want the extra RAM...
+    data = data[lst_mask]  # actually good if this is copied, because we do LST rephase in-place
```
nkern, line 141: can we rectify these comments, and if you do want this to be a copy let's make this explicit via deepcopy

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+    arrays, each containing a single LST bin. Each of the data arrays in each bin
+    are also rephased onto a common LST grid.
+
+    Parameters
+    ----------
+    data
+        The visibility data. Must be shape (ntimes, nbls, nfreqs, npols)
+    data_lsts
+        The LSTs corresponding to each of the time stamps in the data. Must have
+        length ``data.shape[0]``
```
nkern, line 63: specify this is bin centers

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+    rephase
+        Whether to apply re-phasing to the data, to bring it to a common LST grid.
+    antpos
+        3D Antenna positions for each antenna in the data.
+    lat
+        The latitude (in degrees) of the telescope.
+
+    Returns
+    -------
+    data
```
nkern, line 83: also returns lst_bin_centers

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
+"""
+An attempt at a simpler LST binner that makes more assumptions but runs faster.
+
+In particular, we assume that all baselines have the same time array and frequency array,
+and that each is present throughout the data array. This allows a vectorization.
```
nkern, line 5: Can you add, 

```
The main binning function is `lst_bin_files_for_baselines` which has a similar but not identical API as `lstbin.lstbin`
```

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+    Performs the adjustment in-place.
+    """
+    while lst_bin_edges[0] < 0:
+        lst_bin_edges += 2*np.pi
+    while lst_bin_edges[0] >= 2*np.pi:
+        lst_bin_edges -= 2*np.pi
+
+def lst_bin_files_for_baselines(
+    data_files: list[Path | FastUVH5Meta], 
+    lst_bin_edges: np.ndarray,
```
nkern, line 403: same comment about bin edges here

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+    nsamples = np.full(shape, n, dtype=float)
+    return data, flags, nsamples
+
+@profile
+def lst_average(
+    data: np.ndarray, nsamples: np.ndarray, flags: np.ndarray, 
+    flag_thresh: float = 0.7, median: bool = False,
+    mutable: bool = False,
+    sigma_clip_thresh: float = 0.0,
+    sigma_clip_min_N: int=4,
```
nkern, line 307: we also had a `flag_below_min_N` bool before, not sure how hard it is to inculde this

steven-murray: That's easy. Done.

hera_cal/lstbin_simple.py
steven-murray: These are *all* the edges, i.e. the bins are *assumed* to be contiguous, but not necessarily all the same size. The output number of bins will be `len(lst_bin_edges) - 1`. I will clarify this in the docstring. I think this is the most reasonable way to input the bins, as it is more general than just sending in bin-centres, and (IMHO) more easy to reason about.

hera_cal/lstbin_simple.py
steven-murray: I added some documentation mentioning that it actually doesn't matter what the ``data_lsts`` represent. The key assumption is that the data is *treated* as if the integrations occur at a snapshot, i.e. a delta function in LST at the given ``data_lst``. Thus, the whole integration is put into one LST bin, regardless of whether in reality it might straddle between two bins. Whether the data LST is at the start, middle or end of the actual integration will incur slightly different errors in detail, but qualitatively similar.

hera_cal/lstbin_simple.py
steven-murray: I actually think the function that retains the most similarity to the old lstbin module is the `lst_bin_files`, which is different in the new code because it receives a pre-configured "config" file specifying the files to use (and the output LST grid), but otherwise serves the same function (i.e. takes in files, outputs files). There is actually nothing quite similar to the old `lstbin.lst_bin`, which took in a list of data *objects* and returned a list of LST-binned data and their averages. To acheive that in the new code, you would call `lst_align` followed by `reduce_lst_bins`. In contrast, `lst_bin_files_for_baselines` also takes as input the data *files*, and outputs non-averaged lst-bins only. 

I've added some comments in the module docstring to make the point that the main entry-point is the `lst_bin_files` function (and that it is similar to the function of the same name in the old module). I could add more docs about how to use the module as a whole, if you like. I could also add a function that is more similar to the old `lst_bin()`, which would be a small wrapper around the existing functions. It wouldn't be super-duper easy, though not too hard.

hera_cal/lstbin_simple.py
steven-murray: I've used `nintegrations_in_lst`. What do you think?

hera_cal/lstbin_simple.py
steven-murray: I determined what's actually going on, and wrote more extensive comments in-line. The `lst_mask` indexing does do a copy, so we don't want to `deepcopy` because that'd be yet another copy.

hera_cal/lstbin_simple.py
steven-murray: Again, I've used `nintegrations_in_lst` instead of `ntimes_per_lst`. Lmk what you think

hera_cal/lstbin_simple.py
steven-murray: I've updated the docstring to match the one for `lst_align` based on your previous comment. I still think that edges are better, but I guess I could be persuaded.

hera_cal/lstbin_simple.py
steven-murray: I was doing that, but ended up wanting to have the new module fully distinct from the old, since at some time we must just replace the old one with this.

hera_cal/lstbin_simple.py
steven-murray: These are *all* the edges, i.e. the bins are *assumed* to be contiguous, but not necessarily all the same size. The output number of bins will be `len(lst_bin_edges) - 1`. I will clarify this in the docstring. I think this is the most reasonable way to input the bins, as it is more general than just sending in bin-centres, and (IMHO) more easy to reason about.

hera_cal/lstbin_simple.py
steven-murray: Good catch.

hera_cal/lstbin_simple.py
steven-murray: I added some documentation mentioning that it actually doesn't matter what the ``data_lsts`` represent. The key assumption is that the data is *treated* as if the integrations occur at a snapshot, i.e. a delta function in LST at the given ``data_lst``. Thus, the whole integration is put into one LST bin, regardless of whether in reality it might straddle between two bins. Whether the data LST is at the start, middle or end of the actual integration will incur slightly different errors in detail, but qualitatively similar.

hera_cal/lstbin_simple.py
steven-murray: Good catch.

hera_cal/lstbin_simple.py
steven-murray: I actually think the function that retains the most similarity to the old lstbin module is the `lst_bin_files`, which is different in the new code because it receives a pre-configured "config" file specifying the files to use (and the output LST grid), but otherwise serves the same function (i.e. takes in files, outputs files). There is actually nothing quite similar to the old `lstbin.lst_bin`, which took in a list of data *objects* and returned a list of LST-binned data and their averages. To acheive that in the new code, you would call `lst_align` followed by `reduce_lst_bins`. In contrast, `lst_bin_files_for_baselines` also takes as input the data *files*, and outputs non-averaged lst-bins only. 

I've added some comments in the module docstring to make the point that the main entry-point is the `lst_bin_files` function (and that it is similar to the function of the same name in the old module). I could add more docs about how to use the module as a whole, if you like. I could also add a function that is more similar to the old `lst_bin()`, which would be a small wrapper around the existing functions. It wouldn't be super-duper easy, though not too hard.

.gitignore
```diff
@@ -17,7 +17,7 @@ docs/_templates
 hera_cal/data/test_output/*
 
 # Test notebooks
-hera_cal/notebooks
+hera_cal/notebooks[
```
jsdillon, line 20: is this intentional?

ci/hera_cal_tests.yml
```diff
@@ -27,8 +27,7 @@ dependencies:
   # - pyuvdata==2.2.12
 
   - pip:
-    - git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata
-    # - git+https://github.com/HERA-Team/aipy
+    - git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata@cal-init-method
```
jsdillon, line 30: I don't love pulling in this PR while the necessary software is still on a branch over at pyuvdata... any chance that'll get pulled in first?

hera_cal/io.py
```diff
@@ -418,7 +429,7 @@ def nan_empty(shape, dtype):
     return rv
 
 
-def get_blt_slices(uvo, tried_to_reorder=False):
+def get_blt_slices(uvo, tried_to_reorder=False, assume_time_first: bool=False, assume_bl_first: bool=False):
```
jsdillon, line 432: These two new arguments should be documented.

hera_cal/io.py
```diff
@@ -430,20 +441,39 @@ def get_blt_slices(uvo, tried_to_reorder=False):
...
+            blt_slices[antp] = slice(i, uvo.Nblts, uvo.Nbls)
+        assert uvo.Nbls == len(blt_slices)
+
+    elif assume_time_first:
+        if not uvo.Nblts != uvo.Nbls * uvo.Ntimes:
+            raise ValueError("Assumption of time first ordering cannot be correct if Nblts != Nbls * Ntimes")
+        for i in range(uvo.Nbls):
+            antp = (uvo.ant_1_array[i*uvo.Ntimes], uvo.ant_2_array[i*uvo.Ntimes])
+            blt_slices[antp] = slice(i, uvo.Nbls + i, 1)
+        assert uvo.Nbls == len(blt_slices)
```
jsdillon, line 460: Not wild about none of this being covered by tests. Do we need it?

hera_cal/io.py
```diff
@@ -1533,7 +1568,7 @@ def write_filter_cache_scratch(filter_cache, cache_dir=None, skip_keys=None):
         warnings.warn("No new keys provided. No cache file written.")
 
 
-def load_flags(flagfile, filetype='h5', return_meta=False):
+def load_flags(flagfile, filetype='h5', return_meta=False, assume_time_first: bool = False, assume_bl_first: bool = False):
```
jsdillon, line 1571: Let's add these kwargs to the docstring

hera_cal/io.py
```diff
@@ -2009,9 +2045,14 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
...
 
-    flags : type=DataContainer, holds data flags, matching data in shape.
+    flags : type=DataContainer or array, holds data flags, matching data in shape. If
+            an array, must be shape (Nbls, Ntimes, Nfreqs, Npols).
 
-    nsamples : type=DataContainer, holds number of points averaged into each bin in data (if applicable).
+    nsamples : type=DataContainer or array, holds number of points averaged into each bin in data 
+               (if applicable). If an array, must be shape (Nbls, Ntimes, Nfreqs, Npols).
+
+    data_array : type=array, if given, use this data as the data_array, instead of
```
jsdillon, line 2054: This feels very hacky. Is it necessary? And if so, why aren't we also doing it with nsamples and flags?

An alternative is a bool (default False) where the user can put in arrays for `data`, `flags`, and `nsamples` and they are treated as such.

hera_cal/io.py
```diff
@@ -2106,21 +2150,38 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
...
         nsample_array = np.ones_like(data_array, float)
     else:
-        nsample_array = np.moveaxis(list(map(lambda p: list(map(lambda ap: nsamples[str(p)][ap], antpairs)), pols)), 0, -1)
+        if isinstance(nsamples, DataContainer):
+            nsample_array = np.zeros((Nbls, Ntimes, Nfreqs, Npols), dtype=float)
+            for i, antpair in enumerate(antpairs):
+                for j, pol in enumerate(pols):
+                    nsample_array[i, :, :, j] = nsamples[antpair + (str(pol),)]
+        else:
+            nsample_array = nsamples
```
jsdillon, line 2171: Why is this being handled differently? And why aren't we checking if it has the right shape (if it's not None)?

hera_cal/io.py
```diff
@@ -2106,21 +2150,38 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
...
         flag_array = np.zeros_like(data_array, float).astype(bool)
     else:
-        flag_array = np.moveaxis(list(map(lambda p: list(map(lambda ap: flags[str(p)][ap].astype(bool), antpairs)), pols)), 0, -1)
+        if isinstance(flags, DataContainer):
+            flag_array = np.zeros((Nbls, Ntimes, Nfreqs, Npols), dtype=bool)
+            for i, antpair in enumerate(antpairs):
+                for j, pol in enumerate(pols):
+                    flag_array[i, :, :, j] = flags[antpair + (str(pol),)]
+        else:
+            flag_array = flags
```
jsdillon, line 2184: Same comment here.

hera_cal/io.py
```diff
@@ -2184,6 +2245,240 @@ def write_vis(fname, data, lst_array, freq_array, antpos, time_array=None, flags
         return uvd
 
 
+def create_uvd_from_hera_data(
```
jsdillon, line 2248: I'm confused why `write_vis` isn't using this function? Seems like they do something very similar, no?

jsdillon: Also, a ton of uncovered lines in this function...

hera_cal/io.py
```diff
@@ -2665,3 +2960,100 @@ def throw_away_flagged_ants_parser():
                     help="Also throw away baselines that have all channels and integrations flagged.")
     ap.add_argument("--clobber", default=False, action="store_true", help='overwrites existing file at outfile')
     return ap
+
+def uvdata_from_fastuvh5(
```
jsdillon, line 2964: Lots of un-covered code in this function... also, feel's like we've got a lot of duplicated functionality between these three functions... Can we condense and re-use code better?

jsdillon (commented):
just finished going over io.py, so I'll leave my comments here as I move on to other modules

jsdillon (commented):
just finished going over io.py, so I'll leave my comments here as I move on to other modules

jsdillon (commented):
just finished going over io.py, so I'll leave my comments here as I move on to other modules

jsdillon (commented):
just finished going over io.py, so I'll leave my comments here as I move on to other modules

setup.py
```diff
@@ -44,20 +44,20 @@ def package_files(package_dir, subdirectory):
...
     'install_requires': [
         'numpy>=1.10',
-        'scipy',
+        'scipy>=1.9.0',
         'h5py',
         'hdf5plugin',
         'astropy',
         'astropy-healpix',
-        'pyuvdata',
+        'pyuvdata @ git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata@cal-init-method',
```
jsdillon, line 56: Again, let's make sure to get rid of these before we merge.

hera_cal/types.py
```diff
@@ -0,0 +1,5 @@
+from typing import Literal
+
+Antpair = tuple[int, int]
+Pol = Literal['xx', 'yy', 'ee', 'nn', 'xy', 'yx', 'en', 'ne']
+Baseline = tuple[int, int, Pol]
```
jsdillon, line 5: This feels like a thing that could go in `utils.py` unless you feel like this is a more intuitive or "standard" place to find these.

hera_cal/utils.py
```diff
@@ -1397,10 +1485,232 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
+        file_list[0], 
+        blts_are_rectangular=blts_are_rectangular, 
+        time_axis_faster_than_bls=time_axis_faster_than_bls
+    )
+    if blts_are_rectangular is None:
+        blts_are_rectangular = meta.blts_are_rectangular
+    if time_axis_faster_than_bls is None:
+        time_axis_faster_than_bls = meta.time_axis_faster_than_bls
+
+    file_list = [
```
jsdillon, line 1560: I find it confusing that this is also called `file_list`... it's really `metadata_list` or something similar.

hera_cal/utils.py
```diff
@@ -1513,3 +1827,10 @@ def list_of_int_tuples(v):
                                                                                 "Ex2: '200 300, 500 650' --> [(200, 300), (500, 650), ...]")
     ap.add_argument("--clobber", default=False, action="store_true", help='overwrites existing file at outfile')
     return ap
+
+def mergedicts(*dicts) -> dict:
```
jsdillon, line 1831: This feels unnecessary when you can just do `{k: v for d in dicts for k, v in d.items()}`

hera_cal/tests/test_io.py
```diff
@@ -896,59 +896,95 @@ def test_read_data(self):
...
-            for ant in hd1.data_antpos:
-                np.testing.assert_array_almost_equal(hd1.antpos[ant] - hd2.antpos[ant], 0)
-            for ap in hd1.times_by_bl:
-                np.testing.assert_array_equal(hd1.times_by_bl[ap], hd2.times_by_bl[ap])
-            for ap in hd1.lsts_by_bl:
-                np.testing.assert_allclose(hd1.lsts_by_bl[ap], hd2.lsts_by_bl[ap])
+        hd1 = io.HERADataFastReader(infile)
+        hd1.read(check=True)
+        hd2 = io.HERAData(infile)
+        #hd2.read()
```
jsdillon, line 971: Did you mean to keep this?

hera_cal/tests/mock_uvdata.py
```diff
@@ -0,0 +1,284 @@
...
+import yaml
+from hera_cal.data import DATA_PATH
+from pathlib import Path
+from hera_cal.red_groups import RedundantGroups
+
+HERA_LOC = EarthLocation.from_geodetic(
+    lat=-30.721526120690307,
+    lon=21.428303826863015,
+    height=1051.690000070259,
+)
```
jsdillon, line 18: Normally we load this from this from pyuvdata's `KNOWN_TELESCOPES` so that we don't need to have duplicate sources of hera's position

hera_cal/tests/mock_uvdata.py
```diff
@@ -0,0 +1,284 @@
...
+
+with open(f"{DATA_PATH}/hera_antpos.yaml", "r") as fl:
+    HERA_ANTPOS = yaml.safe_load(fl)
+
+
+def create_mock_hera_obs(
+    jdint: int = 2459855, 
+    integration_time=10.7,
```
jsdillon, line 26: I'd suggest making the default parameters exactly match real data... for example, this integration time is wrong

jsdillon (changes requested):
Ok, I've finished going through all the changes to ancillary files (i.e. not lstbin.py and lstbin_simple.py) and have a handful of minor notes

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+    dlst: float | None=None, 
+    atol: float=1e-10, 
+    lst_start: float | None=None, 
+    lst_width: float = 2*np.pi,
+    ntimes_per_file: int=60,
+    blts_are_rectangular: bool | None = None,
+    time_axis_faster_than_bls: bool | None = None,
+    jd_regex: str = r"zen\.(\d+\.\d+)\.",
+    lst_branch_cut: float | None=None,
+):
```
jsdillon, line 1626: This needs a docstring.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        nbls_so_far += len(bl_chunk)
+
+    return out_files
+
+
+@profile
+def lst_bin_files(
+    config_file: str | Path,
+    output_file_select: int | Sequence[int] | None=None, 
+    include_autos: bool=True,
```
jsdillon, line 1126: These params are not documented

jsdillon (changes requested):
Got a chance to go through lstbin_simple.py. I've got a handful of questions and clarifications, largely related to documentation.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1677 @@
...
+try:
+    profile
+except NameError:
+    def profile(fnc):
+        return fnc
+
+logger = logging.getLogger(__name__)
+
+@profile
+def simple_lst_bin(
```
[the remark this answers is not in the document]
nkern: that's fine, but can you actually name it something slightly different than `lst_align` b/c a function of the same name appears in `lstbin.py`. anything that isn't exactly identical to the function  in lstbin.py is fine w/ me, thanks

hera_cal/lstbin_simple.py
nkern: cool thanks for the added comments

.gitignore
steven-murray: Nope. Fixed.

ci/hera_cal_tests.yml
steven-murray: That PR is in now, so I've updated this.

hera_cal/io.py
steven-murray: I've removed the arguments because the new pyuvdata can check this internally now, rather than having to "assume" it.

hera_cal/io.py
steven-murray: This speeds up getting the blt slices quite significantly, which is very useful when reading in a lot of files. In saying that, I've been able to clean up the code a little bit here because pyuvdata does some of the heavy lifting now (it'll do even more of it in the future). I'll make sure this gets covered by tests in the meantime.

hera_cal/io.py
steven-murray: Removed them as per comment above.

hera_cal/io.py
steven-murray: I no longer use this in the LST-binner so I've removed these changes. It might be useful one day in the future, but we should be more careful than I was if we want to add it then. I did leave in the change that updates from using a map of a lambda to just doing a for loop.

hera_cal/io.py
steven-murray: Removed.

hera_cal/io.py
steven-murray: Removed.

hera_cal/io.py
steven-murray: Yeah, I'd say that `write_vis` should use this function, to keep things tidy. I'll have a crack at it.

steven-murray: Ah, I don't use this any more either, so I've removed it from this PR. In general it might be a nice thing to have (sometimes you want to get a UVData object without writing a file), but again we can think about that more clearly later.

hera_cal/io.py
steven-murray: This is the function that is now actually used. I'll try make sure I cover it.

setup.py
steven-murray: Done, as above.

hera_cal/utils.py
```diff
@@ -838,40 +846,120 @@ def lst_rephase(data, bls, freqs, dlst, lat=-30.721526120689507, inplace=True, a
...
         # dot bl with difference of pointing vectors to get new u: Zhang, Y. et al. 2018 (Eqn. 22)
-        u = np.einsum("...i,i->...", s_diff, bl)
-
-        # get delay
-        tau = u / const.c.value
-
+        # note that we pre-divided s_diff by c so this is in units of tau.
+        tau = np.einsum("...i,i->...", s_diff, bl)
+        #if i==0:
+        #    print("TAU OLD: ", tau)
```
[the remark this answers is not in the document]
steven-murray: Certainly did. Done.

hera_cal/utils.py
```diff
@@ -838,40 +846,120 @@ def lst_rephase(data, bls, freqs, dlst, lat=-30.721526120689507, inplace=True, a
         data = {'data': data}
         bls = {'data': bls}
 
+    # get new s-hat vector
+    s_prime = np.einsum("...ij,j->...i", rot, np.array([0.0, 0.0, 1.0]))
+    s_diff = (s_prime - np.array([0., 0., 1.0])) / const.c.value
```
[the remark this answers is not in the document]
steven-murray: The point was just to remove an operation from within the loop for the sake of performance. I can rename the variable if you think that's better?

hera_cal/utils.py
```diff
@@ -838,40 +846,120 @@ def lst_rephase(data, bls, freqs, dlst, lat=-30.721526120689507, inplace=True, a
...
         # multiply into data
         data[k] *= phs
-
     if array:
         data = data['data']
 
     if not inplace:
         return data
 
+def lst_rephase_vectorized(
```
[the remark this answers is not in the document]
steven-murray: I think I wanted to be as careful as possible not to break other things. I'll think about trying to make them one function.

hera_cal/utils.py
```diff
@@ -1397,10 +1485,232 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
 
+def match_files_to_lst_bins(
+    lst_edges: tuple[float], 
+    file_list: Sequence[str | Path],
+    files_sorted: bool = False,
+    sort_fn: callable = sorted,
+    jd_regex: str | None = r"zen\.(\d+\.\d+)\.",
+    blts_are_rectangular: bool | None = None,
+    time_axis_faster_than_bls: bool | None = None,
+    atol: float = 1e-5,
```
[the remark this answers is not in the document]
steven-murray: Fixed.

hera_cal/utils.py
```diff
@@ -1397,10 +1485,232 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
+    a regular grid of JDs, but works either way due to the binary search). This latter
+    speedup is really only a speed up if the filename-time matching is not available.
+
+    If the filename-matching is not available, and the LST-edges cover a full 2pi
+    steradians, then this function is as slow (or perhaps a bit slower) than something
+    like ``config_lst_bin_files``. 
+
+    Parameters
+    ----------
+    lst_range
```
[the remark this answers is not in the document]
steven-murray: Noted. Fixed.

hera_cal/utils.py
steven-murray: OK, changed to `metadata_list`.

hera_cal/utils.py
steven-murray: Fair. I've replaced occurrences of the function with this.

hera_cal/tests/test_io.py
steven-murray: Nope. Removed.

hera_cal/tests/mock_uvdata.py
steven-murray: Oh, I didn't realize that existed! I've updated it.

hera_cal/tests/mock_uvdata.py
steven-murray: Good point. I've changed this to 9.66xxx and the frequencies to match the phase II array.

hera_cal/lstbin.py
```diff
@@ -559,39 +604,70 @@ def config_lst_bin_files(data_files, dlst=None, atol=1e-10, lst_start=None, verb
...
         lst_start = all_lsts[np.argmin(all_times)]
     begin_lst = lst_start
 
     # make 24 hour LST grid
-    lst_grid = make_lst_grid(dlst, begin_lst=begin_lst, verbose=verbose)
-    dlst = np.median(np.diff(lst_grid))
+    lst_grid = make_lst_grid(dlst, begin_lst=begin_lst, lst_width=lst_width)
+    dlst = lst_grid[1] - lst_grid[0]
+
+    grid_range = (lst_grid[0] - dlst/2, lst_grid[-1] + dlst/2)
```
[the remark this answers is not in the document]
steven-murray: Fair enough. Changed that.

hera_cal/lstbin.py
```diff
@@ -926,54 +1023,76 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
         garbage_collector.collect()
 
 
-def make_lst_grid(dlst, begin_lst=None, verbose=True):
+def make_lst_grid(
+    dlst: float, 
+    begin_lst: float | None = None, 
+    lst_width: float = 2*np.pi,
```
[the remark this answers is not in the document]
steven-murray: Fixed.

hera_cal/lstbin.py
```diff
@@ -926,54 +1023,76 @@ def lst_bin_files(data_files, input_cals=None, dlst=None, verbose=True, ntimes_p
...
+    else:
+        begin_lst = 0.0
 
+    lst_grid = lst_grid[lst_grid < (begin_lst + lst_width)]
+    
     return lst_grid
 
 
-def sigma_clip(array, sigma=4.0, min_N=4):
+def sigma_clip(array, sigma=4.0, min_N=4, axis: int = 0):
```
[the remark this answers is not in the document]
steven-murray: Fixed.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
+"""
+An attempt at a simpler LST binner that makes more assumptions but runs faster.
+
+In particular, we assume that all baselines with a particular observation have the same 
+time and frequency arrays, and use this to vectorize many of the calculations.
```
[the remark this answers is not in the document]
steven-murray: Yes, this will break with BDA. I think with BDA we should be easily able to put in another loop over BDA groups and maintain the overall vectorization here. I'll make an issue.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+    return lst_bin_centres, _data, _flags, _nsamples
+
+def get_lst_bins(lsts: np.ndarray, edges: np.ndarray) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """Get the LST bin indices for a set of LSTs.
+    
+    Parameters
+    ----------
+    lsts
+        The LSTs to bin, in radians.
+    edges
```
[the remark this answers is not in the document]
steven-murray: I can certainly do that, though I disagree with the premise here: they're called `lst_bin_edges` in other functions because otherwise it would not be clear what the edges were referring to. In this function the context makes it clear what edges they are.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+    lsts %= 2*np.pi
+    lsts[lsts < edges[0]] += 2* np.pi
+    bins = np.digitize(lsts, edges, right=True) - 1
+    mask = (bins >= 0) & (bins < (len(edges)-1))
+    return bins, lsts, mask
+
+def reduce_lst_bins(
+    data: list[np.ndarray], flags: list[np.ndarray], nsamples: list[np.ndarray],
+    mutable: bool = False,
+    sigma_clip_thresh: float = 0.0,
```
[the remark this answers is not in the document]
steven-murray: I agree, but using None means it's a little more tricky to deal with in the CLI (which maybe isn't a great argument for not doing it). OK, I've convinced myself that I'll go fix it.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+    ----------
+    data
+        The data to compute the statistics over. Shape ``(nnights, nbl, nfreq, npol)``.
+    nsamples
+        The number of samples for each measurement. Same shape as ``data``.
+    flags
+        The flags for each measurement. Same shape as ``data``.
+    flag_thresh
+        The fraction of times a baseline/frequency/pol must be flagged in order to
+        entirely flag the baseline/frequency/pol.
```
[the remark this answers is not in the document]
steven-murray: I also hesitated over this. One thing I haven't quite landed on yet that I don't quite agree with you on is using "nights" here, because it's possible that later we use two time-integrations per night in each LST bin, then this doesn't correspond to nights, but maybe "integrations"?

So how about "The fraction of integrations within an LST bin for a given baseline (including pol) and channel that must be flagged in order for this function to flag that baseline-channel completely"?

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="Degrees of freedom <= 0 for slice.")
+        print('MEAN: ', meandata)
+        print("norm: ", norm)
+        std = np.square(data.real - meandata.real) + 1j*np.square(data.imag - meandata.imag)
+        std = np.nansum(std * nsamples, axis=0)
+        print("SUM OF SQUARES: ", std)
+        std[~f_min] /= norm[~f_min]
+        print("MEAN OF SQUARES: ", std)
+        std = np.sqrt(std.real) + 1j*np.sqrt(std.imag)
```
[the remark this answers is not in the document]
steven-murray: absolutely not. fixed.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        A list of antenna pairs to read from each file. Each pair should be a tuple
+        of antenna numbers. Note that having pairs in this list that are not present
+        in a particular file will not cause an error -- that file will simply not
+        contribute for that antpair.
+    freqs
+        Frequencies contained in the files. If not provided, will be read from the
+        first file in ``data_files``.
+    pols
+        Polarizations to read. If not provided, will be read from the first file in
+        ``data_files``.
```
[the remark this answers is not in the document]
steven-murray: Yes, because we anyway have to write out a full UVData file, that has a data shape that is rectangular in (antpairs, pols). If we pull out specific 3-tuple baselines here, then we have to do work to spoof values that are in the full cross-product of antpairs-pols but don't exist in our 3-tuple baselines etc, and it can't be vectorized as well.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        A dictionary mapping antenna numbers to antenna positions. Only required
+        if ``rephase`` is True. If not provided, and required, will be determined
+        by reading as many of the files as required to obtain all antenna positions
+        in antpairs.
+    lsts
+        A list of LST arrays for each file. If not provided, will be read from the
+        files. If provided, must be the same length as ``data_files``.
+    freq_min, freq_max
+        Minimum and maximum frequencies to include in the data. If not provided,
+        all frequencies will be included.
```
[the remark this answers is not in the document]
steven-murray: So, at an earlier stage (before IDR2) I was running a pipeline that had a step before LST-binning that wrote out my particular "SHARP" data selection as files with reduced frequency range (to increase the speed of the computation). In early stages of IDR2 processing, I wanted the ability to also output LST-binned data with that reduced frequency range to do comparisons with my older runs, but I didn't want a whole new pipeline step that wrote out all the data again with that reduced frequency range -- I wanted it to happen within the LST-binner itself. So I added this to support down-selecting frequencies on the fly. 

I don't use it now, but I'm wondering if having something like this wouldn't be useful in the long run?

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        else:
+            out_data[:, lstbin] = 1.0
+            out_flags[:, lstbin] = True
+            out_std[:, lstbin] = 1.0
+            out_nsamples[:, lstbin] = 0.0
+
+        
+    return out_data, out_flags, out_std, out_nsamples
+
+def _allocate_dnf(shape: tuple[int], d=0.0, f=0, n=0):
```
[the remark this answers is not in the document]
steven-murray: Yeah, good point. Fixed.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        if not bls_to_load or not np.any(tind):
+            # If none of the requested baselines are in this file, then just 
+            # set stuff as nan and go to next file. 
+            logger.warning(f"None of the baseline-times are in {meta.path}. Skipping.")
+            data[slc] = np.nan
+            flags[slc] = True
+            nsamples[slc] = 0
+            continue
+
+        # TODO: use Fast readers here instead.
```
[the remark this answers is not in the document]
steven-murray: So the catch is that the `HERADataFastReader` has no ability to down-select on times, which I need. I had tried to add this functionality in this branch, but it was not working quite right somehow, so I gave up. There will be a more robust solution coming soon (given pending PRs on pyuvdata) so I decided to wait on those. This will definitely be something I'll come back to.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+    bins = get_lst_bins(lsts, lst_bin_edges)[0]
+    times = np.concatenate(time_arrays)
+    times_in_bins = []
+    for i in range(len(bin_lst)):
+        mask = bins == i
+        times_in_bins.append(times[mask])
+
+    return bin_lst, data, flags, nsamples, times_in_bins
+
+def apply_calfile_rules(
```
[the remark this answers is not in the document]
steven-murray: Yup. Added.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        lst_grid += begin_lst
+    else:
+        begin_lst = 0.0
+
+    lst_grid = lst_grid[lst_grid < (begin_lst + lst_width)]
+    
+    return lst_grid
+
+
+def lst_bin_files_single_outfile(
```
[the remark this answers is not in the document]
steven-murray: Done.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+
+    return out_files
+
+
+@profile
+def lst_bin_files(
+    config_file: str | Path,
+    output_file_select: int | Sequence[int] | None=None, 
+    include_autos: bool=True,
+    **kwargs
```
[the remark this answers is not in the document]
steven-murray: I agree, though this is more an issue with `hera_cal`'s lack of sphinx-documentation, I think. I equally don't love copy-pasting massive chunks of docstrings, because they go stale real quick. If we were making sphinx docs that we could look at, it could be dynamically all put together. For now, I've left the kwargs in because I don't want to set the precedent of copy-pasting massive parameter lists. I could probably get some sphinx docs together sometime in the near future.

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        calfile is missing.
+    save_channels
+        A list of channels for which to save the a full file of LST-gridded data. 
+        One REDUCEDCHAN file is saved for each output file, corresponding to the
+        first LST-bin in that file. The data in that file will have the shape 
+        ``(Nbls*Ndays, Nsave_chans, Npols)``. This can be helpful for debugging.
+    golden_lsts
+        A list of LSTs for which to save the a full file of LST-gridded data.
+        One GOLDEN file is saved for each ``golden_lst``, with shape ``(Nbls*Ndays, Nfreqs, Npols)``.
+        This can be helpful for debugging.
```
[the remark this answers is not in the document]
steven-murray: I updated the docstring here to the following:

```
        A list of LSTs for which to save a full file of LST-aligned (but not 
        averaged) data. One GOLDEN file is saved for each ``golden_lst``, with shape 
        ``(Nbls*Ndays, Nfreqs, Npols)`` -- that is, the normal "time" axis of a 
        UVData array is replaced by a "night" axis. This is an easy way to load up
        the full data that goes into a particular LST-bin after the fact.
```

hera_cal/lstbin_simple.py
```diff
@@ -0,0 +1,1740 @@
...
+        timefirst = bool(f['Header']['time_axis_faster_than_bls'][()])
+        if not timefirst and ntimes > 1:
+            raise NotImplementedError("Can only do time-first files for now.")
+        
+        slc = slice(slc.start*ntimes, slc.stop*ntimes, 1)
+        f['Data']['visdata'][slc] = data.reshape((-1, data.shape[2], data.shape[3]))
+        f['Data']['flags'][slc] = flags.reshape((-1, data.shape[2], data.shape[3]))
+        f['Data']['nsamples'][slc] = nsamples.reshape((-1, data.shape[2], data.shape[3]))
+
+def config_lst_bin_files(
```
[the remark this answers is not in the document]
steven-murray: Ah... yes. I've removed the import, because I did update this function a little bit.

hera_cal/lstbin_simple.py
steven-murray: Done.

hera_cal/lstbin_simple.py
steven-murray: Done.

hera_cal/utils.py
```diff
@@ -1397,10 +1485,232 @@ def red_average(data, reds=None, bl_tol=1.0, inplace=False,
...
+        function that takes a list of str or Path objects and returns a list of the same
+        in order.
+    jd_regex
+        The regex to use to extract the julian date from the file name. If provided,
+    approximate
+        If True, return a list of files that is a superset of those that exist in the
+        LST bins -- either the exact list or more than the true list. This will be based
+        on times inferred from the filename (if possible) and the integration time of
+        the first file. 
+
```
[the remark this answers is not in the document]
steven-murray: Fixed, thanks!

hera_cal/utils.py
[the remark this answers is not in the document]
jsdillon: Yeah, I'd prefer that. I'd even be fine with something like `s_diff_over_c`

hera_cal/utils.py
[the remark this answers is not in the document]
jsdillon: What was the verdict on this?

hera_cal/lstbin_simple.py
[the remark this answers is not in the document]
jsdillon: That works for me.

hera_cal/lstbin_simple.py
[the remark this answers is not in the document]
jsdillon: Your call. I could see it being useful interactively, but I don't think we want to run in this mode most of the time.

hera_cal/utils.py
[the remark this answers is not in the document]
steven-murray: Cool. Done.

hera_cal/utils.py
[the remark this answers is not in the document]
steven-murray: I made it one function, though that one function is a reasonable amount more complicated than the original. Not quite as large as the two separate functions combined though.

jsdillon (approved):
I think I'm basically satisfied. I can look at any minor changes you push from here, but I think we should merge in (even if we lose a tiny sliver of coverage fraction) and then work out any issues through battle testing.

PR #847, opened 2022-12-09
Make amplitude calibration optional in `post_redcal_abscal`

jsdillon:
This PR makes it possible to skip amplitude calibration entirely in `post_redcal_abscal`. This is useful for doing amplitude calibration with just autos completely separately from phase calibration with crosses.


zacharymartinot (approved):
looks fine

PR #850, opened 2022-12-17
Improve memory usage in RFI_delay_slope_cal

jsdillon:
I noticed when running notebooks at NRAO that I was getting memory errors I didn't expect. Doing an mprof, I found a spike which I traced back to `RFI_delay_slope_cal`

image

This PR takes part of of the calculation of the per-baseline delay in RFI cal out of np.einsum and puts it into a loop to improve memory usage. It's about 15% slower (on my laptop, it's 2.85 seconds vs. 2.48 seconds), but a factor of Nrfi_chans more memory efficient, which seems like a great trade-off.


jsdillon:
Here's the after:

image

steven-murray (approved):
This looks right to me, and seems like a reasonable trade-off. The `rfi_chans` axis is not assumed to be a long axis ever, which is the only case in which I would think the time performance would be seriously affected.

PR #852, opened 2022-12-28
perf: replace all map(lambda) with list comp

steven-murray:
This replaces all instances of `map(lambda ....)` with list comprehension. This is almost always easier to read, and I've found it to be sometimes MUCH more efficient (though I haven't figured out exactly in which cases it is more efficient).


jsdillon (approved):
Found and fixed a bug, but otherwise good to go! Thanks @steven-murray !

PR #853, opened 2022-12-28
Remove deprecated numpy types

jsdillon:
This PR removes types like np.int and np.float and np.str that were deprecated in numpy 1.24: https://numpy.org/devdocs/release/1.24.0-notes.html

This PR relies on https://github.com/HERA-Team/hera_filters/pull/7

This PR also fixes the pyuvdata version to 2.12.2 until we can fix the phasing issue (see #851) and moves to using the newest version of aipy on github, since the version on pypi is out of date

### DIFF
--- a/ci/hera_cal_tests.yml
+++ b/ci/hera_cal_tests.yml
@@ -8,7 +8,7 @@ dependencies:
   - astropy
   - h5py
   - hdf5plugin
-  - numpy>=1.24.1
+  - numpy>=1.24
   - pytest
   - scipy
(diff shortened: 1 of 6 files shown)


steven-murray (commented):
Do you need to also update `setup.py` to use correct versions? The problem with that is that you can't use github links in setup.py if you wanna upload to pypi...

steven-murray (approved):
This looks good -- we just need to remember to relax the versions when we can.

PR #854, opened 2022-12-29
Faster slice func

steven-murray:
This caches the result of `polstr2num` for a particular HERAData object because I was finding that when reading a single file from H6C, when getting the slices for each baseline, it took about 12 seconds, or ~20% of the total read time. When LST binning, this can be an even higher percentage because you might only read one out of the two times -- decreasing actual read time but not decreasing the number of baselines that need to be indexed.

This PR makes the indexing time essentially negligible.

FWIW, on a single-shot read (the above was done reading 350 baselines at a time, similar to what would be done in a real pipeline), about 1/3 of the read time is taken in doing copies of the data into the datacontainers. I wonder if it would be useful to have a non-copying option.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -583,9 +583,15 @@ class HERAData(UVData):
             polnum: i for i, polnum in enumerate(self.polarization_array)
         }
         pols = [polnum2str(polnum, x_orientation=self.x_orientation) for polnum in self.polarization_array]
-        self._polstr_indices = {
-            pol: self._polnum_indices[polstr2num(pol, x_orientation=self.x_orientation)] for pol in pols
-        }
+        self._polstr_indices = {}
+        # Add upper-case indices as well, so we don't need to use .lower() on input
+        # keys (for which there can be many tens of thousands).
+        for pol in pols:
+            indx = self._polnum_indices[polstr2num(pol, x_orientation=self.x_orientation)]
+            self._polstr_indices[pol.lower()] = indx
+            self._polstr_indices[pol.upper()] = indx
+            self._polstr_indices[pol[0].upper() + pol[1].lower()] = indx
+            self._polstr_indices[pol[0].lower() + pol[1].upper()] = indx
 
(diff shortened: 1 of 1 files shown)


jsdillon:
This looks quite reasonable (though I'm surprised by the level of speed-up you're seeing), but it's failing tests...

hera_cal/io.py
```diff
@@ -579,9 +579,19 @@ def _determine_blt_slicing(self):
...
+        }
+        pols = [polnum2str(polnum, x_orientation=self.x_orientation) for polnum in self.polarization_array]
+        self._polstr_indices = {}
+        # Add upper-case indices as well, so we don't need to use .lower() on input
+        # keys (for which there can be many tens of thousands).
+        for pol in pols:
+            indx = self._polnum_indices[polstr2num(pol, x_orientation=self.x_orientation)]
+            self._polstr_indices[pol.lower()] = indx
+            self._polstr_indices[pol.upper()] = indx
+            self._polstr_indices[pol[0].upper() + pol[1].lower()] = indx
```
jsdillon, line 593: If for some weird reason the pol is a single char, this would break. Not sure if it broke before, so perhaps worth checking?

Instead of handling edge cases like this, what if we used dictionary default behavior? For example:

`self._polstr_indices[key[2], self._polnum_indices[polstr2num(key[2], x_orientation=self.x_orientation)]`

jsdillon (changes requested):

hera_cal/io.py
steven-murray: Yeah, good idea.

steven-murray: I left the overall lower/upper, but not the weird cases where one one letter is upper and the other lower.

jsdillon (approved):
Looks great!

I personally find these extremely spread-out code bits that `black` produces to be hard to read, especially since we're not at all strict about line lengths in hera_cal. But since I also take a somewhat laissez faire attitude toward these things, I don't think I have a principled objection to the formatting.

PR #856, opened 2023-01-04
Update frf.py to accomodate updated UVBeam power beam shapes.

aewallwi:
UVBeam seems to now require dummy index for e-field beam in power beams. Here, we update some frf code to accomodate this.


aewallwi:
This is actually breaking a fix to address pyuvdata interface updates.

PR #857, opened 2023-01-06
perf: don't discard imag autos Nbls times

steven-murray:
Un-indents the discard autocorr imag calls so they're not run for every baseline.

Interesting fact: I found that these three lines were taking >75% of the total time for doing the delay filter on H6C data.


jsdillon (approved):
Looks good to me.

PR #860, opened 2023-01-12
perf: faster partial-writing by not deepcopying

steven-murray:
Makes the `partial_write` method a bit faster by not deepcopying the whole datacontainer object, but instead just converting the data/flags/nsamples to arrays before writing out. 

These deepcopy's are taking ~50% of the time for a delay filter that is done in chunks of baselines, so it is quite significant.


steven-murray:
@jsdillon there's a "chunking" step before the delay filter that puts 8 2-integration files together. It's not clear to me whether this is a good or bad idea, but I inherited it from the H4C workflow. Maybe @aewallwi can comment. With this much data, reading all baselines does not fit in the 32GB mem I allocate.

jsdillon:
Sounds like an order of operations that could be better optimized...

hera_cal/io.py
```diff
@@ -883,6 +883,30 @@ def update(self, data=None, flags=None, nsamples=None, tSlice=None, fSlice=None)
...
+        if nsamples is not None:
+            self.datacontainer_to_data_array(nsamples, self.nsample_array, tSlice=tSlice, fSlice=fSlice)
+        
+    def datacontainer_to_data_array(
+        self, dc: DataContainer, data_array = None, tSlice=None, fSlice=None
+    ) -> np.ndarray:
+        '''Convert a datacontainer to an array with uvdata format.
+
+        Note that if ``data_array`` is not provided, and tSlice or fSlice are, then
+        the elements outside the slices will be returned as zeros.
```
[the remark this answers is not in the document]
steven-murray: Good catch, fixed that up.

steven-murray:
@jsdillon this is now passing tests except for a weird linsolve issue that I'm sure doesn't come from my changes...?

hera_cal/datacontainer.py
```diff
@@ -67,6 +67,16 @@ def __init__(self, data):
                     setattr(self, attr, getattr(data, attr))
                 else:
                     setattr(self, attr, None)
+            
+            # Define the dtype of the data itself. next(iter()) gets the first element
+            # of keys() without creating a whole list.
+            if self.keys():
+                try:
+                    self.dtype = self._data[next(iter(self.keys()))].dtype
```
jsdillon, line 75: I think this probably fine most of the time, though I wonder whether it makes more sense to to make `.dtype` an `@property`? As it stands, one can init an empty datacontainer with `{}`, and then end up with `self.dtype` as `None`.

jsdillon (changes requested):

hera_cal/datacontainer.py
steven-murray: Ah yeah, I see your point... you could then do a `read` or whatever and the `dtype` wouldn't be updated. I'll make the change.

hera_cal/io.py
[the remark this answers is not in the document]
jsdillon: I don't think you actually addressed this... though perhaps this raises the question of why `data_array` is an optional parameter and not a required one. If it's required, that solves this issue. And if it's required, perhaps you want to call this function something like `set_data_array_with_datacontainer()`

hera_cal/io.py
```diff
@@ -900,18 +924,19 @@ def _set_subslice(data_array, bl, this_waterfall):
...
-            for bl in data.keys():
-                _set_subslice(self.data_array, bl, data[bl])
-        if flags is not None:
-            for bl in flags.keys():
-                _set_subslice(self.flag_array, bl, flags[bl])
-        if nsamples is not None:
-            for bl in nsamples.keys():
-                _set_subslice(self.nsample_array, bl, nsamples[bl])
+        if data_array is None:
+            data_array = np.zeros_like(self.data_array)
```
jsdillon: Missing coverage

jsdillon (changes requested):
I don't think my issue about default values was addressed, though I think the whole issue can be avoided by requiring a data_array in datacontainer_to_data_array

jsdillon (approved):

PR #861, opened 2023-01-12
perf: logging outputs current memory usage, and auto-profile scripts

steven-murray:
This adds some logging config that makes it easy (in fact, default) to output the current memory usage as a column in the log. This is obviously helpful for profiling. Furthermore, it adds a wrapper function that can turn on line-profiling for any given modules/functions. It then defines some argument parser groups to configure these settings for scripts, and adds these to all current scripts that use imported argparser objects.


steven-murray:
To explain the workings of this PR more:

It adds one new module, `_cli_utils`, which has some helper functions/classes for adding functionality to scripts. The first main bit of functionality is a new `Logger`, which is based on the `rich.RichHandler` (see here: https://rich.readthedocs.io/en/stable/logging.html). This makes logging output much prettier by default (eg. red text for ERROR, orange for WARNING, etc).

Most importantly, I've implemented a subclass of this handler which also adds a column for current memory usage, which acts as a cheap option for memory profiling (not as good as using `mprof` and saving the results to a file, but gives a good enough picture on a general run). It also adds the option of reporting the time as a diff on the start time of the script, instead of the full absolute date/time. The way that logging works in python, this is all implemented as opt-in, so all the logging statements throughout the code are done exactly the same way, and this module has to be loaded and "setup" to actually use this handler. It makes most sense to use the handler when running a script, which is why this appears in `_cli_utils`. Furthermore, there is a function that adds a new parameter group to an `ArgumentParser` that can control the logging (all the new arguments begin with `--log-xxx`). 

The second thing added here is a function that makes it easy to run the line-profiler. Generally, when you run `line_profiler`, you need to run `line_profiler python script.py` on the CLI, and additionally, you need to add `@profile` decorators to all the functions you want to be profiled. This can be super annoying, because you have to modify the code itself, then make a commit, push, install the code on NRAO, run the code through the line profiler, look at the results, then go back and remove the `@profile` decorators. 

However, `line_profiler` gives a programmatic way to run it, and this addition uses that functionality. By running the `run_with_profiling` function, given an overall function to run, it will add that top-level function to the list of functions it will report on. It also adds an argparse option `--profile-funcs` which takes in a comma-separated list of importable functions to also add to that list. In the event that `--profile` is not on the command-line options, then `run_with_profiling()` is essentially a no-op and the original function just runs. In total, this allows really easy specification of running line-profiling on select functions/classes on any of the scripts in hera_cal.

hera_cal/_cli_tools.py
```diff
@@ -0,0 +1,351 @@
+"""Useful helper functions and argparsers for scripts."""
```
aewallwi, line 1: If this is boilerplate code, put in a reference to it's origin / some explanation. If not, then I think it'd be helpful to document all of these functions.

aewallwi (changes requested):
Mostly good. My main request is that we should document the cli tools code or point to online documentation/tutorials.

aewallwi:
@steven-murray whenever you think things are ready for another look, just let me know!

steven-murray:
@aewallwi I think it's ready to be looked at again now! I added significanta mounts of documentation in the module itself. I can add a separate tutorial if you like.

hera_cal/_cli_tools.py
```diff
@@ -0,0 +1,582 @@
...
+    rich_tracebacks: bool=True,
+    show_mem: bool = True,
+    mem_backend: Literal["tracemalloc", "psutil"] = "tracemalloc",
+    show_path: bool = False,
+):
+    """Setup the logger for hera_cal scripts.
+    
+    Parameters
+    ----------
+    level : str, optional
```
aewallwi, line 147: What are the different options for this?

aewallwi (approved):
Thanks for all the hard work @steven-murray. I think these logging tools will be especially useful for identifying pain points made clear by the explosion in data volume in H6C. One minor comment but I've approved.

PR #863, opened 2023-01-18
Delay filter speedup

steven-murray:
This PR is a culmination of effort to speed up the delay filter (which I started in #857). In fact, in the previous PR, the whole job was not done because there is an outer loop over keys in another function, and in H4C we're actually doing the loop over keys there, not in the `fourier_filter` function itself. This PR fixes that, and also the excessive calling of `polstr2num` and its variants, which was supposed to have been fixed in #854 but inadvertantly got broken by using the `dict.get(key, default)` notation, which *evaluates* the default regardless of whether the key exists in the dict. In this PR, we fix it by simply using the `lru_cache` on the `polnum2str` function. I actually made a PR on pyuvdata (https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1250) which would mean we can omit this decorator in hera_cal, if approved.

This branch is based off the logging PR #861, so that should be merged before even reviewing this one (it'll make the review much easier...)


aewallwi (changes requested):
Looks like this PR is failing tests because of a typo in the install script. Can you fix this @steven-murray? Happy to look this over if it's ready.

steven-murray:
@aewallwi / @jsdillon do you know what the error is for Python >= 3.9? I've gotten it a few times before, but I don't think it arises from anything that I've changed?

aewallwi:
> @aewallwi / @jsdillon do you know what the error is for Python >= 3.9? I've gotten it a few times before, but I don't think it arises from anything that I've changed?

Not sure where this comes from. I think we should make an issue for it but I don't see a reason to let it hold up the fixes in these PRs.

aewallwi (approved):
This looks good to me, I approve!

PR #864, opened 2023-01-19
Temporarily restore old array shape compatibility in HERAData for the sake of pspec

steven-murray:
This fixes compatibility with `hera_pspec`, which does not yet have future array shape support. I opened https://github.com/HERA-Team/hera_pspec/issues/373 to make this happen, so we can revert to future array shapes after that.


adeliegorce:
Hi, I don't think I can review because I don't know `hera_cal` but just for info: I do think it might be a `numpy` version problem because now calling `np.float` throws an error when before is was only a `DeprecationWarning`

adeliegorce (dismissed):

jsdillon (approved):
Seems fine to me. @steven-murray can you make an issue (and assign yourself) to undo this PR when hera_pspec is ready?

PR #868, opened 2023-01-30
Fix some errors with coming pyuvdata changes.

bhazelton:
External hera_cal tests are failing on a pyuvdata PR where we are starting to always have the `flex_spw_id_array` on UVData, UVCal and UVFlag objects. It's not yet required, but that PR is initializing it when we create objects from files.

This PR fixes some bugs in selection operations in hera_cal that are not currently handling that parameter it if it is set, resulting in it being the wrong size after the operation.


bhazelton:
@jsdillon One test is failing in one of the CI jobs, but it doesn't seem to be related to the changes I made in this PR, so I'm not sure what is going on there.

jsdillon (approved):

PR #869, opened 2023-01-31
feat: add partial time-reading and conjugation to HERADataFastReader

steven-murray:
Adds:
  - `keep_times` argument to `read_hera_hdf5`, which keeps certain times at the end of the read (not really partial I/O, but still better than `.select()` on a DC).
  - ability to get conjugate baselines in `read_hera_hdf5`.
  - more tests to ensure partial bls/pols/times is done right.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -1109,19 +1109,23 @@ class HERAData(UVData):
         self.nsample_array = (np.zeros_like(self.nsample_array) if self.nsample_array is not None else None)
 
 
-def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
-                   read_data=True, read_flags=False, read_nsamples=False,
-                   check=False, dtype=np.complex128, verbose=False):
+def read_hera_hdf5(
+    filenames, bls=None, pols=None, keep_times: list[float] | None=None, full_read_thresh=0.002,
+    read_data=True, read_flags=False, read_nsamples=False,
+    check=False, dtype=np.complex128, verbose=False
+):
     '''A potentially faster interface for reading HERA HDF5 files. Only concatenates
     along time axis. Puts times in ascending order, but does not check that
     files are contiguous. Currently not BDA compatible.
-
     Arguments:
         filenames: list of files to read
         bls: list of (ant_1, ant_2, [polstr]) tuples to read out of files.
              Default: all bls common to all files.
         pols: list of pol strings to read out of files. Default: all, but is
               superceded by any polstrs listed in bls.
+        keep_times: list of times to read out of files. Default: all. Note: all times 
+              are always read from the files, setting this only down-filters times
+              after reading.
         full_read_thresh (0.002): fractional threshold for reading whole file
                                   instead of baseline by baseline.
         read_data (bool, True): read data
@@ -1130,7 +1134,6 @@ def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
         check (bool, False): run sanity checks to make sure files match.
         dtype (np.complex128): numpy datatype for output complex-valued arrays
         verbose: print some progress messages.
-
     Returns:
         rv: dict with keys 'info' and optionally 'data', 'flags', and 'nsamples',
             based on whether read_data, read_flags, and read_nsamples are true.
@@ -1148,6 +1151,8 @@ def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
     inds = {}
     # Read file metadata to size up arrays and sort times
     filenames = _parse_input_files(filenames, name='input_data')
+    
+    
     for filename in filenames:
         if verbose:
             print(f'Reading header of {filename}')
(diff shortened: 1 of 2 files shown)


hera_cal/io.py
```diff
@@ -1109,19 +1110,23 @@ def empty_arrays(self):
         self.nsample_array = (np.zeros_like(self.nsample_array) if self.nsample_array is not None else None)
 
 
-def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
-                   read_data=True, read_flags=False, read_nsamples=False,
-                   check=False, dtype=np.complex128, verbose=False):
+def read_hera_hdf5(
+    filenames, bls=None, pols=None, keep_times: list[float] | None=None, full_read_thresh=0.002,
```
jsdillon, line 1114: I'm not familiar with this style of type hinting. Why is it`| None=None`?

hera_cal/io.py
```diff
@@ -1109,19 +1110,23 @@ def empty_arrays(self):
...
     along time axis. Puts times in ascending order, but does not check that
     files are contiguous. Currently not BDA compatible.
-
     Arguments:
         filenames: list of files to read
         bls: list of (ant_1, ant_2, [polstr]) tuples to read out of files.
              Default: all bls common to all files.
         pols: list of pol strings to read out of files. Default: all, but is
               superceded by any polstrs listed in bls.
+        keep_times: list of times to read out of files. Default: all. Note: all times
```
jsdillon, line 1127: Do we want to implement some kind of tolerance here? I know pyuvdata has a tolerance for figuring out of times match, perhaps we can use their default under the hood?

hera_cal/io.py
```diff
@@ -1148,6 +1152,12 @@ def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
     inds = {}
     # Read file metadata to size up arrays and sort times
     filenames = _parse_input_files(filenames, name='input_data')
+    
+    if bls is not None:
+        antpairs = [bl[:2] for bl in bls]
+    else:
+        antpairs = None
```
jsdillon, line 1159: Why are we ignoring polarization? If I only want to load one bl for one pol, shouldn't that be allowed?

hera_cal/io.py
```diff
@@ -1366,22 +1415,25 @@ def _adapt_metadata(self, info_dict, skip_lsts=False):
         info_dict['data_antpos'] = {ant: info_dict['antpos'][ant] for ant in info_dict['data_ants']}
         info_dict['times'] = np.unique(info_dict['times'])
         info_dict['times_by_bl'] = {ap: info_dict['times'] for ap in info_dict['antpairs']}
-        info_dict['times_by_bl'].update({(a2, a1): info_dict['times'] for (a1, a2) in info_dict['antpairs']})
+        # info_dict['times_by_bl'].update({(a2, a1): info_dict['times'] for (a1, a2) in info_dict['antpairs']})
```
jsdillon, line 1418: why was this removed?

hera_cal/io.py
```diff
@@ -1233,8 +1258,13 @@ def read_hera_hdf5(filenames, bls=None, pols=None, full_read_thresh=0.002,
                 pols = list(pol_indices.keys())
             bls = set(bl for bl in bls if len(bl) == 3)
             bls = bls.union([bl + (p,) for bl in bls_len2 for p in pols])
+        
+        # now, conjugate them if they're not in the data
+        bls = {(i, j, p) if (i, j) in info['bls'] else (j, i, utils.conj_pol(p)) for (i, j, p) in bls}
```
jsdillon, line 1263: This would be slightly cleaner with utils.reverse_bl()

hera_cal/io.py
```diff
@@ -1306,23 +1336,40 @@ def index_exp(i, j, p):
                 else:
                     for i, j, p in bls:
                         data[i, j, p][t:t + ntimes] = d[index_exp(i, j, p)]
-
             t += ntimes
+    
+    # Now subselect the times. 
+    if keep_times is not None:
+        time_mask = np.array([t in keep_times for t in info['times']])
```
jsdillon, line 1343: See comment above about tolerances.

jsdillon (changes requested):
Nice work! Just a handful of questions/comments.

hera_cal/io.py
steven-murray: The following None is the default for the parameter. I agree it could look a litle silly, though I'm not sure how it would look best.

hera_cal/io.py
steven-murray: yeah, I think that should be doable. I'll add it in.

hera_cal/io.py
steven-murray: Done.

hera_cal/io.py
steven-murray: Great, thanks!

hera_cal/io.py
steven-murray: Yeah, you can do that. This is just getting the antpairs for when we check whether all the antpairs are present in the data.

hera_cal/io.py
steven-murray: It wasn't there originally -- I added it when I thought I needed to manually index all the conjugated bls. I'll remove it completely.

PR #870, opened 2023-02-05
Speed up sun altitude calculation for large arrays of JDs

jsdillon:
For a full day of H6C data, I find this speeds things up by a factor of 10-20.


steven-murray (approved):
Oh yeah, I heard that this was definitely a thing you should do.

PR #873, opened 2023-02-21
ci: faster CI hopefully

steven-murray:
This changes the miniconda env to a simple pip setup which is generally much faster. It also adds `--durations` to the pytest call, so we can more easily optimize the tests over time.


jsdillon (approved):
Nice work @steven-murray!

PR #874, opened 2023-02-23
Multi purpose subselect

steven-murray:
Adds a short script to amke it easy to subselect uvdata files on different axes. Obviously, this is easy already from an interpreter, but this means we can do it more easily from a makeflow.

### DIFF
--- a/scripts/subselect.py
+++ b/scripts/subselect.py
@@ -79,10 +79,8 @@ def select(
 ):
     if not os.path.exists(infile):
         raise FileNotFoundError(f"File {ap} does not exist.")
-    if infile != outfile and os.path.exists(outfile) and not clobber:
+    if os.path.exists(outfile) and not clobber:
         raise FileExistsError(f"File {outfile} exists and clobber is False.")
-    if infile == outfile:
-        clobber= True
 
     hd = UVData()
     logger.info(f"Reading metadata from file {infile}")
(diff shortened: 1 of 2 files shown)


scripts/subselect.py
```diff
@@ -0,0 +1,229 @@
...
+    # Get times
+    if time_min is not None or time_max is not None or time_idxs is not None:
+        logger.info("Getting times to read.")
+        times = np.unique(hd.time_array)
+        time_bools = np.ones_like(times, dtype=bool)
+
+        if time_idxs is not None:
+            time_idxs = [tuple(map(int, idx.split("~"))) for idx in time_idxs.split(",")]
+            for idx in time_idxs:
+                time_bools[idx[0]:idx[1]] = False
```
mkolopanis, line 115: We want to keep times between idx[0] and idx[1] right? wouldn't setting this to False deselect them?

It might be easier (to wrap a reader's head around) if the boolean array defaults to False, then selected times are set to True. It would also match the logic style from the freqs block above then

scripts/subselect.py
```diff
@@ -0,0 +1,229 @@
...
+    min_bl_length=None, max_bl_length=None, min_ew_length=None, max_ew_length=None,
+    clobber=False,  check=False, check_acceptability=False, check_uvw_strict=False,
+    check_autos=False, fix_autos=False
+):
+    if not os.path.exists(infile):
+        raise FileNotFoundError(f"File {ap} does not exist.")
+    if infile != outfile and os.path.exists(outfile) and not clobber:
+        raise FileExistsError(f"File {outfile} exists and clobber is False.")
+    if infile == outfile:
+        clobber= True
```
mkolopanis, line 85: This feels a little dangerous. If a user chooses the same file for in and out, I get they _probably_ want it to be clobbered, but you never know. I could just imagine something terrible happening.

mkolopanis (commented):
This looks convenient. I have a few small points.

mkolopanis (approved):
Looks good from my point of view

PR #875, opened 2023-03-01
Enable antenna and waterfall type UVFlag files in smooth_cal

jsdillon:
This PR enables the use of antenna-type and waterfall-type UVFlag files in smooth_cal, instead of just baseline-type. These are now allowed to overlap in time, so for example, one can impose a set of binary antenna flags and a single RFI waterfall simultaneously.

It also adds an option ignore flags from the calfits file, allowing the user to just use the UVFlag files.

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -672,7 +672,7 @@ class CalibrationSmoother():
         Arguments:
             calfits_list: list of string paths to calfits files containing calibration solutions and flags
             flag_file_list: list of string paths to files containing flags as a function of time, frequency, and
-                either baseline, antnena, or just antenna polarization. Must have all baselines/ants/pols for all times.
+                either baseline, antenna, or just antenna polarization. Must have all baselines/ants/pols for all times.
                 Flags on baselines are broadcast to both antennas involved, unless either antenna is completely flagged
                 for all times and frequencies.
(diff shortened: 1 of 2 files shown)


hera_cal/smooth_cal.py
```diff
@@ -656,9 +671,10 @@ def __init__(self, calfits_list, flag_file_list=[], flag_filetype='h5', antflag_
 
         Arguments:
             calfits_list: list of string paths to calfits files containing calibration solutions and flags
-            flag_file_list: list of string paths to files containing flags as a function of baseline, times
-                and frequency. Must have all baselines for all times. Flags on baselines are broadcast to both
-                antennas involved, unless either antenna is completely flagged for all times and frequencies.
+            flag_file_list: list of string paths to files containing flags as a function of time, frequency, and
+                either baseline, antnena, or just antenna polarization. Must have all baselines/ants/pols for all times.
```
tyler-a-cox, line 675: Small typo here "antnena"

tyler-a-cox (approved):
Okay, this looks good to me. There's just one typo in the doc-string of the CalibrationSmoother, but otherwise looks good

hera_cal/smooth_cal.py
jsdillon: fixed

PR #876, opened 2023-03-12
New abscal phase solver

tyler-a-cox:
This PR folds in @zacharymartinot new abscal phase gradient solver. The new solver uses FFTs in position space to solve for the per-frequency phase gradient, and seems much less sensitive to large errors in phase than the existing methods.

### DIFF
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -29,7 +29,7 @@ import argparse
 import numpy as np
 import operator
 from functools import reduce
-from scipy import signal, interpolate, spatial, constants
+from scipy import signal, interpolate, spatial, constants, fft
 from scipy.optimize import brute, minimize
 from pyuvdata import UVCal, UVData
 import linsolve
@@ -1421,6 +1421,7 @@ def _phase_gradient_solution(normalized_data_model_ratio, transformed_b_vecs, we
         Array of weights for each baseline. Shape (n_times, n_freqs, n_bls).
     resolution_factor : int
         Factor by which to increase the resolution of the FFT grid.
+    workers : int
     
     Returns:
     --------
@@ -1460,10 +1461,10 @@ def _phase_gradient_solution(normalized_data_model_ratio, transformed_b_vecs, we
             for nn in range(Ngroups):
                 grid[grid_indices[nn]] = weights_t_f[nn] * nmdr_t_f[nn]
                 
-            # Find grid point maximum in Fourier space corresponding to phase slope here 
-            grid_ft = np.fft.fftshift(np.fft.fftn(grid)) / Ngroups            
+            # Find grid point maximum in Fourier space corresponding to phase slope here
+            grid_ft = np.fft.fftshift(fft.fftn(grid, workers=-1)) / Ngroups            
             max_idx = np.unravel_index(np.argmax(np.real(grid_ft)), grid_ft.shape)
-            Lambda_init = np.array([ft_freqs[kk][ii] for kk,ii in enumerate(max_idx)])
+            Lambda_init = np.array([ft_freqs[kk][ii] for kk, ii in enumerate(max_idx)])
             
             # Refine solution with Newton's method
             Lambda_t_f, _ = _newton_solve_real(Lambda_init, transformed_b_vecs, nmdr_t_f, weights_t_f, 1e-8)
--- a/hera_cal/abscal.py
+++ b/hera_cal/abscal.py
@@ -1354,6 +1354,7 @@ def _grad_and_hess(x, B_vecs, Z_coefficients):
     w_cos_d = Z_coefficients.real
     w_sin_d = Z_coefficients.imag
 
+    # avoid complex data/evaluating complex exponentials with a little trigonometry
     # w*sin(x+y) = w*cos(x)*sin(y) + w*sin(x)*cos(y)
     w_sin_d_xB = cos_xB * w_sin_d + sin_xB * w_cos_d
 
@@ -1487,7 +1488,7 @@ def _phase_gradient_solution(Z_coefficients, transformed_b_vecs, resolution_fact
 
     # Initialize the grid and the corresponding frequencies
     grid = np.zeros(tuple(Nk_use), dtype=np.complex64)
-    ft_freqs = [-2*np.pi * fft.fftshift(fft.fftfreq(n)) for n in Nk_use]
+    ft_freqs = [- 2 * np.pi * fft.fftshift(fft.fftfreq(n)) for n in Nk_use]
 
     # Get the indices of the grid points corresponding to the transformed_b_vecs
     grid_indices = [tuple(ii for ii in transformed_b_vecs[nn]) for nn in range(Ngroups)]
@@ -1595,7 +1596,7 @@ def complex_phase_abscal(data, model, reds, data_bls, model_bls):
     # Get transformed antenna positions and baselines
     transformed_antpos = redcal.reds_to_antpos(reds)
     _put_transformed_array_on_integer_grid(transformed_antpos)
-    transformed_b_vecs = np.rint([transformed_antpos[jj] - transformed_antpos[ii] for (ii,jj,pol) in data_bls]).astype(int)
+    transformed_b_vecs = np.rint([transformed_antpos[jj] - transformed_antpos[ii] for (ii, jj, pol) in data_bls]).astype(int)
 
     # Get number of baselines and times/freqs
     Ngroups = len(data_bls)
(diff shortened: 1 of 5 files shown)


hera_cal/abscal.py
```diff
@@ -1291,6 +1295,280 @@ def global_phase_slope_logcal(model, data, antpos, reds=None, solver='linfit', w
...
+    weights = np.ones((Ntimes, Nfreqs, Ngroups), dtype=float)
+    for nn in range(Ngroups):
+        Vhat_n = data[data_bls[nn]]
+        Vbar_n = model[model_bls[nn]]
+        r_n = Vhat_n / Vbar_n
+        normalized_data_model_ratio[:, :, nn] = r_n / np.abs(r_n)
+        weights[:, :, nn] = np.abs(Vbar_n * Vhat_n)
+
+    # Get solution for degenerate phase slopes
+    weights /= np.mean(weights, axis=-1, keepdims=True)
```
jsdillon, line 1563: For ease of reading and logical flow, I'd put this line with the previous block

hera_cal/abscal.py
```diff
@@ -1291,6 +1295,280 @@ def global_phase_slope_logcal(model, data, antpos, reds=None, solver='linfit', w
...
+    Lambda_sol, Z_sol = _phase_gradient_solution(normalized_data_model_ratio, transformed_b_vecs, weights)
+
+    # turn solution into per-antenna gains
+    phase_angle = {a : np.sum(Lambda_sol * r, axis=-1) for a,r in transformed_antpos.items()}
+    delta_gains = {(a, utils.split_pol(pols[0])[0]): np.exp(1j * (angle)) for a, angle in phase_angle.items()}
+    meta = {'Lambda_sol':  Lambda_sol, 'Z_sol': Z_sol}
```
jsdillon, line 1569: Question for @zacharymartinot: is there anything else that's not too memory intensive that we want to pass through via this meta dict?

hera_cal/abscal.py
```diff
@@ -1291,6 +1295,280 @@ def global_phase_slope_logcal(model, data, antpos, reds=None, solver='linfit', w
...
+        Array of weighted normalized data/model ratios. Shape (n_bls,).
+
+    Returns:
+    --------
+    grad : np.ndarray or jnp.ndarray
+        Gradient of the objective function. Shape (n_dims,).
+    hess : np.ndarray or jnp.ndarray
+        Hessian of the objective function. Shape (n_dims, n_dims).
+    """
+    x_dot_b = jnp.dot(b_vec, x)
```
jsdillon, line 1345: Might be good to add comments (or more to the docstring) explaining the analytics here and why these sines and cosines are the right thing to do.

hera_cal/abscal.py
```diff
@@ -1291,6 +1295,280 @@ def global_phase_slope_logcal(model, data, antpos, reds=None, solver='linfit', w
...
+    i: int
+        Number of iterations performed by the Newton solver.
+    """
+    def inner_function(args):
+        """
+        Function evaluated on each iteration of the while loop
+        """
+        x, i, step_diff = args
+        grad, hess = _grad_and_hess_real(x, transformed_b_vecs, w_ndmr)
+        step = -1 * jnp.linalg.solve(hess, grad)
```
jsdillon, line 1389: Question for @zacharymartinot: Is there any reason to step less than -1 * the hessian (i.e. a damped iterator?)

hera_cal/abscal.py
```diff
@@ -1291,6 +1295,280 @@ def global_phase_slope_logcal(model, data, antpos, reds=None, solver='linfit', w
...
+                if np.max(np.abs(np.rint(this_dim_positions * numerator / denominator) - this_dim_positions * numerator / denominator)) < tol:
+                    break
+            
+            # Apply ratio found
+            for ap in transformed_antpos:
+                transformed_antpos[ap][dim] *= (numerator / denominator)
+
+def complex_phase_abscal(data, model, reds, data_bls, model_bls):
+    """
+    Calculates gains that would absolute calibrate already redundantly-calibrated data.
```
jsdillon, line 1516: This should clarify that it only performs absolute calibration of phase slopes, not amplitudes.

hera_cal/tests/test_abscal.py
```diff
@@ -661,6 +708,49 @@ def test_global_phase_slope_logcal_3D(self):
...
+        transformed_b_vecs = np.rint([transformed_antpos[jj] - transformed_antpos[ii] for (ii, jj, pol) in data_bls]).astype(int)
+
+        model, data = {}, {}
+        ntimes, nfreqs = 1, 2
+
+        # Test that the data is calibrated properly after being moved in phase
+        phase_deg = np.random.normal(0, 1, (ntimes, nfreqs, transformed_b_vecs.shape[-1]))
+        for bi, bls in enumerate(data_bls):
+            model[bls] = np.ones((ntimes, nfreqs))
+            data[bls] = model[bls] * np.exp(-1j * np.sum(transformed_b_vecs[bi][None, None, :] * phase_deg, axis=-1))
```
jsdillon, line 727: In theory, you shouldn't have to use the transformed vectors when "uncalibrating" here.

Also, it might be worth testing this on an array that has extra DoF... perhaps two widely separated hexagons, with a phase difference between then in addition to EW and NS phase gradients.

jsdillon (changes requested):
A few minor clarifications, but this is looking really good @tyler-a-cox. Thanks for cleaning this up, adding more documentation, and writing unit tests!

I'd like to get @zacharymartinot's go-ahead before we merge in.

hera_cal/tests/test_abscal.py
tyler-a-cox: Good point, I'll add that test.

hera_cal/abscal.py
tyler-a-cox: Yeah, that makes sense. I'll move that line up

hera_cal/abscal.py
tyler-a-cox: I'll make that change

hera_cal/abscal.py
tyler-a-cox: I did try this in another implementation of this function. I could just make the change and add it as a keyword argument.

hera_cal/abscal.py
tyler-a-cox: Will do

hera_cal/abscal.py
zacharymartinot: I thought about this briefly, but not very carefully. Its possible that including a damping factor could be useful when we're at the hairy edge of the model being "good enough". But I'm pretty sure a constant damping factor isn't going to do anything than add iterations in general, it would have to be something computed per step. and figuring that out would be a small project, if there's anything there at all. And if you're going down that road you might as well consider alternatives to newton's method entirely.

So until someone actually investigates, keep it simple. No damping.

jsdillon: Yeah, sounds like an unnecessary rabbit hole... just something I thought I'd mention.

zacharymartinot: I thought about this briefly, but not very carefully. Its possible that including a damping factor could be useful when we're at the hairy edge of the model being "good enough". But I'm pretty sure a constant damping factor isn't going to do anything than add iterations in general, it would have to be something computed per step. and figuring that out would be a small project, if there's anything there at all. And if you're going down that road you might as well consider alternatives to newton's method entirely.

So until someone actually investigates, keep it simple. No damping.

hera_cal/abscal.py
zacharymartinot: I added collection of the number of newton's method iterations. We should consider flagging on hitting maxiter.

hera_cal/abscal.py
zacharymartinot: This bit is gone now.

hera_cal/abscal.py
zacharymartinot: avoiding evaluating complex exponentials/creating complex arrays with a little trig

hera_cal/tests/test_abscal.py
zacharymartinot: I think that using the transformed vectors to uncalibrate is a fine unit test. And if we just add `split_core=True` then the test array will have 4 degeneracies.

zacharymartinot: I changed the test array to have 4 degeneracies.

hera_cal/abscal.py
zacharymartinot: Done.

jsdillon (approved):
Ok, I'm good to go!

zacharymartinot (approved):
Looks good to me. Thanks @tyler-a-cox!

PR #878, opened 2023-03-13
Fix the handling of new required UVCal parameters in `io.write_cal()`

jsdillon:
Specifically, this uses pyuvdata's list of known telescopes intelligently handle `lst_array`, `telescope_location`, `antenna_positions` if not provided by the user via `lsts` or `kwargs`.


bhazelton (approved):
This looks good to me.

PR #883, opened 2023-03-30
feat: allow setting freq_scale to zero to avoid smoothing

steven-murray:
This just makes it possible to set `time_scale` and/or `freq_scale` to zero to avoid actually doing smoothing. While this is maybe a little counter-intuitive, it makes it easier to have pipelines with a `CAL_SMOOTH` step where the refant is chosen, and `.smooth.calfits` files are written, but without actually doing smoothing.

### DIFF
--- a/scripts/smooth_cal_run.py
+++ b/scripts/smooth_cal_run.py
@@ -38,11 +38,18 @@ def run():
                                 chan_blacklists=a.chan_blacklists, pick_refant=a.pick_refant, freq_threshold=a.freq_threshold,
                                 time_threshold=a.time_threshold, ant_threshold=a.ant_threshold, verbose=a.verbose)
         if a.axis == 'both':
-            cs.time_freq_2D_filter(freq_scale=a.freq_scale, time_scale=a.time_scale, tol=a.tol,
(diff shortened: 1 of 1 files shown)


jsdillon:
I don't think this will actually work. The proper way to do this is with the `axis` argument.

jsdillon:
But I'm fine adding in an `axis = 'none'` way to not smooth along either axis but still do the refant stuff.

hera_cal/smooth_cal.py
```diff
@@ -1060,7 +1060,7 @@ def smooth_cal_argparser():
                           Only used when the filtering method is 'DPSS'")
     flt_opts.add_argument("--dont_skip_flagged_edges", action="store_true", default=False, help="if True, use DPSS over integrations with flagged edge channels.\
                           Only used when method used is 'DPSS'")
-    flt_opts.add_argument("--axis", default="both", type=str, help="smooth either in 'freq', or 'both' (time and freq) axes.")
+    flt_opts.add_argument("--axis", default="both", type=str, help="smooth either in 'freq', 'both' (time and freq) axes or 'none'.")
```
jsdillon, line 1063: I'd clarify here what 'none' dones.

jsdillon (approved):
Looks good. Just one help string that I think should be clarified.

PR #885, opened 2023-04-08
Make smooth_cal more tolerant of small gains and nans/infs (as long as they are flagged)

jsdillon:
These small tweaks makes smooth_cal a little more robust to unexpected inputs.


tyler-a-cox (approved):
This seems pretty straightforward to me, approved

PR #886, opened 2023-04-09
Fix reconstitute interleaving features

aewallwi:
Modify `vis_clean.time_chunk_from_baseline_chunks` code so that all interleaves will fall into the same final labeled time bin in makeflow which are set by the LSTs every `stride_length`. Without this fix, the time averages for different "interleaves" can fall into different makeflow file indices when we undo the cornerturn after performing time averaging.


jsdillon (changes requested):
Mostly looks good. Just a few questions and clarifications.

hera_cal/tests/test_vis_clean.py
```diff
@@ -1130,46 +1131,53 @@ def test_time_chunk_from_baseline_chunks_argparser(self):
...
             assert len(hd_fragment.bls) == len(baselines)
             assert hd_fragment.Ntimes == 60
             assert hd_fragment.Nfreqs == 20
 
-        fragments = glob.glob(DATA_PATH + '/test_output/temp.fragment.h5.part*')
+        fragments = sorted(glob.glob(DATA_PATH + '/test_output/temp.fragment.h5.part*'))
         # reconstitute the filtered data
         for filenum, file in enumerate(datafiles):
             # reconstitute
+            # AEW -- 5-10-2023 -- I AM HERE!
```
steven-murray, line 1168: Good to know!

hera_cal/vis_clean.py
```diff
@@ -794,8 +794,7 @@ def vis_clean(self, keys=None, x=None, data=None, flags=None, wgts=None,
             # get filter properties
             mfrate = max_frate[k] if max_frate is not None else None
             filter_centers, filter_half_widths = gen_filter_properties(ax=ax, horizon=horizon,
-                                                                       standoff=standoff, min_dly=min_dly,
-                                                                       bl_len=self.bllens[k[:2]], max_frate=mfrate)
+            standoff=standoff, min_dly=min_dly,                                                                    bl_len=self.bllens[k[:2]], max_frate=mfrate)
```
steven-murray, line 797: I can't tell what's happening on this line, but it seems weird

hera_cal/vis_clean.py
```diff
@@ -2023,7 +2022,9 @@ def time_chunk_from_baseline_chunks(time_chunk_template, baseline_chunk_files, o
         output will trim the extra frequencies in the time_chunk and write out trimmed freqs. The same is true
         for polarizations.
     baseline_chunk_files : list of strings
-        list of paths to baseline-chunk files to select time-chunk file from.
+        list of paths to baseline-chunk files to select time-chunk file from. If the files have "_interleave" in their title then
+        the method will automatically identify the number of unique interleaves, chunk the file list up into interleaved sets and
+        retrieve integrations based on the time in the first file.
```
steven-murray, line 2027: For me, just having read this docstring, I'm still not clear on exactly what's happening here. Can you maybe provide a more clear example in the docstring?

hera_cal/vis_clean.py
```diff
@@ -2040,11 +2041,38 @@ def time_chunk_from_baseline_chunks(time_chunk_template, baseline_chunk_files, o
     -------
         Nothing
     """
+    # check whether "interleave" is in baseline chunk filenames. If so, make sure that they all have "interleave",
+    # split them into sets, and make sure that the sets all have the same number of files.
+    interleave_mode = "interleave" in baseline_chunk_files[0]
```
steven-murray, line 2046: this is not consistent with the docstring, which states that if "_interleave" is in the filename, it'll do the interleave mode

hera_cal/vis_clean.py
```diff
@@ -2040,11 +2041,38 @@ def time_chunk_from_baseline_chunks(time_chunk_template, baseline_chunk_files, o
...
+    
+    for fname in baseline_chunk_files:
+        if "interleave" not in fname and interleave_mode or \
+           "interleave" in fname and not interleave_mode:
+            raise ValueError("must not have a subset of files with 'interleave' in name.")
```
steven-murray, line 2051: I think a nicer error here would be "Cannot have some baseline_chunk_files with "_interleave" in their name, while other's don't"

steven-murray: Furthermore, add this info to the docstring (i.e. say that all or none of the files must have `"_interleave"` in the name)

hera_cal/vis_clean.py
```diff
@@ -2040,11 +2041,38 @@ def time_chunk_from_baseline_chunks(time_chunk_template, baseline_chunk_files, o
...
+    if interleave_mode:
+        interleave_indices = np.unique([int(re.findall("interleave_[0-9]{1,10}", fname)[0][11:]) for fname in baseline_chunk_files])
```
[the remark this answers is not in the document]
steven-murray: Agreed. I actually think a much more flexible and future-proof way of doing this would be to have a new argument to the function, like `interleave_regex`, which if provided, will do the above (it could be the above by default, but probably better to define it with a capture group instead).

steven-murray (commented):
Thanks @aewallwi -- I have a few comments that I think will make this a little more future-robust, but otherwise it seems good.

PR #887, opened 2023-04-10
feat: add more robust RedundantGroups class

steven-murray:
This just adds a new `RedundantGroups` class that does a very similar thing to the `RedDataContainer` except that it doesn't worry about actual data, i.e. it's just a management class for the redundant baseline groups themselves. I've added an attribute to the `RedDataContainer` that points to one of these objects, which stands in as the redundant groups of the *array* rather than the data itself.

Some points I noticed on the way:

1. The `test_redcal_run` test was failing on my own system in this PR, and I thought I'd screwed something up, but I tested it on main as well and it failed. That same test is one that randomly fails in `linsolve` sometimes, so I think there's something brittle about it that I don't understand. 
2. I think the fact that the `RedDataContainer` is weirdly mixing data keys and array keys is making some things not work. By that, I mean that the key by which it indexes a redundant group is *either* the first bl in the group that also exists in the data, if such a key exists, OR it's just the first in the group. This is going to mean that when you read different files and set the keys by which redundant averaged data is indexed, they can be different. I think it might make more sense to be a bit more clear about this distinction in the class.


hera_cal/datacontainer.py
```diff
@@ -517,19 +518,30 @@ class RedDataContainer(DataContainer):
...
-                Will error if one tries to provide both reds and antpos. If neither is provided,
-                will try to to use data.antpos (which it might have if its is a DataContainer).
-            bl_error_tol: the largest allowable difference between baselines in a redundant group
-                (in the same units as antpos). Normally, this is up to 4x the largest antenna position
-                error. Will only be used if reds is inferred from antpos.
+        Parameters
+        ----------
+        data: DataContainer or dictionary of visibilities, just as one would pass into DataContainer().
+            Will error if multiple baselines are part of the same redundant group.
+        reds: list of lists of redundant baseline tuples, e.g. (ind1, ind2, pol).
```
jsdillon, line 528: You might want to mention that this can now be a RedundantGroups object, both here and in the type-hint

hera_cal/datacontainer.py
```diff
@@ -517,19 +518,30 @@ class RedDataContainer(DataContainer):
...
+            error. Will only be used if reds is inferred from antpos.
+
+        Attributes
+        ----------
+        reds 
+            list of lists of redundant baseline tuples, e.g. (ind1, ind2, pol).
+        red_groups
+            A :class:`RedundantGroups` object that contains the redundant groups for 
+            the entire array, and methods to manipulate them. 
+            When possible, you should use this object instead of the reds attribute.
```
jsdillon, line 544: Having both is very confusing. I thought the idea was for the new object to be a drop-in replacement.

hera_cal/datacontainer.py
```diff
@@ -558,14 +573,22 @@ def _add_red(self, ubl_key, red):
...
         '''
         # Map all redundant keys to the same underlying data
+        self._red_key_to_bls = {}
+        self._bl_to_red_key = {}
+        
         self.reds = []
+        if isinstance(reds, RedundantGroups):
+            self.red_groups = reds
+        else:
+            self.red_groups = RedundantGroups(red_list=reds, antpos=getattr(self, 'antpos', None))
```
jsdillon, line 590: If it's not a `RedundantGroups` object, we should recast it on __init__

hera_cal/red_groups.py
```diff
@@ -0,0 +1,246 @@
...
+    This class is used to manage redundant baseline groups. It is (roughly)
+    API-compatible with the list-of-lists-of-tuples format returned by, e.g., 
+    :func:`hera_cal.redcal.get_reds`, and in the future that may simply return
+    one of these objects.
+
+    To keep this compatibility, this object is mutable, but we absolutely discourage
+    directly mutating it. Instead, use the methods provided, including the `.append`
+    method, which is able to add a new redundant group to the object. 
+
+    In the future, we may disable direct mutation of this object.
```
jsdillon, line 31: I don't really see this happening...

hera_cal/red_groups.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def __iter__(self):
+        return iter(self._red_list)
+    
+    def __getitem__(self, key: int | AntPair | Baseline) -> list[AntPair | Baseline]:
+        if isinstance(key, int):
+            return self._red_list[key]
+        else:
+            return self.get_red(key)
+
```
jsdillon, line 246: I'd like to also support __setitem__, __add__, extend, and sort

jsdillon, line 246: I'd like to see this object support redcal.filter_reds performed on itself.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,246 @@
...
+    ----------
+    antpos : dict[int, np.ndarray]
+        Dictionary mapping antenna number to antenna position in ENU frame.
+        Not required if ``red_list`` is provided, but highly encouraged even if so.
+    bl_error_tol : float
+        Baseline length error tolerance in meters.
+    include_autos : bool
+        Whether to include autocorrelations in the redundant baseline groups.
+    red_list : list[list[tuple[int, int]]]
+        List of redundant baseline groups. Each group is a list of antenna pairs.
```
jsdillon, line 43: I think we ought to be moving away from antpairs wherever possible. It's an older access pattern that isn't used much and I'm fine restricting this class to only doing len-3 bl tuples. If someone really needs len-2 tuples, they can use a list of lists. But I'd like to push us away from that since it creates so many annoying special cases.

As it stands, does this support RedundantGroups with different lists of baselines per polarization? It's not clear to me.

jsdillon (commented):
not finished going through this yet, but thought I'd leave some comments since I need to put this PR review down for a few hours

hera_cal/datacontainer.py
steven-murray: Yeah, I originally replaced `reds` with the `RedundantGroups` object, but came across a few errors. I think with my additional understanding of how it works now (and the addition of the `.append()` method on the class) I could exactly replace it.

hera_cal/red_groups.py
steven-murray: Fair enough.

hera_cal/red_groups.py
steven-murray: Cool, will do.

steven-murray: Sure!

hera_cal/red_groups.py
steven-murray: Actually, what this class does is natively store the redundancies as simply antpairs, because that's all that's needed. However, when doing things like `__getitem__` with a key, the key can _either_ be an antpair or len-3 baseline tuple, and the result will be in the same form. I felt like this was the cleanest way. 

The answer to your question is _no_, it doesn't support different lists of baselines per polarization, but I think that makes sense, because this class is not meant to represent the actual baselines in any given dataset, but rather the set of redundancies in an array.

hera_cal/datacontainer.py
steven-murray: I'm recasting it there, but of course you can also call `build_red_keys` directly so we need to also cast it here.

hera_cal/red_groups.py
steven-murray: I gave in and let it natively use antpair-pol tuples (it can actually use either).

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def get_full_redundancies(
+        self, pols: Sequence[str] | None = None, 
+        include_autos: bool = True, 
+        bl_error_tol: float = 1.0,
+        pol_mode: str='1pol',
+    ) -> RedundantGroups:
+        """Create a RedundantGroups object from an antpos dictionary."""
+        if pols is None:
+            if len(next(iter(self._bl_to_red_map.values()))) == 2:
```
jsdillon, line 374: Is it possible for this dict to be a mix of 2-tuples and 3-tuples, making the outcome of this command uncertain?

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+                pols = ()
+            else:
+                pols = tuple({bl[2] for bl in self._red_key_to_bls_map.keys()})
+
+        if pols:
+            reds = redcal.get_reds(
+                self.antpos, 
+                pols=pols, # we end up removing the pols anyway.
```
jsdillon, line 382: Out of date comment

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+        pol_mode: str='1pol',
+    ) -> RedundantGroups:
+        """Create a RedundantGroups object from an antpos dictionary."""
+        if pols is None:
+            if len(next(iter(self._bl_to_red_map.values()))) == 2:
+                pols = ()
+            else:
+                pols = tuple({bl[2] for bl in self._red_key_to_bls_map.keys()})
+
+        if pols:
```
jsdillon, line 379: Can you re-use self.from_antpos here?

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+            reds = redcal.get_pos_reds(
+                self.antpos, 
+                bl_error_tol=bl_error_tol, 
+                include_autos=include_autos
+            )
+
+        return attrs.evolve(self, red_list=reds)
+    
+    def keyed_on_bls(self, bls: Sequence[AntPair | Baseline]) -> RedundantGroups:
+        """Return a new RedundantGroups object keyed on the given baselines."""
```
jsdillon, line 397: I'm not quite sure I follow the logic here... perhaps you can clarify in the docstring what happens if it can't find certain baselines or if a redundant group doesn't have a baseline in the list.

hera_cal/datacontainer.py
```diff
@@ -540,67 +558,84 @@ def __init__(self, data, reds=None, antpos=None, bl_error_tol=1.0):
...
-            self._red_key_to_bls[reverse_bl(ubl_key)].append(reverse_bl(bl))
-
-    def build_red_keys(self, reds):
+
+    def build_red_keys(self, reds: RedundantGroups | list[list[Baseline]]):
         '''Build the dictionaries that map baselines to redundant keys.
 
         Arguments:
-            reds: list of lists of redundant baseline tuples, e.g. (ind1, ind2, pol).
+            reds: list of lists of redundant baselin1e tuples, e.g. (ind1, ind2, pol).
```
jsdillon, line 581: typo

hera_cal/datacontainer.py
```diff
@@ -540,67 +558,84 @@ def __init__(self, data, reds=None, antpos=None, bl_error_tol=1.0):
...
+        '''
+        return self._reds_keyed_on_data.get_ubl_key(bl)
+        
     def get_red(self, key):
-        '''Returns the list of baselines redundant with this key.'''
-        return self._red_key_to_bls[self._bl_to_red_key[key]]
+        '''Returns the list of baselines in the array redundant with this key.
+        
+        Note: this is not just baselines existing in the data itself, but in the 
+              entire array.
```
jsdillon, line 616: I'm not sure that this is the behavior we want... is this still true? I could imagine this breaking existing code.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+    
+    def extend(self, reds: Sequence[Sequence[AntPair | Baseline]], inplace: bool = True) -> None:
+        """In-place extend the list of redundant groups with another list of redundant groups.
+        
+        This maintains the list-of-lists duck-typing of the redundant groups.
+        """
+        for red in reds:
+            self.append(red, inplace)
+    
+    def sort(self):
```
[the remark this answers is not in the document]
steven-murray: Ah yeah.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
+"""Module with classes and functions for dealing with redundant baseline groups."""
+from __future__ import annotations
+
+import attrs
```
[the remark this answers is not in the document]
steven-murray: Good point. Done.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+from .utils import reverse_bl
+from . import redcal
+import numpy as np
+import copy
+from functools import cached_property, wraps, partial
+from frozendict import frozendict
```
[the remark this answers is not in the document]
steven-murray: I don't like introducing new dependencies, but I do think it's important to protect against this. Changing the antpos in-place will make the object inconsistent and there's no way for the object to know that this happened. Making sure the antpos really are constant protects against this, and the frozendict module is very small and simple.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+    None of the mutating methods on the object will ever mutate the antpos, only the
+    redundant baseline group lists. The idea here is that the antpos are meant to be
+    *fixed* to the full array antpos, but you may consider subsets of the baselines.
+
+
+    Parameters
+    ----------
+    antpos : dict[int, np.ndarray]
+        Dictionary mapping antenna number to antenna position in ENU frame.
+        Not required if ``red_list`` is provided, but highly encouraged even if so.
```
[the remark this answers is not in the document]
steven-murray: Yup good point. I do mention this in the notes, but would be good to do here also.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+    You may add together two RedundantGroups objects, which will combine their redundant
+    baseline groups and their antennas (as found in ``antpos``)::
+
+        >>> rg2 = rg + rg
+        >>> rg2.data_bls == rg.data_bls
+        True
+    
+    To filter out baselines for a specific purpose (but keeping the full antpos array)::
+
+        >>> filtered = rg.filtered(bls=[(0, 1), (1, 2), (2, 3)])
```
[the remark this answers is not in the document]
steven-murray: Yeah, I've now linked to the `filter_reds` function in the docs here. I didn't want to re-document the whole thing in this little example.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+    def from_antpos(
+        cls, 
+        antpos: dict[int, np.ndarray], 
+        pols: Sequence[str] = (), 
+        include_autos: bool = True, 
+        bl_error_tol: float = 1.0,
+        pol_mode: str='1pol',
+        **kwargs
+    ) -> RedundantGroups:
+        """Create a RedundantGroups object from an antpos dictionary."""
```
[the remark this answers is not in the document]
steven-murray: Good point

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+        if pols:
+            reds = redcal.get_reds(
+                antpos, 
+                pols=pols, # we end up removing the pols anyway.
```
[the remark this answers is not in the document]
steven-murray: woops. removed.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+    
+    @cached_property
+    def data_ants(self) -> frozenset[int]:
+        """The list of antennas that have are in the baseline groups."""    
+        out = set()
+        for red in self._red_list:
+            for ap in red:
+                out.add(ap[0])
+                out.add(ap[1])
+        return out
```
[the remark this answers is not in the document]
steven-murray: Yup, I like that.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+            reds = redcal.get_pos_reds(
+                antpos, 
+                bl_error_tol=bl_error_tol, 
+                include_autos=include_autos
+            )
+        return cls(antpos=antpos, red_list=reds, **kwargs)
+    
+    @cached_property
+    def data_ants(self) -> frozenset[int]:
+        """The list of antennas that have are in the baseline groups."""
```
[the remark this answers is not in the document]
steven-murray: noted

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+                out[bl] = self.key_chooser(red)
+                out[reverse_bl(bl)] = rev 
+        return out
+    
+    def get_ubl_key(self, key: AntPair | Baseline) -> AntPair | Baseline:
+        '''Returns the key used interally denote the data stored. Useful for del'''
+        return self._bl_to_red_map[key]
+        
+    def get_red(self, key: AntPair | Baseline) -> tuple[AntPair | Baseline]:
+        '''Returns the tuple of baselines redundant with this key.'''
```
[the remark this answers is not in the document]
steven-murray: tuples are immutable, and this shouldn't be mutated. But I've made it a list to increase the list-of-lists feel.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+        out = {}
+        for red in self._red_list:
+            rev = reverse_bl(self.key_chooser(red))
+            for bl in red:
+                out[bl] = self.key_chooser(red)
+                out[reverse_bl(bl)] = rev 
+        return out
+    
+    def get_ubl_key(self, key: AntPair | Baseline) -> AntPair | Baseline:
+        '''Returns the key used interally denote the data stored. Useful for del'''
```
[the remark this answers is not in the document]
steven-murray: got it.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+            self.append(value)
+
+        # Reset the maps
+        self.clear_cache()
+
+    def index(self, key: AntPair | Baseline) -> int:
+        """Return the index of a redundant group."""
+        return list(self._red_key_to_bls_map.keys()).index(self.get_ubl_key(key))
+    
+    def filtered(self, **kwargs) -> RedundantGroups:
```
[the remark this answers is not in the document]
steven-murray: I've renamed it. I originally tried to actually just set the `__doc__` to the `redcal.filter_reds.__doc__` but it gave me a ciruclar import. I can just reference those docs in the docs. Also added the `inplace` argument against my judgment (I really do prefer immutability wherever possible...)

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+        """Return a new RedundantGroups object with baselines filtered out.
+
+        """
+        new_reds = redcal.filter_reds(copy.deepcopy(self), **kwargs)
```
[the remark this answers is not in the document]
steven-murray: good point, done.

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+    red_list : list[list[tuple[int, int]]]
+        List of redundant baseline groups. Each group is a list of antenna pairs.
+
+    Examples
+    --------
+    You can create a RedundantGroups object from a list of redundant baseline groups::
+
+        >>> from hera_cal.red_groups import RedundantGroups
+        >>> reds = [[(0, 1), (1, 2), (2, 3)], [(0, 2), (1, 3), (2, 4)]]
+        >>> rg = RedundantGroups(red_list=reds)
```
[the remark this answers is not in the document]
steven-murray: yes, this is already True, so I can use that in the docs here.

hera_cal/redcal.py
```diff
@@ -1887,8 +1887,11 @@ def redcal_iteration(hd, nInt_to_load=None, pol_mode='2pol', bl_error_tol=1.0, e
             # try to solve for gains on antennas excluded from calibration, but keep them flagged
             expand_omni_gains(sol, all_reds_this_pol, data, nsamples, chisq_per_ant=meta['chisq_per_ant'])
 
-            # try one more time to expand visibilities, keeping new ones flagged, but don't update chisq or chisq_per_ant
+            # try one more time to expand visibilities, keeping new ones flagged, but 
+            # don't update chisq or chisq_per_ant
             expand_omni_vis(sol, all_reds_this_pol, data, nsamples)
+            print(sol.vis.bls())
+            print(type(sol.vis))
```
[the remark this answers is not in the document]
steven-murray: damnit

hera_cal/red_groups.py
```diff
@@ -0,0 +1,407 @@
...
+    def data_bls(self) -> frozenset[Baseline | AntPair]:
+        """The list of baselines in the baseline groups."""
+        return set(self._bl_to_red_map.keys())
+
+    @cached_property
+    def _red_key_to_bls_map(self) -> dict[AntPair, tuple[AntPair]]:
+        out = {}
+        for red in self._red_list:
+            out[self.key_chooser(red)] = red
+            out[reverse_bl(self.key_chooser(red))] = tuple(reverse_bl(r) for r in red)
```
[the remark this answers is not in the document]
steven-murray: yeah, I made them all lists. I don't *like* it because lists are mutable, but it's probably not the worst thing.

hera_cal/red_groups.py
steven-murray: No.

hera_cal/red_groups.py
steven-murray: fixed

hera_cal/red_groups.py
steven-murray: good catch!

hera_cal/red_groups.py
steven-murray: OK I've given it a longer explanation. Basically the behaviour is the same as was happening in `RedDataContainer`.

hera_cal/datacontainer.py
steven-murray: I'm fairly confident this was the behaviour of the existing code.

hera_cal/red_groups.py
```diff
@@ -161,11 +198,35 @@ def from_antpos(
...
-        """Create a RedundantGroups object from an antpos dictionary."""
+        """Create a RedundantGroups object from an antpos dictionary.
+
+        Parameters
+        ----------
+        antpos
+            A dictionary mapping antenna numbers to their positions in the array. 
+            The positions should be 3-element arrays of floats in ENU coordinates in 
+            meters. Provide all antennas in the array, even if they are not present
+            in the baselines you want to use in the end. You can always use `.filtered`
```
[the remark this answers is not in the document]
steven-murray: Ah yeah, fixed.

hera_cal/red_groups.py
```diff
@@ -101,24 +127,34 @@ class RedundantGroups:
         >>> rg2.data_bls == rg.data_bls
         True
     
-    To filter out baselines for a specific purpose (but keeping the full antpos array)::
-
-        >>> filtered = rg.filtered(bls=[(0, 1), (1, 2), (2, 3)])
+    To filter out baselines for a specific purpose (but keeping the full antpos array), 
+    the ``filtered`` method can be used. This will return a *new* RedundantGroups object
```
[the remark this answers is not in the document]
steven-murray: Ah yeah, fixed.

hera_cal/red_groups.py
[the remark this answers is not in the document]
jsdillon: Did you still want to do this?

jsdillon (approved):
Just one unaddressed comment, but it's minor so I'm OK approving. When you do pull this in, please make an issue to replace list-of-lists elsewhere int he code with the new object.

hera_cal/red_groups.py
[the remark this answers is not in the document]
steven-murray: Oh! Yeah I'll do this now.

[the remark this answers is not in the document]
steven-murray: Done.

tyler-a-cox (approved):
This looks great! I don't have much more to add that what Josh already covered. Excited to start using this!

PR #891, opened 2023-04-20
LST-binned Calibration

tyler-a-cox:
This PR adds a new module that allows the user to calibrate days within a particular LST-bin with respect to each other. The goal of this approach is to reduce the variance in an LST-bin before binning data together by correcting deviations in per-day calibration. This PR only supports modifying the `abscal` degrees of freedom and does not allow per-antenna calibration to be touched.


jsdillon:
closing in favor of #953

Issue #893, opened 2023-05-09
`datacontainer.RedDataContainer` has slow setter method


tyler-a-cox:
Updating and setting data in `datacontainer.RedDataContainer` is slower (~0.2 seconds per key set) now that we've migrated to using `red_group.RedundantGroups` to track redundancies. It seems that this is due to a new instance of `RedundantGroups` being created when running  `RedDataContainer.__setitem__`, which requires a new mapping between baselines and the baseline key representing a redundant group (in `_bl_to_red_map`) to be computed.

tyler-a-cox:
Resolved by https://github.com/HERA-Team/hera_cal/pull/896

PR #894, opened 2023-05-09
Modify RedDataContainer setter method

tyler-a-cox:
Addresses https://github.com/HERA-Team/hera_cal/issues/893


jsdillon (approved):
Seems reasonable to me. If @steven-murray is happy, I am.

steven-murray:
I am going to have a bit more of a look, because I think this might technically break the logic, and also doesn't solve the general case. I will get to it soon

steven-murray:
@tyler-a-cox actually I think the logic is sound, and this will speed up some cases (I presume your case, at least). It won't speed up the case where you start with no baselines in your object, then you go and set baselines, adding them in iteratively. I think I can make another PR to handle that case.

PR #896, opened 2023-05-10
perf: faster updating of RedundantGroups objects

steven-murray:
This adds the ability to update the underlying dictionaries mapping baseline groups to their keys (and vice versa) in-place without re-calculating the whole dictionary. This should be quite a bit faster than the original. It also speeds up re-keying on a subset of baselines, which was the problem @tyler-a-cox was seeing previously.


tyler-a-cox (approved):
Thanks for adding this! This looks significantly more robust than what I had written in the last PR. Happy to approve this.

PR #897, opened 2023-05-14
Detect and account for 180 degree phase flips in smooth_cal

jsdillon:
This PR adds automatic detection of 180 degree phase flips (on by default) before smoothing, which are incorporated in the rephasor applied before smoothing and then taken back out after smoothing.

It also adds functionality to flag integrations before and after phase flips, under the assumption that the phase flip may have happened mid-integration.

Here's a demonstration:

image

image

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -960,9 +960,9 @@ class CalibrationSmoother():
                         self.flag_grids[ant][:, i] = np.ones_like(self.flag_grids[ant][:, i])
         self.rephase_to_refant(warn=False)
 
-    def time_freq_2D_filter(self, freq_scale=10.0, time_scale=1800.0, tol=1e-09, filter_mode='rect',
-                            window='tukey', maxiter=100, method="CLEAN", fit_method='pinv', eigenval_cutoff=1e-9,
-                            skip_flagged_edges=True, fix_phase_flips=False, flag_phase_flip_ints=False, **win_kwargs):
+    def time_freq_2D_filter(self, freq_scale=10.0, time_scale=1800.0, tol=1e-09, filter_mode='rect', window='tukey',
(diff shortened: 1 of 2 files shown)


hera_cal/smooth_cal.py
```diff
@@ -1000,8 +1033,21 @@ def time_freq_2D_filter(self, freq_scale=10.0, time_scale=1800.0, tol=1e-09, fil
...
-                                                     method=method, fit_method=fit_method, cached_input=cached_input, 
-                                                     skip_flagged_edges=skip_flagged_edges, **win_kwargs)
+                                                     method=method, fit_method=fit_method, cached_input=cached_input,
+                                                     skip_flagged_edges=skip_flagged_edges, fix_phase_flips=fix_phase_flips, **win_kwargs)
+                flipped = (info['phase_flips'] == -1)
+                if np.any(flipped):
+                    print(f'{np.sum(flipped)} phase-flipped integrations detected on antenna {ant} between {self.time_grid[flipped][0]} and {self.time_grid[flipped][-1]}.')
+                    # apply flags before and after phase flips
+                    most_recent_unflagged_tind = 0
+                    if flag_phase_flip_ints:
```
tyler-a-cox, line 1043: If I'm remembering correctly, I think there were some discussions in the general meeting about whether we should completely flag antennas where we find phase flips. Would it be worth adding another option to completely flag an antenna if you find a phase flip?

tyler-a-cox (commented):
Looks good to me, @jsdillon! Just had one question about whether to fully flag an antenna if we find phase flips.

hera_cal/smooth_cal.py
jsdillon: Great idea! Done.

jsdillon:
Here's a demonstration with a full day

image

tyler-a-cox (approved):
This looks good to me, thanks for the fix @jsdillon!

PR #898, opened 2023-05-27
Fix issue with total_quality_array shape that arises in the newest pyuvdata

jsdillon:


hera_cal/io.py
```diff
@@ -984,9 +987,9 @@ def partial_write(self, output_path, data=None, flags=None, nsamples=None,
 
         # else:  # make a copy of this object and then update the relevant arrays using DataContainers
         #     this = copy.deepcopy(self)
-        
-        hd_writer.write_uvh5_part(output_path, d, f, n, 
-                                  run_check_acceptability=output_path in self._writers, 
+
+        hd_writer.write_uvh5_part(output_path, d, f, n,
+                                  run_check_acceptability=output_path in self._writers,
```
bhazelton, line 992: I'm confused about setting `run_check_acceptability=output_path`, I think `run_check_acceptability` should be a boolean.

jsdillon: it's `run_check_acceptability=output_path in self._writers`, which isn't wrong but would be clearer with parentheses. Added those.

bhazelton (approved):
Looks reasonable to me.

PR #899, opened 2023-05-29
Update logic for identifying phase flips

jsdillon:
In analyzing a day outside H6C IDR2 (specifically 2459906), I found that phase shifts in the calibration solution due to bright sources (e.g. the galactic anticenter) could trigger the detection of phase flipping:

image

This is not the intended behavior, so I changed the logic to identify phase flips based on changes of 90 degrees in the phase from one integration to the next. I have verified on 2459866 and 2459870 that the same integrations are being identified as phase flipped with the new logic as with the old.

The new method is very, very slightly slower, but not in a way that increases the runtime of smooth_cal by more than a few seconds compared to the ~30 min runtime of the smooth_cal notebook (which is I/O dominated anyway).

Doing this enabled a more natural way to handle completely flagged integrations, which is to give them have the same flipped-status as the most-recent non-flagged integration. So I've made that change, which required editing a unit test, but should have few practical implications.

### DIFF
--- a/hera_cal/smooth_cal.py
+++ b/hera_cal/smooth_cal.py
@@ -78,11 +78,20 @@ def detect_phase_flips(phases):
         phase_flipped: boolean array the same shape as phases where True means pi radians flipped
             relative to the first unflagged phase.
     """
-    complexes = np.exp(1.0j * phases)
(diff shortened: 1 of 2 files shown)


tyler-a-cox (approved):
Okay, this seems like a pretty straightforward change. Thanks for fixing this, @jsdillon!

PR #901, opened 2023-06-22
fix: future array shapes in chunk test and deprecation in numpy equals

steven-murray:
Attempt to fix the problem with a huge memory allocation in the dta chunker


jsdillon (approved):
Happy with this, though confused about these "uncommitted-changes.json" files

PR #903, opened 2023-06-27
Fix `build_fringe_rate_profiles` bug when number of pols = 1

jsdillon:
This PR fixes a small bug where `.squeeze()`, which is designed to eliminate extraneous dimensions in pyuvdata objects, can also squeeze out dimensions in the `flag_array` when the number of polarizations is 1. This PR prevents that by first doing a np.all over the polarization axis, as was intended, then squeezing (which does nothing when `future_array_shapes == True`).

This PR also fixes a bunch of pep8 stuff.


jsdillon:
@aewallwi Not asking you for a review (I know you're busy!) but I thought I'd just flag this for you in case you were curious or if it impacted H4C analysis (I doubt it).

tyler-a-cox (approved):
Seems like a pretty straightforward fix, approved. Thanks @jsdillon!

aewallwi:
Thanks for keeping me in the loop @jsdillon! This all looks good :)

PR #904, opened 2023-06-28
Allow an advanced user to speed-up `build_fringe_rate_profiles`

jsdillon:
This PR allows an advanced user to skip several time-consuming steps in `frf.build_fringe_rate_profiles()`. I find this particularly useful in the case I'm working on, where I want to compute the FR profile for a single baseline over many different bands. This allows me to speed up that calculation by a factor of ~10 by skipping certain steps that are unnecessary or doing them just once.

There is no change to the default behavior of the function.

### DIFF
--- a/hera_cal/frf.py
+++ b/hera_cal/frf.py
@@ -324,8 +324,9 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0,
     return frate_centers, frate_half_widths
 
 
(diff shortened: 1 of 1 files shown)


jsdillon:
@aewallwi Again, I'm tagging you so that you're aware of this... not necessarily to review.

aewallwi (approved):
This all looks good to me! Glad to see sandboxing options being added to help speed up analysis!

PR #909, opened 2023-09-07
fix some things that will error in pyuvdata 3.0

bhazelton:
This future proofs for pyuvdata v3.0 which is coming soonish.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -9,6 +9,7 @@ import operator
 import os
 import copy
 import warnings
+from packaging import version
 from functools import reduce
 from collections.abc import Iterable
 from pyuvdata import UVCal, UVData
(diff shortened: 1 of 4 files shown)


jsdillon:
This PR is failing a bunch of tests... do we expect it to pass before pyuvdata 3.0?

bhazelton:
Sigh. I though it would but I forgot that I fixed the spelling of the `flags_array` to `flag_array` in `write_uvh5_part` because it didn't match the spelling throughout the rest of the code base. I'll fix it to work now and in the future, it'll just be a little clunky.

jsdillon (approved):
That all looks good to me. If you don't mind adding an issue to remove that reference to the pyuvdata version... eventually, I'm sure we'll require v3.0 or later, at which point we can remove that hacky bit.

bhazelton:
Ok, I made the issue. What do you want to do about the code coverage? I can add a `pragma: no cover` to the code branch for pyuvdata > 3.0 if you like. The other line was apparently already not covered, I just made the syntax pep8 compliant.

jsdillon:
I can live with the tiny decrease in coverage.

PR #912, opened 2023-09-20
Fix LSTBIN for redundantly-averaged inpainted data

steven-murray:
Makes the following fixes:

* Use the unique identifier baseline for a group when processing redundantly-averaged data, in the `.where_inpainted.` file. 
* trim the where-inpainted files in the same way as the calfiles and datafiles (removing empty nights)
* 🤦🏻 write out the MAD data to the MAD file instead of the STD file...


jsdillon (approved):
Looks alright to me, assuming tests pass!

PR #915, opened 2023-10-20
improve handling for future pyuvdata

bhazelton:
This is a small tweak to ensure that hera_cal tests pass on the branch we're using to prepare for version 3.0. Since we haven't reached v3.0 yet the desired code branch isn't being used. This fixes the code to look at the actual function signature rather than the pyuvdata version so the correct branch is used.

The coverage issue is because CI is only running against current pyuvdata versions.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -9,7 +9,7 @@ import operator
 import os
 import copy
 import warnings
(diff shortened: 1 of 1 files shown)


jsdillon (approved):
Looks good to me!

PR #917, opened 2023-11-06
Allow user to impose a u-cut when generating PSWF filters for nucal

tyler-a-cox:
This PR allows users to set maximum and minimum u-bounds when generating PSWF basis functions for a nucal fit. This can be particularly useful when the user is attempting to fit a spectrally redundant group of baselines, where one of the baselines is significantly longer than the others. This can drastically increase the number of parameters without improving the fit of the other baselines. By restricting the u-axis, we can reduce the number of parameters involved in the fit while still allowing longer baselines to contribute to constraining spectral redundancy for the group.


jsdillon (approved):
Looks good to me!

PR #921, opened 2023-11-29
Fix inpaint nonfiring

steven-murray:
I think there were two things wrong here. So the way the code works is that it makes `data` and `nsamples` masked arrays. The masks are inpaint-aware flags (if inpaint mode is one), but the `flags` array is just the raw flags. After setting these masks, I was doing the flag thresholding, which set some new flags based on thresholds over the days for the old flags, but then I reset the masks to these new updated flags, which *include* all the raw flags. So I've fixed that.

Another thing that was happening is that the `std` was using the inpaint-aware flags, but it should be using the raw flags (so a completely flagged (but inpainted) integration should have a mean value, but std=inf. 

I think the code already naturally handles the case where all the days are flagged (for a particular lst/freq/pol) in inpainted mode by giving it a non-nan average, so the discrepancy there was probably due to the other bugs.

### DIFF
--- a/hera_cal/lstbin_simple.py
+++ b/hera_cal/lstbin_simple.py
@@ -429,6 +429,7 @@ def get_masked_data(
         inpainted = flags.copy() * (~allf)
 
     flags = flags | np.isnan(data) | np.isinf(data) | (nsamples == 0)
+    logger.info(f"In inpainted_mode: {inpainted_mode}. Got {np.sum(inpainted)} inpainted samples, {np.sum(flags)} total flags, {np.sum(flags & ~inpainted)} non-inpainted flags.")
     data = np.ma.masked_array(data, mask=(flags & ~inpainted))
     return data, flags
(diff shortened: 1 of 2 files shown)


hera_cal/lstbin_simple.py
```diff
@@ -580,10 +616,18 @@ def lst_average(
         lstbin_flagged[ndays_binned < sigma_clip_min_N] = True
 
     normalizable = norm > 0
+
     meandata[normalizable] /= norm[normalizable]
     # Multiply by nan instead of just setting as nan, so both real and imag parts are nan
     meandata[~normalizable] *= np.nan
 
+    # While the previous nsamples is different for in-painted and flagged mode, whih is
```
jsdillon, line 624: whih --> typo

jsdillon (approved):
Looks good! Just one tiny docstring typo.

PR #923, opened 2023-12-05
Omnical memory improvements

jsdillon:
Despite some misleading results from %memit, it appears that omnical's memory usage is the root of some of the problems we've been having with RTP. 

This PR improves the memory efficiency of omnical. It does so by studiously avoiding copying data unnecessarily, as well as by updating dictionaries with for loops rather than dictionary comprehension, so that we don't have to make a whole new dictionary before deallocating the old one.

It also changes the default value of `wgt_func` from `lambda x: 1.` to `None` and adds special treatment of the `None` case so that it doesn't multiply by 1 and create new arrays.

It has a negligible effect on performance (and actually might be a tiny bit faster). In my tests on the file_calibration notebook with  `zen.2460280.48839.sum.uvh5` locally, runtime improved from 5.07 minutes to 4.63 minutes. It improves peak memory usage from 18.8 GB to 12.3 GB, although it should be noted that ~6.5 GB was the memory usage (mostly raw data) going into omnical.

Here's my memray results before:

image

And after:

image

It's possible we can do even better if we use single precision, which I'll explore next.

This PR relies on https://github.com/HERA-Team/linsolve/pull/51

### DIFF
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -845,7 +845,7 @@ class OmnicalSolver(linsolve.LinProductSolver):
         return {k: eval(k, _sol) for k in keys}
 
     def solve_iteratively(self, conv_crit=1e-10, maxiter=50, check_every=4, check_after=1,
-                          wgt_func=lambda x: 1., verbose=False):
+                          wgt_func=None, verbose=False):
         """Repeatedly solves and updates solution until convergence or maxiter is reached.
         Returns a meta-data about the solution and the solution itself.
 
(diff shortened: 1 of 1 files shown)


hera_cal/redcal.py
```diff
@@ -832,20 +832,25 @@ def __init__(self, data, sol0, wgts={}, gain=.3, **kwargs):
             **kwargs: keyword arguments of constants (python variables in keys of data that
                 are not to be solved for) which are passed to linsolve.LinProductSolver.
         """
-        linsolve.LinProductSolver.__init__(self, data, sol0, wgts=wgts, **kwargs)
+        linsolve.LinProductSolver.__init__(self, data, sol0, wgts=wgts, build_solver=False, **kwargs)
         self.gain = np.float32(gain)  # float32 to avoid accidentally promoting data to doubles.
 
-    def _get_ans0(self, sol, keys=None):
+    def _get_ans0(self, sol, keys=None, to_update_inplace=None):
```
tyler-a-cox, line 838: It might be useful just to quickly say what input might be added to the `to_update_inplace` variable. It is a helper function, though, and I don't think the usage is too confusing, but I had to double-check the `solve_iteratively` function to made sure I understood how you were using it

jsdillon: Done.

tyler-a-cox (approved):
Looks great! Thanks for the fixes, @jsdillon!

PR #924, opened 2023-12-06
Make error catching work with updated pyuvdata errors

bhazelton:
This is a small PR to fix some error handling code so that it will continue to function in the same way with a future pyuvdata release.

The code in question is identifying specific pyuvdata errors by their messages, which will be updated in a future release to make them more clear.


jsdillon (approved):
Looks good to me!

PR #925, opened 2023-12-06
fix: use uint dtype for ant numbers to comply with pyuvdata

steven-murray:
This applies the patch suggested by @mkolopanis [here](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1357#issuecomment-1830436169).

Basically it allow the antenna numbers to be `np.uint` type instead of basic `int`.


steven-murray:
Leaving this as draft in case we wanna do the "right" thing later.

jsdillon:
I'm not convinced that this is the "right" thing... I much prefer the other solution.

PR #926, opened 2023-12-06
fix: enforce that ant numbers are uint64

steven-murray:
Fixes writing out uvdta objects according to this PR: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1357#issuecomment-1830436169.


jsdillon (changes requested):
Do we want to raise an error if doing this casting changes the value? (i.e. if the antpairs are actually non-intergers and/or negative?)

jsdillon (approved):

PR #929, opened 2024-01-29
Apply flag to N-sample before delay filtering

Kai-FengChen:
Add an option in `load_delay_filter_and_write` so that we can add the flags to the N-sample array before delay filtering (i.e., setting `nsample[flag == True] = 0`). Write a new test to support this new option. Also modify the run script.

### DIFF
--- a/hera_cal/delay_filter.py
+++ b/hera_cal/delay_filter.py
@@ -20,6 +20,7 @@ import logging
 
 logger = logging.getLogger(__name__)
 
+
 class DelayFilter(VisClean):
     """
     DelayFilter object.
@@ -98,7 +99,8 @@ def load_delay_filter_and_write(datafile_list, baseline_list=None, calfile_list=
                                 res_outfilename=None, CLEAN_outfilename=None, filled_outfilename=None,
                                 clobber=False, add_to_history='', polarizations=None,
                                 skip_flagged_edges=False, overwrite_flags=False,
-                                flag_yaml=None, read_axis=None, **filter_kwargs):
+                                flag_yaml=None, read_axis=None, apply_flag_to_nsample=False,
+                                **filter_kwargs):
     '''
     Uses partial data loading and writing to perform delay filtering.
     While this function reads from multiple files (in datafile_list)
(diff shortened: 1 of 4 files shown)


hera_cal/delay_filter.py
```diff
@@ -194,19 +198,27 @@ def load_delay_filter_and_write(datafile_list, baseline_list=None, calfile_list=
...
                 logger.info("  Factorizing flags")
                 df.factorize_flags(time_thresh=time_thresh, inplace=True)
+            if apply_flag_to_nsample:
+                logger.info("  Applying flags to nsample arrays")
+                for bl in baseline_list[i:i + Nbls_per_load]:
+                    _nsample = np.copy(df.hd.get_nsamples(bl))
+                    _nsample[df.hd.get_flags(bl)] = 0
+                    if _nsample.ndim == 2:
+                        _nsample = _nsample[:, :, np.newaxis]
+                    df.hd.set_nsamples(_nsample, bl)
```
jsdillon, line 208: This modifies self.hd but not self.nsamples (see https://github.com/HERA-Team/hera_cal/blob/d45df271d6fc16519e22b2a78c89c2cf92e2b367/hera_cal/vis_clean.py#L481)

Are you sure that's what you want to do?

(Unrelated, this is one of my big issues with vis_clean and its related classes... it attaches data is multiple ways that are had to keep track of. It's why I've generally been avoiding vis_clean and related classes for H6C and later.)

Kai-FengChen: That's a very good question, thanks for raising this. 

It seems to be more complicated than I thought. Here, when writing to file we set `partial_write=Nbls_per_load < len(baseline_list)`. If `partial_write` is `True`, we write from `self.hd`, otherwise we construct a new data container using `self.nsamples`. I am not exactly sure the reasoning behind these two modes of writing, maybe I can add a condition so that I modify `self.hd.nsamples` vs `self.nsamples` depending on the `partial_write` value we will be using in the end?

jsdillon: Maybe modify both?

scripts/delay_filter_run.py
```diff
@@ -45,7 +45,7 @@
     cache_dir=ap.cache_dir, res_outfilename=ap.res_outfilename,
     clobber=ap.clobber, write_cache=ap.write_cache, external_flags=ap.external_flags,
     read_cache=ap.read_cache, mode=ap.mode, overwrite_flags=ap.overwrite_flags,
-    factorize_flags=ap.factorize_flags, time_thresh=ap.time_thresh,
+    factorize_flags=ap.factorize_flags, apply_flag_to_nsample=ap.apply_flag_to_nsample, time_thresh=ap.time_thresh,
```
jsdillon, line 48: I don't think you actually added this to the argparser

jsdillon (changes requested):
Looks mostly good. Just a couple of questions

scripts/delay_filter_run.py
Kai-FengChen: Thanks! Should be fixed now!

jsdillon (approved):

PR #935, opened 2024-02-08
Lstbin subpackage

steven-murray:
This moves the new lst-binning functionality to a sub-package, called `lst_stack`. Suggestions on the best name are welcome.
I've kept the old `lstbin.py` module in place but added a deprecation warning, and *removed tests for it*. It should be removed after H4C is done and dusted.

#927 should be merged first.


hera_cal/__init__.py
```diff
@@ -24,7 +24,6 @@
 from . import io
 from . import delay_filter
 from . import abscal
-from . import lstbin
```
jsdillon, line 27: Do we want to add lst_stack?

hera_cal/lst_stack/stats.py
```diff
@@ -0,0 +1,282 @@
+"""Statistics for LST-binned data."""
+from __future__ import annotations
```
jsdillon: Why is so much in this submodule not covered by tests?

jsdillon (approved):
Thanks for organizing this, @steven-murray! This PR looks OK to me... just a couple of questions but nothing stopping approval

PR #936, opened 2024-02-08
Add ruff formatting

steven-murray:
Here I simply ran `ruff fix .` on the codebase (ruff is a linter and formatter that is super-fast and rolls flake8 and black and some other tools into one nice package). 

@jsdillon I know you have some aversion to some of the `black` styling, but I thought it would be useful to run once on everything and give you a chance to have a quick look through to see if there's anything super objectionable here. If there is, we can always either:

1. Switch off auto-formatting for that bit of code and keep automatic `ruff` formatting into the future.
2. Keep the changes that ruff made here, but manually change those bits back, and not keep ruff as an automatic fixer in the future (so use this as a once-off style upgrade). 

Note that many of the changes made here are unused imports etc. *which can be problems*.


mkolopanis:
regarding the structure thing you can usually do something like `# fmt: off` and later a `# fmt: on` to toggle auto-formatting. or skip for one liners
you can see more here https://docs.astral.sh/ruff/formatter/#format-suppression

mkolopanis:
also you can port all your flake8 configurations over to a pyproject.toml for ruff too https://docs.astral.sh/ruff/configuration/#configuring-ruff

jsdillon:
I'm fine adding ruff, but I'd really like to not modify the things currently ignored by https://github.com/HERA-Team/hera_cal/blob/main/.flake8

My biggest issue is with how hard to works to keep individual lines short. You end up with stuff like this:
image
I personally find the new version substantially harder to read.

hera_cal/abscal.py
```diff
@@ -48,9 +49,20 @@
...
+    split_bl,
+    join_bl,
+    join_pol,
+)
+
+PHASE_SLOPE_SOLVERS = [
+    "linfit",
+    "dft",
+    "ndim_fft",
+]  # list of valid solvers for global_phase_slope_logcal
```
steven-murray, line 65: This re-lining can be fixed by moving the comment

hera_cal/abscal.py
```diff
@@ -112,8 +124,15 @@ def abs_amp_logcal(model, data, wgts=None, verbose=True, return_gains=False, gai
...
-                                                     i, split_pol(k[-1])[1])) for i, k in enumerate(keys)])
+    eqns = odict(
+        [
+            (
+                k,
+                "a{}*eta_{}+a{}*eta_{}".format(i, split_pol(k[-1])[0], i, split_pol(k[-1])[1]),
+            )
+            for i, k in enumerate(keys)
+        ]
+    )
```
steven-murray, line 135: could be fixed by using a normal dict (which is equivalent to odict now anyway)

hera_cal/abscal.py
```diff
@@ -480,8 +522,17 @@ def amp_logcal(model, data, wgts=None, verbose=True):
...
+        [
+            (
+                k,
+                "eta_{}_{} + eta_{}_{}".format(
+                    k[0], split_pol(k[-1])[0], k[1], split_pol(k[-1])[1]
+                ),
+            )
+            for i, k in enumerate(keys)
+        ]
+    )
```
steven-murray, line 535: again use normal dict

hera_cal/abscal.py
```diff
@@ -549,8 +600,15 @@ def phs_logcal(model, data, wgts=None, refant=None, verbose=True):
...
-                                                     k[1], split_pol(k[2])[1])) for i, k in enumerate(keys)])
+    eqns = odict(
+        [
+            (
+                k,
+                "phi_{}_{} - phi_{}_{}".format(k[0], split_pol(k[2])[0], k[1], split_pol(k[2])[1]),
+            )
+            for i, k in enumerate(keys)
+        ]
+    )
```
steven-murray, line 611: and again

hera_cal/abscal.py
```diff
@@ -681,8 +760,15 @@ def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, f0=0., solv
...
-                                                     k[1], split_pol(k[2])[1])) for i, k in enumerate(keys)])
+    eqns = odict(
+        [
+            (
+                k,
+                "tau_{}_{} - tau_{}_{}".format(k[0], split_pol(k[2])[0], k[1], split_pol(k[2])[1]),
+            )
+            for i, k in enumerate(keys)
+        ]
+    )
```
steven-murray, line 771: and again

hera_cal/abscal.py
```diff
@@ -711,14 +797,21 @@ def delay_lincal(model, data, wgts=None, refant=None, df=9.765625e4, f0=0., solv
...
-                                                     k[1], split_pol(k[2])[1])) for i, k in enumerate(keys)])
+    eqns = odict(
+        [
+            (
+                k,
+                "phi_{}_{} - phi_{}_{}".format(k[0], split_pol(k[2])[0], k[1], split_pol(k[2])[1]),
+            )
+            for i, k in enumerate(keys)
+        ]
+    )
```
steven-murray, line 808: and again

hera_cal/abscal.py
```diff
@@ -931,36 +1061,61 @@ def RFI_delay_slope_cal(reds, antpos, red_data, freqs, rfi_chans, rfi_headings,
...
+            np.exp(
+                -2j
+                * np.pi
+                * np.dot(unique_blvecs[red[0]], rfi_headings)
+                * freqs[rfi_chans]
+                / constants.c
+            )
+            for red in reds
+        ]
+    )
```
steven-murray, line 1092: I think this is more reasonable, but @jsdillon might disagree

hera_cal/abscal.py
```diff
@@ -931,36 +1061,61 @@ def RFI_delay_slope_cal(reds, antpos, red_data, freqs, rfi_chans, rfi_headings,
...
+            [
+                f"T_{red[0][2]}_{dim} * {ibl_comp}"
+                for dim, ibl_comp in enumerate(idealized_blvecs[red[0]])
+            ]
+        )
         ls_data[eq_str] = delay_sols[red[0]]
-        ls_wgts[eq_str] = len(red) * np.ones_like(delay_sols[red[0]])  # weight by number of baselines in group
+        ls_wgts[eq_str] = len(red) * np.ones_like(
+            delay_sols[red[0]]
+        )  # weight by number of baselines in group
```
steven-murray, line 1118: can be fixed by moving comment

hera_cal/abscal.py
```diff
@@ -1207,43 +1390,61 @@ def global_phase_slope_logcal(model, data, antpos, reds=None, solver='linfit', w
...
         bl_vecs[eqn_str] = antpos[rk[0]] - antpos[rk[1]]
         pols[eqn_str] = rk[2]
 
         # calculate median of unflagged angle(data/model)
         # ls_weights are sum of non-binary weights
         dm_ratio = avg_data[rk] / avg_model[rk]
-        dm_ratio /= np.abs(dm_ratio)  # This gives all channels roughly equal weight, moderating the effect of RFI (as in firstcal)
+        dm_ratio /= np.abs(
+            dm_ratio
+        )  # This gives all channels roughly equal weight, moderating the effect of RFI (as in firstcal)
```
steven-murray, line 1407: move the comment

hera_cal/abscal.py
```diff
@@ -1544,21 +1769,37 @@ def _put_transformed_array_on_integer_grid(transformed_antpos, tol=1e-8, max_num
...
             for numerator in range(2, max_numerator + 1):
-                if np.max(np.abs(np.rint(this_dim_positions * numerator) - this_dim_positions * numerator)) < tol:
+                if (
+                    np.max(
+                        np.abs(
+                            np.rint(this_dim_positions * numerator) - this_dim_positions * numerator
+                        )
+                    )
+                    < tol
+                ):
```
steven-murray, line 1781: this is another spacious re-lining. probably nicer if we save the interior thing as a variable on its own line.

hera_cal/abscal.py
```diff
@@ -1544,21 +1769,37 @@ def _put_transformed_array_on_integer_grid(transformed_antpos, tol=1e-8, max_num
...
-                if np.max(np.abs(np.rint(this_dim_positions * numerator / denominator) - this_dim_positions * numerator / denominator)) < tol:
+                if (
+                    np.max(
+                        np.abs(
+                            np.rint(this_dim_positions * numerator / denominator)
+                            - this_dim_positions * numerator / denominator
+                        )
+                    )
+                    < tol
+                ):
```
steven-murray, line 1797: and for this one

hera_cal/abscal.py
```diff
@@ -1904,16 +2165,14 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
 
     # raise warning on flags
     if flags is not None and medfilt_flagged is False:
-        print("Warning: flags are fed, but medfilt_flagged=False. \n"
-              "This may cause weird behavior of interpolated points near flagged data.")
+        print(
+            "Warning: flags are fed, but medfilt_flagged=False. \n"
+            "This may cause weird behavior of interpolated points near flagged data."
+        )
```
steven-murray, line 2171: this should be an actual warning...

hera_cal/abscal.py
```diff
@@ -1990,13 +2249,33 @@ def interp2d_vis(model, model_lsts, model_freqs, data_lsts, data_freqs, flags=No
...
+        )(data_freqs, data_lsts)
+        interp_imag = interpolate.interp2d(
+            model_freqs,
+            model_lsts,
+            imag,
+            kind=kind,
+            copy=False,
+            bounds_error=False,
+            fill_value=fill_value,
+        )(data_freqs, data_lsts)
```
steven-murray, line 2269: this might be cleaner with a partial

hera_cal/abscal.py
```diff
@@ -2621,25 +2948,57 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
+            (
+                model,
+                model_flags,
+                model_antpos,
+                model_ants,
+                model_freqs,
+                model_lsts,
+                model_times,
+                model_pols,
+            ) = io.load_vis(model, pop_autos=True, return_meta=True, filetype=filetype)
```
steven-murray, line 2966: this is more readable with the transformation, IMO, but could be better again

hera_cal/abscal.py
```diff
@@ -2621,25 +2948,57 @@ def __init__(self, model, data, refant=None, wgts=None, antpos=None, freqs=None,
...
+            (
+                data,
+                flags,
+                data_antpos,
+                data_ants,
+                data_freqs,
+                data_lsts,
+                data_times,
+                data_pols,
+            ) = io.load_vis(data, pop_autos=True, return_meta=True, filetype=filetype)
```
steven-murray, line 2984: and this

hera_cal/abscal.py
```diff
@@ -2783,21 +3156,31 @@ def phs_logcal(self, avg=False, verbose=True):
...
-                        + 1j * np.median(np.imag(np.exp(1j * self._ant_phi[k])))
+            self._ant_phi = odict(
+                [
+                    (
+                        k,
+                        np.ones_like(self._ant_phi[k])
+                        * np.angle(
+                            np.median(np.real(np.exp(1j * self._ant_phi[k])))
+                            + 1j * np.median(np.imag(np.exp(1j * self._ant_phi[k])))
+                        ),
```
steven-murray, line 3176: this needs a re-structure

jsdillon:
While I agree in general that all the odicts should be replaced with regular dicts, that's a separate PR. If you want to do that separately, I'll review it. Likewise, if you want to replace prints with actual warnings, that's yet another PR that I'd be in favor of as a separate thing. I'm really strongly against giant linting PRs that also have substantive code changes. They are terrible to review.

I really don't want to get into a discussion of whether the code is more or less readable in each particular instance. Some are improvements, some are steps backward. There's 33k lines of changes here... it's just not a good use of our time. I picked one instance from abscal because it comes at the front of the alphabet, not because I'm focusing on that particular module.

I'm fine running a linter to make a minimal set of changes so that flake8 (as defined in https://github.com/HERA-Team/hera_cal/blob/main/.flake8) passes. I'd really prefer not to go further than that at this time.

steven-murray:
@jsdillon I understand the hesitancy. On the other hand it takes _me_ time every time I write a new PR to manually fix the formatting, and furthermore the style in each module is inconsistent, which makes it hard to read. I can run the fixer with minimal settings as you suggest (and not run the formatter), but I think it's shooting ourselves in the foot long term. If instead we just bite the bullet and do one big PR that sets the style forever, we never have to worry about it again. But I'm not going to make that call. I'll rewind this PR and when I can I'll do another one with more minimal fixes.

jsdillon:
I get that frustration. Can we integrate ruff (or similar) to always perform the minimal fix to new code so that it passes the style checks? That doesn't fix the issue of style being inconsistent between modules (which doesn't bother me and will still be the case in terms of function choices and variable names that reflect the original authors), but it does mean you won't have to manually fix any formatting, right?

steven-murray:
Yeah, I will try that. Perhaps I'll try it once the few fresh PRs are merged

PR #938, opened 2024-02-19
Fix red groups

steven-murray:
This fixes a subtle bug in defining redundant groups.

The bug was manifesting when using the `keyed_on_bls()` method. This method is meant to take a particular set of redundant baselines, which have a particular unique key, and re-key so that the key must be one of a given set of baselines. However, when the given set of baselines contains the _reverse_ of the given key, this was failing. 

Along with fixing this problem, this adds the `.get_bls_in_set()` method, which is a more intuitive way of doing the same operation (but is not cached for re-use as the re-keying is).


hera_cal/red_groups.py
```diff
@@ -565,18 +584,42 @@ def keyed_on_bls(self, bls: Sequence[BlLike], inplace: bool = False) -> Redundan
...
 
     def __call__(self, red: Sequence[BlLike]) -> BlLike:
         filtered_red = [bl for bl in red if bl in self.bls]
-        return self.chooser(filtered_red) if filtered_red else self.chooser(red)
-    
-    
\ No newline at end of file
+
+        if not filtered_red:
+            raise NonExistentBaselineError("No baselines in the group are in the provided set.")
```
jsdillon, line 623: This is causing `hera_cal/tests/test_datacontainer.py::test_RedDataContainerKeyManipulation` to fail. Do we need to raise an error here rather than just do the previous default behavior?

hera_cal/tests/test_red_groups.py
```diff
@@ -235,26 +236,26 @@ def test_keyed_on_bls(self):
...
+        print(new._red_key_to_bls_map)
+        assert (0, 2) in new
+        assert new.get_ubl_key((0, 2)) == (0, 3)
+        assert new.get_ubl_key((0, 1)) == (0, 1)
 
         new.keyed_on_bls(bls=[(0, 2)], inplace=True)
-        print(new._red_list)
-        assert new.get_ubl_key((0,2)) == (0,2)
+        print(new._red_key_to_bls_map)
+        print(new._bl_to_red_map)
```
jsdillon, line 247: Is there a reason we have all these print statements? Are they helpful?

jsdillon (changes requested):
Thanks for tracking down this very subtle bug, @steven-murray. Just a few comments and one question about how to handle a new test failure on `RedDataContainer`

hera_cal/lst_stack/binning.py
```diff
@@ -486,8 +486,8 @@ def lst_bin_files_for_baselines(
         else:
             inpainted = None
 
-        if redundantly_averaged:
-            keyed = reds.keyed_on_bls(_data.antpairs())
+        # if redundantly_averaged:
+        #     keyed = reds.keyed_on_bls(_data.antpairs())
```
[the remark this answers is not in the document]
steven-murray: Nope, I removed it.

hera_cal/tests/test_red_groups.py
steven-murray: Nope. Cruft. Removed.

jsdillon (approved):
One minor error, I think, in a docstring example, but otherwise I'm happy with this!

hera_cal/red_groups.py
```diff
@@ -590,16 +594,67 @@ def get_reds_in_bl_set(
...
+            include_conj_only_if_missing to True.
+
+        Returns
+        -------
+        set
+            A set of baselines redundant with bl that are in bl_set.
+
+        Examples
+        --------
+        >>> rg = RedundantGroups.from_antpos([[(0, 1), (1, 2)], [(0, 2)]])
```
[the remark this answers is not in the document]
steven-murray: Ah yeah... fixed.

PR #942, opened 2024-02-27
Count phase flips in smooth_cal

jsdillon:
When an antenna experiences a spontaneous phase flip, we now print out how many times that flip happens, not just the number of flips. 

Also some pep8 stuff.


bhazelton (approved):
Looks good, thanks!

PR #943, opened 2024-03-06
Update setup.py with python version requirement

adeliegorce:
Changed python requirement to `python>3.8` to ensure compatibility with `optax` (which dropped support for earlier versions).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup_args = {
                 'scripts/flag_all.py', 'scripts/throw_away_flagged_antennas.py', 'scripts/select_spw_ranges.py',
                 'scripts/multiply_gains.py', 'scripts/lstbin_simple.py', 'scripts/subselect.py'],
     'package_data': {'hera_cal': data_files},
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,6 +46,7 @@ def package_files(package_dir, subdirectory):
                 'scripts/flag_all.py', 'scripts/throw_away_flagged_antennas.py', 'scripts/select_spw_ranges.py',
                 'scripts/multiply_gains.py', 'scripts/lstbin_simple.py', 'scripts/subselect.py'],
     'package_data': {'hera_cal': data_files},
+    'python_requires': '>3.8',
```
jsdillon, line 49: shouldn't it be >=3.9? Otherwise, I think python 3.8.10 or whatever would qualify

jsdillon (changes requested):

setup.py
steven-murray: Yes, I think >=3.9 is the way to go

jsdillon (approved):

PR #949, opened 2024-06-03
Enable advanced users to provide transformed_antpos to complex_phase_abscal

jsdillon:
This would be instead of inferring them from the reds. It's useful for @tyler-a-cox's LST-cal.


tyler-a-cox (approved):
This looks good to me! Thanks, @jsdillon

PR #952, opened 2024-06-13
Remove an unused import that will break in pyuvdata v3

bhazelton:
This import isn't used, so there should be no affect to the code. Generally, it'd be a good idea to remove unused imports. There are linters that can help you find those.


steven-murray (approved):
Thanks @bhazelton. Hopefully things like this will be gone soon when I add some ruff checks.

PR #953, opened 2024-06-19
Add LST-binned absolute calibration to `lst_stack`

tyler-a-cox:


hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+            of the antennas in the redundant baseline group.
+        inpaint_bands : tuple of slices, default=(slice(0, None, None),)
+            Defines the frequency bands to use for smoothing. Each slice object in the tuple
+            defines a separate frequency band to use for smoothing. The default is to smooth over
+            all frequencies at once.
+        auto_stack : LSTStack, default=None
+            An LSTStack object containing the stack of auto-correlations. If provided, the
+            auto-correlations will be calibrated using the same gains calculated from the
+            cross-correlations. Note that the auto-correlations are not used to calculate the
+            gains if provided.
```
jsdillon, line 56: I think we want to use these in amplitude calibration, at least by default

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+                    if smooth_gains:
+                        basis = smoothing_functions[bandidx]
+                        bl_gain_here = np.array(
+                            [
+                                np.dot(basis, _fmat.dot(_raw_gain))
+                                for _fmat, _raw_gain in zip(
+                                    fmats[pol][bandidx], raw_gain
+                                )
+                            ]
+                        )
```
jsdillon, line 351: First off, I think we should rephase before we smooth (see smooth_cal for how to do that). Second, I'm wondering if we should smooth antenna gains rather than gain products.

jsdillon (changes requested):
Looking pretty good. A handful of feature requests / method changes and removing some hardcoding.

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+            containing the antenna numbers and polarization of the baseline. If return_gains is
+            set to False, this dictionary will be empty.
+    """
+    # Assert some calibration done
+    assert (
+        run_amplitude_cal or run_phase_cal
+    ), "At least one calibration mode must be used"
+    assert (
+        stack.data.shape[1:] == model.shape
+    ), "Model must have the same number of antpairs/freqs/pols as stack.data"
```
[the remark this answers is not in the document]
steven-murray: Yeah, it is better practice. It's not that it will really make a difference here, but `assert` is meant to be used where you expect something to _never fail_ (not dependent on user error etc.). Assertions are possible to _ignore_ when compiling python with a certain flag for optimization, so they're not recommended for catching when users pass bad parameters etc.

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+    run_phase_cal: bool = True,
+    smoothing_scale: float = 10e6,
+    eigenval_cutoff: float = 1e-12,
+    calibrate_inplace: bool = True,
+    return_gains: bool = True,
+    smooth_gains: bool = True,
+):
+    """
+    This function performs calibration on LSTStack object by comparing each day to an input
+    reference model. Each day is calibrated independently using only the absolute calibration
```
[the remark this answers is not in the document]
tyler-a-cox: Changed this

hera_cal/lst_stack/calibration.py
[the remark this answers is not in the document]
tyler-a-cox: Changed these to ValueErrors

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+    if run_amplitude_cal and not all_nights_flagged:
+        # Store gain solutions in paramter dictionary
+        solution = abscal.abs_amp_lincal(
+            model=abscal_model, data=data_here, wgts=wgts_here, verbose=False
+        )
+
+        for pol in pols:
+            # Calibration parameters store in an N_nights by N_freqs array
+            amplitude_gain = np.where(
+                np.isfinite(solution[f"A_J{pol}"]), solution[f"A_J{pol}"], 1.0 + 0.0j
```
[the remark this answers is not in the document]
tyler-a-cox: I think I've addressed this correctly, but let me know if I should be doing this a better way!

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+            amplitude_gain = np.where(
+                np.isclose(amplitude_gain, 0.0), 1.0 + 0.0j, amplitude_gain
+            )
```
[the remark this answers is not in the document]
tyler-a-cox: Removed this

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+            calibration_parameters[f"A_J{pol}"] = amplitude_gain
+    else:
+        # Fill in amplitude w/ ones if no running amplitude calibration
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+        for pol in pols:
+            calibration_parameters[f"A_J{pol}"] = np.ones(
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+    if run_phase_cal and not all_nights_flagged:
+        gain_ants = set()
+        for ap in antpairs:
+            gain_ants.update(ap)
+        gain_ants = list(gain_ants)
+
+        transformed_antpos = None
+
+        # Gains for the phase component
+        phase_gains = {(ant, "J" + pol): [] for ant in gain_ants for pol in pols}
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+
+        # Add tip-tilt gain parameters to calibration_parameters
+        calibration_parameters.update({f"T_J{pol}": [] for pol in pols})
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed

hera_cal/lst_stack/calibration.py
```diff
@@ -0,0 +1,371 @@
...
+            for polidx, pol in enumerate(pols):
+                cal_bls = []
+
+                _data_here = {}
+                _abscal_model = {}
+
+                if np.all(stack.flags[nightidx, :, :, polidx]):
+                    # Store phase gains
+                    for ant in gain_ants:
+                        phase_gains[(ant, f"J{pol}")].append(
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed

hera_cal/lst_stack/calibration.py
tyler-a-cox: I added in rephasing and smoothing antenna gains rather than the antenna gain product

tyler-a-cox:
I made a few changes here, including refactoring the main calibration function into smaller modules to improve readability, fixing the polarization issues so that the J-strings are no longer hard-coded, and so the function can handle cross-polarized visibilities, added rephasing of antenna gains prior to smoothing, and smoothing antenna gains rather than antenna gain products.

hera_cal/lst_stack/calibration.py
```diff
@@ -132,240 +174,350 @@ def lstbin_absolute_calibration(
...
+                auto_gain = gains[antpol1] * gains[antpol2].conj()
+                auto_gain = np.where(
+                    stack.flags[:, 0, :, polidx], 1.0 + 0.0j, auto_gain
+                )
 
-                # Store gains to return
-                if return_gains:
-                    gains[(ant1, ant2, pol)] = bl_gain
+                # Calibrate out gains
+                auto_stack.data[:, apidx, :, polidx] /= auto_gain
```
jsdillon, line 521: This feels like unnecessarily duplicated code.

hera_cal/lst_stack/calibration.py
```diff
@@ -132,240 +174,350 @@ def lstbin_absolute_calibration(
...
+        eigenval_cutoff=eigenval_cutoff,
+    )
+
+    # Iterate for each baseline in the stack and calibrate out gains
+    if calibrate_inplace:
+        for polidx, pol in enumerate(stack.pols):
+            for apidx, (ant1, ant2) in enumerate(stack.antpairs):
+                antpol1, antpol2 = utils.split_bl((ant1, ant2, pol))
+
+                # Compute gain and remove nans/infs
```
jsdillon, line 501: I would hope that all nans/infs are flagged, but is it true?

hera_cal/lst_stack/calibration.py
```diff
@@ -132,240 +174,350 @@ def lstbin_absolute_calibration(
...
-                    )
-                    bl_gain[:, band] = bl_gain_here
+    Parameters:
+    ----------
+        stack : LSTStack
+            The LSTStack object to calibrate
+        model : np.ndarray
+            The reference model to calibrate the data to. The model should have the same
+            number of baselines, frequencies, and polarizations as the data in stack. The model
+            can simply be the mean of the data in the stack if data are already abscal'd.
```
jsdillon, line 368: I'd clarify which axis you suggest taking the mean over. It's the 0th, right?

hera_cal/lst_stack/calibration.py
```diff
@@ -132,240 +174,350 @@ def lstbin_absolute_calibration(
...
+
+    # Get the unique antenna-polarizations
+    unique_pols = list(
+        set(sum(map(list, [utils.split_pol(pol) for pol in stack.pols]), []))
+    )
+
+    for pol in unique_pols:
+        # Calibration parameters store in an N_nights by N_freqs array
+        amplitude_gain = np.where(
+            np.isfinite(solution[f"A_{pol}"]), solution[f"A_{pol}"], 1.0 + 0.0j
```
jsdillon, line 212: Do we want to set this to 1.0 + 0.0j here, or rely on passing a nan/inf up to the higher level function where a decision can be made as to what to do with it? 

I'd be more comfortable if there was a check that all the weights are 0 rathert han if the check was just on whether `solution` were finite

jsdillon: Or maybe just use the flags directly?

hera_cal/lst_stack/calibration.py
```diff
@@ -132,240 +174,350 @@ def lstbin_absolute_calibration(
...
+            the auto-correlations will be used to calculate the gain amplitude if they are provided.
+            Note that the auto_model must also be provided if this flag is set to True and autos are
+            provided. If set to False, the auto-correlations will not be used for calibration.
+
+    Returns:
+    -------
+        calibration_parameters : dict
+            A dictionary containing the calibration parameters. The keys are as follows:
+                - 'A_J{pol}' : The amplitude calibration parameters for each polarization.
+                - 'T_J{pol}' : The tip-tilt calibration parameters for each polarization
```
jsdillon, line 410: I'd clarify the shape and type of these dictionary entires

jsdillon (changes requested):
Just a few minor comments, and then I think we'll be good to go!

hera_cal/lst_stack/calibration.py
tyler-a-cox: I think I'll just use the flags to be safe.

hera_cal/lst_stack/calibration.py
tyler-a-cox: Clarified this

hera_cal/lst_stack/calibration.py
tyler-a-cox: Good catch, I've clarified that here

hera_cal/lst_stack/calibration.py
tyler-a-cox: It should be true, I think I was slightly overdoing the removing nans/infs because they were bleeding into other channels when I was smoothing. If we're getting nans in places without flags, that 'sprobably indicative of other issues that we probably should just allow to fail so that we can fix them later. I removed it here.

jsdillon (approved):
Nice work! Thanks @tyler-a-cox !

PR #956, opened 2024-07-02
Basic ruff

steven-murray:
This PR swaps out out flake8 checks for the _same_ checks with `ruff` (to the extent possible). This is _faster_, and also allows them to be auto-fixed at pre-commit time. I am _not_ currently running `ruff format`, which would result in black-like formatting (we could decide to add that at a later time if we want), only doing the flake8 checks/fixes. I ran the checks on the entire code-base and made all necessary fixes. 

I also deleted the `calibrations/` sub-folder as it wasn't being imported or used anywhere in the code. It's possible that it's being used from outside the code in some pipeline, in which case I'll put it back in.


jsdillon (approved):
Looks good to me. Thanks @steven-murray !

PR #957, opened 2024-07-02
Add Ruff Rule A* (names shadowing builtins)

steven-murray:
This should only be merged after #956 

This adds rules that check for names shadowing python built-ins.


tyler-a-cox (approved):
Looks good, thanks @steven-murray!

PR #958, opened 2024-07-03
fix: lstcal with multiple bands

steven-murray:
Fixes issues found when running lststack notebook with new calibration / inpainting code in hera-cal.


tyler-a-cox (approved):
Looks good, thanks for the fix @steven-murray!

PR #960, opened 2024-07-16
Suppress mean of empty slice warnings in smooth_cal

jsdillon:
Nothing fancy here.


steven-murray (approved):
Thanks @jsdillon. Interestingly, I think these warnings randomly pop up because we're not always fixing the seed of the data-generation process. So we could get more in the future on other tests...

PR #961, opened 2024-07-22
make sure to copy attributes on subtraction

dannyjacobs:
Subtracting two datacontainers returned a new datacontainer shorn of all attributes. This fixes the problem by always returning the difference with all the attributes of the A, where the difference being computed is A - B.

### DIFF
--- a/hera_cal/datacontainer.py
+++ b/hera_cal/datacontainer.py
@@ -218,6 +218,7 @@ class DataContainer:
     def __add__(self, D):
         '''
         Addition operator overload.
+        returns a copy of this container.
 
         Add the values of this DataContainer with
         the value of D. If D is another DataContainer, add
@@ -255,6 +256,8 @@ class DataContainer:
         Subtraction operator overload.
 
         Subtract D with the values of this DataContainer.
+        returns a copy of this container.
+
         If D is another DataContainer, subtract
         their values and form a new container,
         otherwise subtract D from each ndarray in self.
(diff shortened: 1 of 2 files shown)


hera_cal/datacontainer.py
```diff
@@ -243,22 +243,21 @@ def __sub__(self, D):
...
             # iterate over D keys
             for i, k in enumerate(D.keys()):
                 if self.__contains__(k):
                     newD[k] = self.__getitem__(k) - D[k]
 
-            return DataContainer(newD)
+            return newD
 
         else:
             newD = copy.deepcopy(self)
```
jsdillon, line 263: This line is now redundant.

hera_cal/datacontainer.py
```diff
@@ -243,22 +243,21 @@ def __sub__(self, D):
         otherwise subtract D from each ndarray in self.
```
jsdillon, line 243: I'd edit the docstring to clarify that the original object is always duplicated and that the metadata of the original object is preserved (but not the new object).

jsdillon (changes requested):
Thanks @dannyjacobs!

If you're going to make this change, I'd suggest making parallel changes to `__add__` and `__mult__` and `__floordiv__` and `__truediv__`. I'd also like to see at least one unit test that exposes the issue you were having and is fixed by this change.

dannyjacobs:
Have made the following changes:
1. Change propagated to add, sub, mul, div
2. Docstring informs about returning a copy object
3. updated test so that it fails without this fix.
4. branch adder_fail adds just the test to the main branch. See main fail!

jsdillon (approved):
Assuming my change passes tests, I'm good to go. Thanks @dannyjacobs !

PR #964, opened 2024-07-25
Add cross-polarization relative phase calibration to LST-cal

tyler-a-cox:


hera_cal/abscal.py
```diff
@@ -992,6 +992,73 @@ def RFI_delay_slope_cal(reds, antpos, red_data, freqs, rfi_chans, rfi_headings,
...
+        If True, return gains instead of the relative phase difference between the two polarizations. 
+        Default is False.
+    gain_ants : list of tuples
+        List of antenna-pol tuples to return gains for. Default is None.
+
+    Returns:
+    --------
+    If return_gains is False:
+        delta : np.ndarray
+            Array of relative phase differences between the two polarizations.
```
jsdillon, line 1021: In radians, right?

hera_cal/abscal.py
```diff
@@ -992,6 +992,73 @@ def RFI_delay_slope_cal(reds, antpos, red_data, freqs, rfi_chans, rfi_headings,
...
+        model_pol1, model_pol2 = utils.split_pol(model_bl[-1])
+        if model_pol1 == model_pol2:
+            continue
+
+        # Skip if the polarizations are the same
+        data_pol1, data_pol2 = utils.split_pol(data_bl[-1])
+        if model_pol1 == model_pol2:
+            continue
+
+        if data_pol1 < data_pol2:
```
jsdillon, line 1041: I find the use of < for strings to be very confusing and could result in difference answers depending on whether you use xx/yy or ee/nn (depending on one's x_orientation) but seems really unsatisfying. If you want to convert these to pyuvdata polarization numbers, I'd be fine with that.

hera_cal/lst_stack/calibration.py
```diff
@@ -335,6 +341,51 @@ def _lstbin_phase_calibration(
     return calibration_parameters, phase_gains
 
 
+def _lstbin_cross_pol_phase_calibration(
+    stack: LSTStack,
+    model: np.ndarray,
+):
+    """
+    This function performs cross-polarization phase calibration on LSTStack object by comparing each day to an input
+    reference model.
```
jsdillon, line 350: I'd at least mention what the output is and what it's units are

hera_cal/lst_stack/calibration.py
```diff
@@ -335,6 +341,51 @@ def _lstbin_phase_calibration(
...
+            ).astype(float)
+            abscal_model[(ant1, ant2, pol)] = model[apidx, :, polidx] * np.ones(
+                (len(stack.nights), 1)
+            )
+            data[(ant1, ant2, pol)] = stack.data[:, apidx, :, polidx]
+            baselines.append((ant1, ant2, pol))
+
+    # Perform cross-polarization phase calibration
+    delta = abscal.cross_pol_phase_cal(
+        model=abscal_model, data=data, model_bls=baselines, data_bls=baselines, wgts=wgts
```
jsdillon, line 383: for clarity, perhaps explicitly set return_gains to False

hera_cal/abscal.py
jsdillon: Alternatively, you could explicitly have the user specify which is the reference polarization (and therefore, effectively, which isn't)

hera_cal/lst_stack/calibration.py
```diff
@@ -490,6 +545,38 @@ def lstbin_absolute_calibration(
...
+            # Get antennas
+            gain_ants = set()
+            for ap in stack.antpairs:
+                gain_ants.update(ap)
+
+            gain_ants = list(gain_ants)
+
+            for ant in gain_ants:
+                for pol in unique_pols:
+                    if pol != "Jee":
```
jsdillon, line 571: I don't like this hard-coded here. I'd either make this an explicit parameter (refpol) or perhaps have it follow the same logic as above about comparing pyuvdata polnums

hera_cal/lst_stack/tests/test_calibration.py
```diff
@@ -277,3 +285,47 @@ def test_value_errors(self):
...
+            ants=[0, 1, 2, 3],
+            time_axis_faster_than_bls=False,
+            pols=["xx", "yy", "xy", "yx"],
+            freqs=mockuvd.PHASEII_FREQS[:100],
+        )
+        stack = LSTStack(uvd)
+        stack.data[1:] = stack.data[0]  # All nights exactly the same
+
+        stack_copy = stack.copy()
+        delta = np.random.uniform(-1, 1, size=(20, 1, 1)) * np.ones((1, 10, 100))
```
jsdillon, line 303: you're going to need to seed this. The new way to do that is to create `rng = np.random.default_rng(seed=21)` (or whatever seed) and then `rng.uniform()`

hera_cal/lst_stack/tests/test_calibration.py
```diff
@@ -277,3 +285,47 @@ def test_value_errors(self):
...
+            all_reds=[],
+            run_amplitude_cal=False,
+            run_phase_cal=False,
+            run_cross_pol_phase_cal=True,
+            calibrate_inplace=True,
+            smooth_gains=False,
+        )
+
+        post_cal_std = np.std(stack_copy.data, axis=0)
+        assert np.all(post_cal_std[..., 2:] <= pre_cal_std[..., 2:])
```
jsdillon, line 331: Can we do a more stringent test here... not sure that the just reducing the std is good enough. Can we show that we actually got the right answer?

jsdillon (changes requested):
Looks pretty good @tyler-a-cox! Just a handful of suggested changes.

hera_cal/lst_stack/tests/test_calibration.py
tyler-a-cox: Now checking that the standard deviation across nights is close to zero for all bls/freqs/times

hera_cal/lst_stack/tests/test_calibration.py
tyler-a-cox: Changed this

hera_cal/lst_stack/calibration.py
tyler-a-cox: Changed this

hera_cal/lst_stack/calibration.py
tyler-a-cox: Fixed this

hera_cal/lst_stack/calibration.py
```diff
@@ -490,6 +545,38 @@ def lstbin_absolute_calibration(
...
 
+    if run_cross_pol_phase_cal and not all_nights_flagged:
+        for pol in stack.pols:
+            pol1, pol2 = utils.split_pol(pol)
+
+            if pol1 != pol2:
+                cross_pols_in_data = True
+                break
+        else:
+            cross_pols_in_data = False
```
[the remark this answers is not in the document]
tyler-a-cox: Fixed this

hera_cal/abscal.py
tyler-a-cox: Yeah, I'll make that more clear

hera_cal/abscal.py
tyler-a-cox: Settled on reference polarization

hera_cal/lst_stack/calibration.py
tyler-a-cox: Fixed this with a reference polarization

hera_cal/abscal.py
```diff
@@ -1013,12 +1013,14 @@ def cross_pol_phase_cal(model, data, model_bls, data_bls, wgts={}, return_gains=
         Default is False.
     gain_ants : list of tuples
         List of antenna-pol tuples to return gains for. Default is None.
+    refpol : str
+        Reference polarization to use for relative phase calibration. Default is 'Jee'.
```
jsdillon, line 1017: I'd be even more explicit: gains in the `refpol` are not change by relative polarization calibration.

jsdillon: Also, do we want to do any checking that one of the two pols is the refpol and that there are only two pols?

hera_cal/lst_stack/calibration.py
```diff
@@ -463,6 +465,9 @@ def lstbin_absolute_calibration(
             the auto-correlations will be used to calculate the gain amplitude if they are provided.
             Note that the auto_model must also be provided if this flag is set to True and autos are
             provided. If set to False, the auto-correlations will not be used for calibration.
+        refpol : str, default='Jee'
+            The reference polarization to use for cross-polarization phase calibration. This is
+            the polarization that the other polarizations are calibrated to. The default is 'Jee'.
```
jsdillon, line 470: "other polarization is" since there are only two

jsdillon (changes requested):
Just a couple documentation changes and data checks and I think we're good to go.

hera_cal/lst_stack/calibration.py
tyler-a-cox: Fixed

hera_cal/abscal.py
tyler-a-cox: Added these checks

jsdillon (approved):
If tests pass, I'm happy.

PR #966, opened 2024-07-25
Explicitly handle cross-polarized autos in HERAData.update()

jsdillon:


tyler-a-cox (approved):
This looks good to me, thanks @jsdillon

PR #968, opened 2024-07-30
Fix error when running LST-cal smoothing and a day is fully flagged

tyler-a-cox:
This PR fixes an error in LST-cal when `smooth_gains=True` and a full day is flagged. Currently, this smoothing code sets gains to zeros when the entire band is flagged. This update identifies nights where all baselines are flagged and replaces the smoothed gains with ones.

### DIFF
--- a/hera_cal/lst_stack/calibration.py
+++ b/hera_cal/lst_stack/calibration.py
@@ -100,16 +100,26 @@ def _expand_degeneracies_to_ant_gains(
                 )
                 fmats[split_pol1].append(fmat)
 
+    # Dictionary for storing gains
     gains = {}
-    for polidx, pol in enumerate(stack.pols):
-        split_pol1, split_pol2 = utils.split_pol(pol)
-        if split_pol1 != split_pol2:
-            continue
+
+    # Get flags common to all baselines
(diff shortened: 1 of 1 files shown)


hera_cal/lst_stack/calibration.py
```diff
@@ -100,16 +100,26 @@ def _expand_degeneracies_to_ant_gains(
...
+
+    # Map antenna-polarizations to visibility indices
+    antpol_to_idx = {
+        utils.split_pol(pol)[0]: polidx
+        for polidx, pol in enumerate(stack.pols) 
+        if utils.split_pol(pol)[0] == utils.split_pol(pol)[1]
+    }
+    unique_pols = list(antpol_to_idx.keys())
+
+    for polidx, pol in enumerate(unique_pols):
```
jsdillon, line 117: Is there a reason you're not looping over antpol_to_idx.items()

jsdillon (approved):
Looks fine. One possible readability change if you want

hera_cal/lst_stack/calibration.py
tyler-a-cox: No, you're right, that makes more sense. Fixing this now

PR #969, opened 2024-07-30
Fix caching in average_and_inpaint_simultaneously() to not turn an empty dict into a different empty dict 

jsdillon:


tyler-a-cox (approved):
Looks good, thanks @jsdillon!

PR #970, opened 2024-07-31
Fix update to handle pstokes autos

jsdillon:
This fixes a bug introduced in #966 where you can't actually reverse polarizations like `pI`


tyler-a-cox (approved):
This looks fine to me, thanks @jsdillon!

PR #973, opened 2024-08-06
feat: add ability to compute sim inpainting on nonavg data

steven-murray:
Adds ability to compute simultaneous inpainting for non-redundantly averaged data.

### DIFF
--- a/hera_cal/lst_stack/averaging.py
+++ b/hera_cal/lst_stack/averaging.py
@@ -703,8 +703,12 @@ def average_and_inpaint_simultaneously(
     lstavg = reduce_lst_bins(
         stack, get_std=False, get_mad=False, inpainted_mode=False, mean_fill_value=0.0
     )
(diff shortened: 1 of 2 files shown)


hera_cal/lst_stack/averaging.py
```diff
@@ -742,11 +736,15 @@ def average_and_inpaint_simultaneously(
             avg_flgs = lstavg["flags"][iap, :, polidx]
 
             antpol1, antpol2 = utils.split_pol(pol)
+
+            auto_iap = [a for a, b in auto_stack.antpairs].index(antpair[0])
```
jsdillon, line 740: Is this actually used anywhere?

steven-murray: Nope. Removed.

jsdillon (approved):
Approved, assuming tests pass

PR #977, opened 2024-09-26
Add FRF noise covariance calculation library

mwilensky768:
This branch adds some noise propagation functions for the fringe rate filter + coherent averaging steps. 

I've added the basic functions that are demonstrated in the single baseline power spectrum notebook, though they've been modified here to be slightly more general since the notebook versions used notebook-specific info. This PR and [the corresponding one on that repo](https://github.com/HERA-Team/hera_notebook_templates/pull/54) should probably coordinate.

I still need to add unit tests and think I should add a wrapper that goes from ee + nn autos/crosses all the way to a pI time-time cov. per freq so that there is a clearly implemented "pipeline" somewhere, though maybe that should just live in another repo as part of actual pipeline code.


hera_cal/frf.py
```diff
@@ -1998,3 +1999,245 @@ def time_average_argparser():
...
+            Whether to coherently average on the t_avg cadence or not.
+        wgt_tavg_by_nsample (bool):
+            Whether to weight the time average by nsamples. 
+            Otherwise do uniform weighting.
+        nsamples (array):
+            Array proportional to how many nights contributed to each sample.
+        rephase (bool):
+            Whether to rephase before averaging.
+        bl_vec (array):
+            Baseline vector in ENU
```
jsdillon, line 2038: in meters, right?

hera_cal/frf.py
```diff
@@ -1998,3 +1999,245 @@ def time_average_argparser():
...
+
+    Returns:
+        var: array
+            Noise variance (Ntimes, Nfreqs)
+    """
+    var = predict_noise_variance_from_autos(cross_antpairpol, data,
+                                            nsamples=nsamples, auto_ant=auto_ant)
+    var = var[:, freq_slice]
+
+    var_isinf = np.isinf(var)
```
jsdillon, line 2163: you might want to is `~np.isfinite()` unless for some reason you don't want to treat nans as infinite variance (or you're very confident you won't have any nans)

hera_cal/frf.py
```diff
@@ -1998,3 +1999,245 @@ def time_average_argparser():
...
+    Ntimes = cov.shape[1]
+    diags = cov[:, np.arange(Ntimes), np.arange(Ntimes)]
+    corr = cov / np.sqrt(diags[:, None] * diags[:, :, None])
+    
+    return corr
+
+def get_correction_factor_from_cov(cov, tslc=None):
+    """
+    Get a correction factor for PS noise variance prediction in the absence
+    of propagating the noise covariances all the way to delay space.
```
jsdillon, line 2219: Do you want to say a bit more about how this factor is defined?

hera_cal/tests/test_frf.py
```diff
@@ -1066,3 +1067,241 @@ def test_tophat_linear_argparser(self):
...
+
+
+    
+
+
+
+
+    
+
+
```
jsdillon, line 1307: don't think you meant to add all this whitespace

hera_cal/noise.py
```diff
@@ -92,7 +98,11 @@ def predict_noise_variance_from_autos(bl, data, dt=None, df=None, nsamples=None)
         df = np.median(np.ediff1d(data.freqs))
 
     ap1, ap2 = split_pol(bl[2])
-    auto_bl1, auto_bl2 = (bl[0], bl[0], join_pol(ap1, ap1)), (bl[1], bl[1], join_pol(ap2, ap2))
+    if auto_ant is None:
+        auto_bl1, auto_bl2 = (bl[0], bl[0], join_pol(ap1, ap1)), (bl[1], bl[1], join_pol(ap2, ap2))
+    else:
+        auto_bl1, auto_bl2 = (auto_ant, auto_ant, join_pol(ap1, ap1)), (auto_ant, auto_ant, join_pol(ap2, ap2))
+
```
jsdillon, line 105: Can you add a little unit test for this new functionality?

jsdillon (changes requested):
Thanks @mwilensky768. This all looks basically fine, just a couple of minor suggestions

hera_cal/frf.py
```diff
@@ -1998,3 +1999,245 @@ def time_average_argparser():
...
+            (Interleaved) Julian Dates on hd, converted to seconds.
+        filter_cent_use (float):
+            Filter center for main lobe filter, in Hz.
+        filter_half_wid_use (float):
+            Filter half width for main lobe filter, in Hz.
+        freqs (array):
+            Frequencies in the data (in Hz).
+        t_avg (float):
+            Desired coherent averaging length, in seconds.
+        cutoff (float):
```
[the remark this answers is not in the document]
mwilensky768: Done

hera_cal/frf.py
mwilensky768: Yep, added this to the docstring

hera_cal/frf.py
mwilensky768: Thanks, that's a much safer suggestion. I've implemented it.

hera_cal/frf.py
mwilensky768: I've now pointed the reader to the HERA memo and explained the memo in one sentence.

hera_cal/tests/test_frf.py
mwilensky768: Correct -- rampant pressing of the enter key. I've removed it.

hera_cal/noise.py
mwilensky768: This new functionality is checked in `test_prep_var_for_frop` -- do you have a different way you'd like to check it? The only difference in cutting out the `prep_var_for_frop` wrapper is whether nonfinite values get resassigned.

mwilensky768: I know it's sort of funky design to test something in `noise` in `test_frf`

jsdillon: meh, I think that's good enough

jsdillon (approved):
Assuming tests pass... Thanks @mwilensky768 !

PR #978, opened 2024-10-16
Fix pyuvdata deprecation warning related to self.x_orientation vs. self.telescope.x_orientation

jsdillon:
And a few other fixes, plus some formatting issues


tyler-a-cox (approved):
This looks good to me. Thanks @jsdillon!

PR #979, opened 2024-10-21
Add verbose option

mwilensky768:
This is just adding a verbose option to a preparation function for error propagation. Just trying to avoid unnecessarily printing 6 bajillion times.


jsdillon (approved):
Looks good to me!

PR #982, opened 2024-12-24
Rewrite abscal.match_times() to be much faster when dealing with very long lists of modelfiles

jsdillon:
Between lustre's general slowness and the huge number of abscal model files for validation (>8000), it was taking 10s of minutes to find the data's matching model files in LST for abscal. 

This cuts it down to a few seconds by performing a binary search and only actually reading O(logN) files' metadata.


jsdillon:
Errors appear to be related to the pending pyuvsim PR

steven-murray:
I found a small bug in the function and fixed it, but would love to get your take on it, @jsdillon. It was breaking the Validation runs. 

While looking at this, I realized that my function for LST binning does do roughly the same thing, but it is sufficiently different such that it's difficult to merge them (it operates on JDs instead of LSTs and directly uses the fact that the files are UVH5 files).

jsdillon:
@steven-murray if you think this is ready, let's get it merged in

steven-murray:
@jsdillon it would be good to have a test that would have failed without the fix.

steven-murray (approved):

PR #983, opened 2025-01-06
Graduate deinterleave code from notebook

jsdillon:
This code, which separates a datacontainer into multiple alternating streams, currently lives in https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/single_baseline_postprocessing_and_pspec.ipynb 

This adds the function as a class method on DataContainer, so we can remove it from the single baseline power spectrum notebook

### DIFF
--- a/hera_cal/datacontainer.py
+++ b/hera_cal/datacontainer.py
@@ -596,6 +596,41 @@ class DataContainer:
 
         return obj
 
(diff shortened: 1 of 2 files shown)


jsdillon:
Thanks @tyler-a-cox!

PR #984, opened 2025-01-08
Hermitian pinv in simultaneous inpainting

jsdillon:
I've noticed some instances of pinvs not working because an SVD doesn't converge, but adding this option seems to avoid that (not sure why). It's also supposed to be slightly faster!


steven-murray (approved):
Nice! Didn't know about this option, but it makes sense to be faster/more stable.

PR #987, opened 2025-01-29
Add Cholesky decomposition to `nucal._linear_fit`

tyler-a-cox:
Cholesky decomposition with `scipy.linalg.cho_factor` can be up to twice as fast as LU decomposition (`scipy.linalg.lu_factor`). This PR adds solving linear least-squares problems with Cholesky decomposition to `nucal._linear_fit` which should improve the speed of the function when running with large design matrices.

### DIFF
--- a/hera_cal/nucal.py
+++ b/hera_cal/nucal.py
@@ -761,10 +761,11 @@ def _linear_fit(XTX, Xy, solver='lu_solve', alpha=1e-15, cached_input={}):
     # Assert that the method is valid
     assert solver in [
         "lu_solve",
+        "cho_solve",
         "solve",
         "pinv",
         "lstsq",
-    ], "method must be one of {}".format(["lu_solve", "solve", "pinv", "lstsq"])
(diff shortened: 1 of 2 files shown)


hera_cal/nucal.py
```diff
@@ -789,6 +790,19 @@ def _linear_fit(XTX, Xy, solver='lu_solve', alpha=1e-15, cached_input={}):
         # Save info
         cached_output = {'LU': L}
 
+    elif solver == "cho_solve":
+        # Factor XTX using scipy.linalg.cho_factor
+        if "c_and_lower" in cached_input:
+            c_and_lower = cached_input.get('c_and_lower')
```
jsdillon, line 796: Might be good to add a test that involves caching.

jsdillon (changes requested):
Looks good. I'd suggest adding a test that uses the cache (and might as well do it for some of the other solvers while we're at it)

jsdillon (approved):
Looks good to me!

PR #988, opened 2025-02-02
Future pyuvdata compatibility

bhazelton:
Fix error handling to account for slight upcoming changes to error messages when no data matches a selection.

Note that the new changes in pyuvdata make it so that no error is raised if some of the criteria are not met, but we still issue an error if none of the criteria are met. I'm not really sure why you want to allow for that...

I also fixed a test that fails for me locally because it tries to write to a non-existent directory, which I presume you all create. I changed it to use the pytest fixture to write to a temporary directory as I consider that much better behavior for tests.

I'll also note that I had some trouble getting a consistent environment set up to be able to test hera_cal. Apparently the aipy conda package is not compatible with numpy 2.0 (https://github.com/conda-forge/aipy-feedstock/pull/30). If hera_cal really needs to depend on aipy that needs to be fixed. Although I note that at least some of the errors seem to be related to using aipy to interface with miriad files which pyuvdata can do. I **strongly** suggest that at least one CI job uses conda to set up the test environment so you can catch this kind of problem.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -1872,9 +1872,9 @@ def partial_time_io(hd, times=None, time_range=None, lsts=None, lst_range=None, 
                          return_data=False, **kwargs)
         except ValueError as err:
             # check to see if the read failed because of the time range or lst range
-            if 'No elements in time range between ' in str(err) or 'No elements in time_array between ' in str(err):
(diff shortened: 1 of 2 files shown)


jsdillon (approved):
Looks fine to me!

PR #990, opened 2025-02-17
Fix returned flags

steven-murray:
Fixes the flags returned by the per-night in-painter. Before this fix, post-inpaint flags identified by too-large flag gaps in spectral windows were not set on the returned flag array.


jsdillon (approved):
Looks good to me!

PR #991, opened 2025-02-20
fix: avg_flgs needed updating in-place to affect the wrapper function

steven-murray:
Fixes my fix to the `avg_flgs`. They needed to be set in-place because the wrapper function assumes that this is happening.


jsdillon (approved):
Looks good to me!

PR #992, opened 2025-02-21
Add new sparse 2d dpss solver to smooth_cal

tyler-a-cox:
This PR adds the new sparse 2D DPSS solver included in hera_filters to `smooth_cal`. When the time/frequency filtering scales are small, this method is faster and more memory-efficient than `smooth_cal.solve_2D_DPSS`.


jsdillon (approved):
Thanks @tyler-a-cox . good to go when tests pass

PR #993, opened 2025-02-24
Add hidden "waterfall_blacklist" option to enable downweighting of specific freqs/times in smooth_cal

jsdillon:


tyler-a-cox (changes requested):
Thanks, @jsdillon! This mostly looks good to me. I don't love the idea of the hidden parameter, but I think as long as we make an issue for it this is fine with me. I think we should probably either make it a parameter when the object is constructed or add a setter method later down the line if this does become more commonly used.

jsdillon:
You're right @tyler-a-cox. I've added a proper setter method.

tyler-a-cox (approved):
Looks good to me! Thanks @jsdillon

PR #998, opened 2025-04-29
Upgrade pypi publishing

steven-murray:
This is to fix the PyPI publish action, which I think is failing simply because its using an old version of `build` or `twine`. It also adds support for dependabot, which will once a month go through and check our CI yamls to see if there are updates to any of the actions available (which I think was one of the reasons this started failing, because we hadn't updated it in a while). 

When updates are available, a PR will be made. If the PR passes tests and doesn't have a MAJOR version change in it, then it will be automatically merged.

I recommend that we move to using Trusted Publishing on PyPI. I can set that up if you want, @jsdillon, but you'd have to give me manager permissions to the pypi package. Otherwise I'm sure you can set it up.

### DIFF
--- a/.github/workflows/check-build.yaml
+++ b/.github/workflows/check-build.yaml
@@ -1,13 +1,13 @@
-name: Publish Python distributions to PyPI and TestPyPI
+name: Check Build
 
 on:
-  push:
-    tags:
-      - '*'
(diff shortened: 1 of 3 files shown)


jsdillon (approved):
Looks fine to me.

PR #1000, opened 2025-04-30
Update to reflect pyuvdata 3.2 changes to x_orientation handling

jsdillon:
This PR updates to use `.telescope.get_x_orientation_from_feeds()`, rather than `.telescope.x_orientation` which will eventually be deprecated in pyuvdata. The goal is to remove the various deprecation warnings. This cuts down the warnings significantly, though there's still several (not all our fault!) 

When writing data to disk with a given x_orientation, we need to assume that the feed is linearly polarized "x y" (and not circularly polarized "l r"). This should be fine for any HERA application.

This PR had to update some of the fast readers, since the underlying metadata in calfits and uvh5 is changing. Right now, if you use pyuvdata 3.2.0+ and you write a file with .write_uvh5 and then try to read it with HERADataFastReader, it will error (which turns out to be a problem in the new part of the H6C IDR3 pipeline I'm working on). This fixes this and adds new unit tests to cover the problem.

### DIFF
--- a/hera_cal/io.py
+++ b/hera_cal/io.py
@@ -95,7 +95,7 @@ class HERACal(UVCal):
         '''Extract and store useful metadata and array indexing dictionaries.'''
         self.freqs = np.unique(self.freq_array)
         self.times = np.unique(self.time_array)
-        self.pols = [jnum2str(j, x_orientation=self.telescope.x_orientation) for j in self.jones_array]
+        self.pols = [jnum2str(j, x_orientation=self.telescope.get_x_orientation_from_feeds()) for j in self.jones_array]
         self._jnum_indices = {jnum: i for i, jnum in enumerate(self.jones_array)}
         self.ants = [(ant, pol) for ant in self.ant_array for pol in self.pols]
         self._antnum_indices = {ant: i for i, ant in enumerate(self.ant_array)}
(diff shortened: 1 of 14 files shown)


hera_cal/io.py
```diff
@@ -345,8 +345,18 @@ def read_hera_calfits(filenames, ants=None, pols=None,
...
+                if 'XORIENT' in hdr:
+                    # using old standard for x_orientation (prior to pyuvdata 3.2)
+                    info['x_orientation'] = hdr['XORIENT']
+                else:
+                    # using new standard for x_orientation
+                    feed_array = np.array([[item.strip().lower() for item in antdata["POLTYA"]],
+                                           [item.strip().lower() for item in antdata["POLTYB"]]]).T
+                    feed_angle = np.radians([antdata["POLAA"], antdata["POLAB"]]).T
+                    print(feed_array, feed_angle)
+                    print(uvutils.pol.get_x_orientation_from_feeds(feed_array, feed_angle))
```
steven-murray, line 357: Let's remove the print statements

steven-murray (approved):
Thanks @jsdillon! One comment to remove some print statements, otherwise go ahead and merge.

hera_cal/io.py
jsdillon: Good catch, thanks!

jsdillon:
Thanks @steven-murray!

PR #1002, opened 2025-05-15
Generalize get_minimal_slices()

jsdillon:
I've found a (relatively rare) problem inpainting that stems from different bands having different unflagged slices in time (which is now allowed with my Round 3 flagger). 

This code generalizes the get_minimal_slices() function to allow different frequency bands to have different minimal time slices. 

This will break some of my pipeline code, but it feel like it's best to rip the api change band-aid off now and generalize it.


tyler-a-cox (approved):
This looks good to me! Thanks @jsdillon

PR #1003, opened 2025-05-30
Rework how skip_flagged_edges works to allow for smoothing of multiple bands separately

jsdillon:


jsdillon:
As an update, I'm re-running the pipeline on this branch and this seemed to fix the problem more broadly:

image

PR #1005, opened 2025-06-19
Basic implementation of LSTBinConfiguratorSingleBaseline()

jsdillon:
I'm starting to think how to perform LST-binning on single baseline files and the first step is to start this configurator. Right now it's not very fully featured, but I think this is the basics of what I'll need to build a new makeflow type in hera_opm


hera_cal/lst_stack/config.py
```diff
@@ -665,6 +668,84 @@ def _read(cls, group: h5py.Group):
         return cls(**dct)
 
 
+class LSTBinConfiguratorSingleBaseline():
+    '''This is a much simpler, less fully featured version of LSTBinConfigurator,
+    designed for LST-binning single baseline (generally 4-pol) files together
+    across nights. Its main goal is '''
```
steven-murray, line 674: Let's fix up (and finish) the docstring.

hera_cal/lst_stack/config.py
```diff
@@ -665,6 +668,84 @@ def _read(cls, group: h5py.Group):
...
+
+    def __init__(self, datadir: str | Path, nights: list[str], fileglob: str):
+        '''Constuctors object from a datadir, nights, and fileglob.
```
steven-murray, line 677: Constructors --> Construct?

hera_cal/lst_stack/config.py
```diff
@@ -665,6 +668,84 @@ def _read(cls, group: h5py.Group):
...
+            The path to a toml file has makeflow_type = "lstbin_single_baseline".
+            The toml file should have a section called "FILE_CFG" with a sub-section
+            called "datafiles" that contains "datadir", "nights", and "fileglob" keys.
+        '''
+        dct = toml.load(Path(toml_file))
+        if dct['Options']['makeflow_type'] != 'lstbin_single_baseline':
+            raise ValueError(f"Expected 'makeflow_type' to be 'lstbin_single_baseline', "
+                             f"but got {dct['Options']['makeflow_type']}.")
+        datafiles_cfg = dct['FILE_CFG']['datafiles']
+        return cls(datafiles_cfg['datadir'], datafiles_cfg['nights'], datafiles_cfg['fileglob'])
```
steven-murray, line 714: Let's use keywords instead of position args here

hera_cal/lst_stack/config.py
steven-murray: Also, let's document that if we have redundantly averaged baselines, this class requires that the keyed baseline on each night needs to be the same.

steven-murray (commented):
Thanks @jsdillon this looks like a good start -- just a few requests for tightening docs. Is the plan to write out the configuration (i.e. make it easy for a particular notebook to read a list of the files it needs)?

hera_cal/lst_stack/config.py
jsdillon: Oops, my bad. Added those details.

hera_cal/lst_stack/config.py
jsdillon: fixed

hera_cal/lst_stack/config.py
jsdillon: done

jsdillon:
Thanks @steven-murray! I've addressed your comments. 

Right now, I'm not planning to write a new file with the configuration. It's fast enough (since it only looks at filenames, not hdf5 metadata) to just redo the globbing for every baseline. If that turns out to be wrong, I can revisit.

PR #1011, opened 2025-06-27
Graduate distance_to_nearest_nonzero() from notebook to hera_cal

jsdillon:
This function is currently in https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/single_baseline_2D_informed_inpaint.ipynb as `distance_to_nearest_nonzero_vectorized()`. I'd like to use it for my single-baseline LST-binner and re-inpainter that I'm working on. Even if that doesn't pan out, it's good to graduate code from notebooks.

This PR generalizes it to higher dimensions (the distance is always computed over the last dimension and usually all you care about is distance in frequency when the first axis is time). It also improves the documentation and adds unit tests (thanks chat-gpt!)


tyler-a-cox (approved):
Okay, I think this all looks reasonable to me. Thanks @jsdillon

PR #1014, opened 2025-08-12
Add get_FR_buffer_from_spectra()

jsdillon:
This graduates a function in https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/single_baseline_2D_filtered_SNRs.ipynb and in https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/single_baseline_2D_informed_inpaint.ipynb that I'm also planning to put in my single baseline LST-binner and re-inpainter. 

The point of this function is to use the autocorrelation FR spectrum to estimate the additional width of sky-like FRs by adding a beam-based buffer to the range of FRs we'd expect from a source on a given baseline with a uniform beam (just the intrinsically sky-like FRs).


hera_cal/frf.py
```diff
@@ -330,6 +332,56 @@ def sky_frates(uvd, keys=None, frate_standoff=0.0, frate_width_multiplier=1.0,
...
+
+    # take top frequency (widest FR) per band
+    band_top_fr_spectrum = fr_spec_interpolator(freqs[-1])
+    band_top_fr_spectrum /= np.max(band_top_fr_spectrum)
+
+    # fit gaussian to get a decent estimate of the width without being too sensitive to the FT of the limited time range
+    def _gaussian(x, a, sigma):
+        return a * np.exp(-(x**2) / (2 * sigma**2))
+    initial_guess = [1.0, np.std(frates[band_top_fr_spectrum > 1e-3])]
+    popt, _ = optimize.curve_fit(_gaussian, frates, band_top_fr_spectrum, p0=initial_guess)
```
tyler-a-cox, line 379: Should we be giving a warning if `curve_fit` produces a poor fit to the fringe rate profile? I think the fits to the file we've been using in the H6C notebooks should be fine since those profiles should be very Gaussian, but it may be nice to warn users if the fit hasn't converged or the parameter values are unreasonable.

tyler-a-cox (approved):
I think this looks good. I've left a comment on the fitting portion of the code, but am happy with this otherwise!

hera_cal/frf.py
jsdillon: I robustified this calculation so that the starting std is based on at least 3 values. Not sure how much better I can do here..

PR #1015, opened 2025-08-12
Add `SingleBaselineStacker()` class 

jsdillon:
This is a utility class wrapped around `lst_stack.binning.lst_bin_files_for_baselines()` and using `lst_stack.config.LSTBinConfiguratorSingleBaseline()` in order to make it easier to load and store data, eliminating empty times and rolling to the LST branch cut. 

This is part of the single baseline stacking notebook I'm working on and I'm trying to graduate the bits of code that seem useful and self-contained and worthy of unit testing.

### DIFF
--- a/hera_cal/lst_stack/binning.py
+++ b/hera_cal/lst_stack/binning.py
@@ -646,8 +646,8 @@ class SingleBaselineStacker:
 
     @classmethod
     def from_configurator(cls,
-            bl_str: str,
             configurator: LSTBinConfiguratorSingleBaseline,
+            bl_str: str,
             lst_bin_edges: np.ndarray,
             lst_branch_cut: float | None = None,
             where_inpainted_file_rules: list[list[str]] | None = None,
@@ -673,10 +673,10 @@ class SingleBaselineStacker:
 
         Parameters
         ----------
-        bl_str : str
-            The baseline string in the filenames of data to load, e.g., '0_4'.
         configurator : LSTBinConfiguratorSingleBaseline
             The configurator object that contains the mapping from baseline strings to data files.
+        bl_str : str
+            The baseline string in the filenames of data to load, e.g., '0_4'.
         lst_bin_edges : np.ndarray
             Array of LST bin edges (should be one more than the number of bins).
         lst_branch_cut : float | None
--- a/hera_cal/lst_stack/tests/test_binning.py
+++ b/hera_cal/lst_stack/tests/test_binning.py
@@ -456,7 +456,7 @@ def real_files_and_grid():
 def sbs_default(real_files_and_grid):
     p = real_files_and_grid
     return binning.SingleBaselineStacker.from_configurator(
-        p["baseline_string"], p["configurator"], p["lst_bin_edges"]
+        p["configurator"], p["baseline_string"], p["lst_bin_edges"]
     )
 
 
@@ -464,7 +464,7 @@ def sbs_default(real_files_and_grid):
 def sbs_keep_all(real_files_and_grid):
     p = real_files_and_grid
     return binning.SingleBaselineStacker.from_configurator(
-        p["baseline_string"], p["configurator"], p["lst_bin_edges"], to_keep_slice=slice(None)
+        p["configurator"], p["baseline_string"], p["lst_bin_edges"], to_keep_slice=slice(None)
     )
 
 
@@ -472,8 +472,8 @@ def sbs_keep_all(real_files_and_grid):
 def sbs_branchcut(real_files_and_grid):
     p = real_files_and_grid
     return binning.SingleBaselineStacker.from_configurator(
-        p["baseline_string"],
         p["configurator"],
+        p["baseline_string"],
         p["lst_bin_edges"],
         lst_branch_cut=5.4,
         where_inpainted_file_rules=p["where_rules"],


hera_cal/lst_stack/binning.py
```diff
@@ -603,6 +604,109 @@ def lst_bin_files_for_baselines(
...
+                 lst_branch_cut: float | None = None,
+                 where_inpainted_file_rules: list[list[str]] = None,
+                 to_keep_slice: slice | None = None,
+                ):
+        """Creates a SingleBaselineStacker object that loads data for a single baseline, optionally rolls to start after a branch cut,
+        and removes any times at the beginning or end of the data set that have no data.
+
+        Data are stored internally as lists of numpy arrays, just as lst_stack.binning.lst_bin_files_for_baselines() returns.
+
+        self.data: list of length Nlst, each element is (Nnights, Nfreqs, )
```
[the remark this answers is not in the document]
jsdillon: fixed

hera_cal/lst_stack/binning.py
```diff
@@ -603,6 +604,115 @@ def lst_bin_files_for_baselines(
...
+    _list_objects = ('data', 'flags', 'nsamples', 'where_inpainted', 'times_in_bins', 'lsts_in_bins')
+
+    def __init__(self,
+                 bl_str: str,
+                 configurator: LSTBinConfiguratorSingleBaseline,
+                 lst_bin_edges: np.ndarray,
+                 lst_branch_cut: float | None = None,
+                 where_inpainted_file_rules: list[list[str]] = None,
+                 to_keep_slice: slice | None = None,
+                ):
```
steven-murray, line 621: Instead of having the `__init__` take all of these "setup" parameters, instead let the `__init__` just take the final lists that the class needs to hold, e.g.:

```python
def __init__(self, data: list[np.ndarray], flags: list[np.ndarray], ...):
    self.data = data
    self.flags = flags
    ...
```

This makes the class very easy to instantiate, using mocked/simulated data from memory, rather than *having* to create the object from a set of files and a configurator. This is easier to test, and easier to reason about (since the important attributes of the class are displayed up-front, rather than being secretly created within the `__init__` via some other `_load_data` function). 

Even better (in my mind, but I am happy either way) would be to use attrs:

```python
import attrs

@attrs.define(frozen=True)
class SingleBaselineStacker:
    data: list[np.ndarray] = attrs.field()
    flags: list[np.ndarray] = attrs.field()
    # etc...

    @flags.validator
    @nsamples.validator
    @where_inpainted.validator   
    def _list_validator(self, att, val):
        if len(val) != len(self.data):
            raise ValueError(f"{att.name} has a different number of entries than data!")
```

steven-murray: Of course you can do more validation of the entries, this is just an example. This makes the class itself solid -- you know what kind of data struct it represents right up front, and it is self-validating.

hera_cal/lst_stack/binning.py
```diff
@@ -603,6 +604,115 @@ def lst_bin_files_for_baselines(
...
+            start = ts[0].stop if ts[0].start == 0 else 0
+            stop = ts[-1].start if ts[-1].stop == len(self.flags) else len(self.flags)
+            self._to_keep_slice = slice(start, stop)
+        else:
+            self._to_keep_slice = to_keep_slice
+        for list_obj in self._list_objects:
+            setattr(self, list_obj, getattr(self, list_obj)[self._to_keep_slice])
+        self.bin_lst = self.bin_lst[self._to_keep_slice]
+
+    def _load_data(self):
```
steven-murray, line 685: So, to get this logic in, instead of making this an instance method, make it a class method, which takes in the parameters you're currently giving to `__init__`, e.g.:

```python
@classmethod
def from_files_and_configurator(cls, 
    bl_str: str,
    configurator: LSTBinConfiguratorSingleBaseline,
    lst_bin_edges: np.ndarray,
    lst_branch_cut: float | None = None,
    where_inpainted_file_rules: list[list[str]] = None,
    to_keep_slice: slice | None = None,
): -> Self:
    # all your logic, ending with a "return cls(data, flags, ...)"
```

hera_cal/lst_stack/tests/test_binning.py
```diff
@@ -404,29 +407,164 @@ def test_redavg_with_where_inpainted(self, request, tmp_path_factory):
...
+
+# --- fixtures using real test files -------------------------------------------------
+
+@pytest.fixture(scope="module")
+def real_files_and_grid():
+    baseline_string = "0_4"
+    filenames = [
+        "zen.2459861.baseline.0_4.sum.smooth_calibrated.red_avg.inpainted.uvh5",
+        "zen.2459862.baseline.0_4.sum.smooth_calibrated.red_avg.inpainted.uvh5",
+        "zen.2459863.baseline.0_4.sum.smooth_calibrated.red_avg.inpainted.uvh5",
```
steven-murray, line 430: While it's a good idea to test the reading of real (downsampled) files, the problem is that in one year, we've forgotten what's inside these files, so the tests below that use these files for things like testing if the LST wrap worked are harder to know whether they're doing what we expect, because we don't know the properties of the input data. 

Using the framework I laid out above would allow you to also construct instances directly from small test arrays that you can define in each function, making it more obvious what the test is testing for.

PR #1025, opened 2025-10-22
Add phase_flip_time_scale option

jsdillon:
This PR adds an option to set the time scale for smoothing to find phase flips differently from the overall smoothing timescale.


tyler-a-cox (approved):
This looks good to me. Thanks @jsdillon!

PR #1042, opened 2026-03-12
Add new module to handle fitting polarized sources

tyler-a-cox:
## Summary

This PR adds `hera_cal/polfilt.py`, a small module for fitting polarized point sources in per-night data. It's the fitting backend for [`per_night_source_filtering_notebook.ipynb`](https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/per_night_source_filtering_notebook.ipynb) (and its `phase_II/` copy). For each catalog source, the notebook uses it to refine RA, Dec, and rotation measure (RM) before building a Faraday-rotating DPSS model of the source to subtract. `radec_to_lmn` is also used by `single_baseline_pI_FRF_SNR.ipynb` and `single_baseline_scaffolded_and_feathered_inpainter.ipynb`.

Those notebooks on hera_notebook_templates `master` already do `from hera_cal import polfilt`, so they need this merged to run against hera_cal `main`.

## What's in `polfilt.py`

- **`iteratively_fit_polarized_source_params`** is the main entry point. Starting from the catalog RA/Dec/RM, it alternates between two fits:
  - `_fit_polarized_source_position`: phases the visibilities to the current position, removes the Faraday rotation, and solves a weighted linear least-squares problem for small (l, m) offsets.
  - `_fit_rotation_measure`: phases to the new position, averages into one spectrum, grid-searches RM for the most coherent sum, then refines with a bounded `minimize_scalar`.

  It stops when RA and Dec change by less than 1e-4° and RM by less than 1e-3 rad/m², or after `maxiter` iterations. If everything is flagged, it returns the starting values unchanged. The notebook then checks each fit (image SNR and distance from the catalog values) and falls back to the catalog if the fit looks off.
- **`unpack_data_containers`** turns data, flags, and nsamples DataContainers for one polarization into arrays of shape (baselines × times × freqs). Each baseline and its conjugate are included, and it also returns weights (`nsamples × ~flags`), uvw in wavelengths, times, and freqs. The notebook reuses it for its Faraday depth spectra and source images.
- **`radec_to_lmn`** gives the East/North/Up direction cosines of a fixed ICRS position at each time.
- **`estimate_polarized_source_delay`** and **`estimate_freq_from_polarized_source_delay`** give the delay at which a Faraday-rotating source shows up, τ(ν) = 2c²·RM / (πν³), and the inverse. The notebook doesn't use these yet.

## Tests

`hera_cal/tests/test_polfilt.py` tests each function on synthetic data and covers 100% of `polfilt.py` in about 5 s. It checks:
- the coordinate and delay helpers
- unpacking: shapes, slicing, weights, and conjugate baselines
- RM recovery
- exact position recovery at zero offset, and convergence from small offsets
- end-to-end recovery of RA, Dec, and RM by the iterative fitter, starting from offsets in one or all three parameters

The random antenna layout used by the end-to-end tests is seeded, so results are deterministic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hera_cal/polfilt.py
```diff
@@ -0,0 +1,560 @@
...
+        Refined declination in degrees.
+    """
+    # Faraday de-rotation phasor, shape (n_freqs,).
+    lambda_sq = (constants.c.value / freqs) ** 2
+    rm_phasor = np.exp(2j * lambda_sq * rotation_measure)
+
+    # Design matrix M is time-independent: columns are [1, 2πiu, 2πiv].
+    # Shape: (n_bls * n_freqs, 3).
+    u_flat = uvw[:, 0, :].ravel()
+    v_flat = uvw[:, 1, :].ravel()
```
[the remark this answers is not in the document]
tyler-a-cox: Each step is a Gauss–Newton update with an approximate Jacobian, but the data are re-phased every iteration and the true position is an exact fixed point, so the converged fit is unbiased. Tested at Dec −60° to 0° with 10–120 min windows: it converges to within 1e-5° in 1–3 iterations.

hera_cal/polfilt.py
```diff
@@ -0,0 +1,563 @@
...
+from scipy.optimize import minimize_scalar
+
+from hera_cal import datacontainer, utils
+
+# Constants
+SIDEREAL_DAY_SECONDS = 86164.0905  # Sidereal day in seconds
+SPEED_OF_LIGHT = constants.c.value  # m/s
+
+
+def unpack_data_containers(
```
jsdillon: This feels like a function that belongs in utils, rather than here.

Also, when this is loaded in, can we remove it from https://github.com/HERA-Team/snapshot_imager/?

hera_cal/polfilt.py
```diff
@@ -0,0 +1,563 @@
...
+    )
+
+    # Accumulate weighted normal equations over time.
+    XTX = np.zeros((3, 3), dtype=complex)
+    XTy = np.zeros(3, dtype=complex)
+
+    # Direction cosines of the current sky position at all times, shape (3, n_times).
+    lmn0 = radec_to_lmn(ra, dec, times, location)
+
+    for ti in range(vis.shape[1]):
```
jsdillon: Claude suggests that this can be be vectorized instead of looped over:

>The per-integration loop calls radec_to_lmn once per time, and each call is a full astropy AltAz transform: 0.79 s for 200 integrations versus 0.010 s for one vectorized call. Since M is time-independent, the normal equations collapse to XTX = Mᴴ diag(Σₜ w) M and XTy = Mᴴ Σₜ (w · vis_phased); I checked that this matches the loop exactly. With ~10 iterations per source and many sources per night, this is likely the notebook's dominant cost.

hera_cal/polfilt.py
```diff
@@ -0,0 +1,563 @@
...
+    dtest: int = 500,
+) -> float:
+    """
+    Fit the Faraday rotation measure (RM) via a coherent grid search refined
+    with scalar minimisation.
+
+    Phases the visibilities to the supplied sky position, collapses over
+    baselines and times into a Stokes-Q/U spectrum, then evaluates the
+    coherent sum over a grid of trial RM values. The grid maximum is used as
+    the starting point for a bounded scalar minimisation
```
jsdillon: minimisation? is Claude british now? lol

hera_cal/polfilt.py
```diff
@@ -0,0 +1,563 @@
+"""
+Radio astronomy signal processing utilities.
```
jsdillon: A bit general, no?

hera_cal/polfilt.py
```diff
@@ -0,0 +1,563 @@
...
+        as weights.
+    right_ascension : float
+        Initial right ascension in degrees.
+    declination : float
+        Initial declination in degrees.
+    rotation_measure : float
+        Initial rotation measure in rad/m².
+    location : astropy.coordinates.EarthLocation
+        Observatory location used for coordinate transforms.
+    maxiter : int, optional
```
jsdillon: If maxiter is reached, should it error or at least warn?

hera_cal/polfilt.py
```diff
@@ -0,0 +1,563 @@
...
+
+        # Update running estimates for the next iteration.
+        right_ascension = fit_ra
+        declination = fit_dec
+        rotation_measure = fit_rm
+
+    return fit_ra, fit_dec, fit_rm
+
+
+def radec_to_lmn(
```
jsdillon: We might want to move this to utils as well.

hera_cal/tests/test_polfilt.py
```diff
@@ -0,0 +1,317 @@
...
+    lambda_sq = (C / freqs) ** 2
+    spectrum = np.cos(2 * lambda_sq * rm) if real_sky else np.exp(-2j * lambda_sq * rm)
+    return np.exp(2j * np.pi * phase) * spectrum
+
+
+# ---------------------------------------------------------------------------
+# radec_to_lmn
+# ---------------------------------------------------------------------------
+
+class TestRadecToLmn:
```
jsdillon: I'd suggest that we some compare this to pyuvdata's calc_app_coords. Here's Claude's version of that test:

```
def test_matches_pyuvdata_apparent_coords(self):
    """radec_to_lmn agrees with pyuvdata's apparent coords rotated into ENU."""
    from pyuvdata.utils.phasing import calc_app_coords
    from hera_cal import utils
    ra, dec = 45.0, -30.0
    times = np.linspace(2459000.0, 2459000.0 + 4 / 24, 25)
    lat, lon, alt = HERA_LOCATION.lat.deg, HERA_LOCATION.lon.deg, HERA_LOCATION.height.value
    lst = utils.JD2LST(times, latitude=lat, longitude=lon, altitude=alt)
    app_ra, app_dec = calc_app_coords(lon_coord=np.radians(ra), lat_coord=np.radians(dec),
                                      time_array=times, telescope_loc=HERA_LOCATION)
    eq = np.array([np.cos(app_dec), np.zeros_like(app_dec), np.sin(app_dec)])
    expected = np.einsum("tij,jt->it", utils.eq2top_m(app_ra - lst, np.radians(lat)), eq)
    np.testing.assert_allclose(pf.radec_to_lmn(ra, dec, times, HERA_LOCATION), expected, atol=1e-6)
```

hera_cal/polfilt.py
```diff
@@ -0,0 +1,563 @@
...
+import astropy.units as u
+from astropy import constants
+from astropy.time import Time
+from astropy.coordinates import AltAz, EarthLocation, SkyCoord
+from scipy.optimize import minimize_scalar
+
+from hera_cal import datacontainer, utils
+
+# Constants
+SIDEREAL_DAY_SECONDS = 86164.0905  # Sidereal day in seconds
```
jsdillon: Not used

jsdillon (approved):
A few minor suggestions, but otherwise this looks good to go. Thanks @tyler-a-cox !

hera_cal/polfilt.py
tyler-a-cox: Moved unpack_data_containers and radec_to_lmn to utils. polfilt still imports both, so existing notebooks work. Happy to remove them from snapshot_imager once this is merged.

hera_cal/polfilt.py
tyler-a-cox: Moved this as well

hera_cal/polfilt.py
tyler-a-cox: Removed

hera_cal/tests/test_polfilt.py
tyler-a-cox: Added. The agreement is ~1e-12, so I tightened atol to 1e-9.

hera_cal/polfilt.py
tyler-a-cox: Ahhh yes, that's been changed now.

hera_cal/polfilt.py
tyler-a-cox: Haha apparently, Americanized this

PR #1045, opened 2026-04-12
Anderson acceleration in redcal.OmnicalSolver.solve_iteratively

AaronParsons:
I learned recently about Anderson Acceleration (thank you Bahram Khalichi), which uses history to accelerate fixed-point iteration solvers. This is an implementation for our Omnical solver, and it seems to converge with about 2/3 the iterations of the original for negligible extra compute. However, it might change the details of convergence for pathological cases, so it would need to be battle tested before being enabled for HERA. Default (anderson_history=0) is off, so this shouldn't change anything until we want it to.

### DIFF
--- a/hera_cal/redcal.py
+++ b/hera_cal/redcal.py
@@ -812,6 +812,82 @@ def parse_pol_mode(reds):
         return 'unrecognized_pol_mode'
 
 
+class _OmnicalAndersonAccelerator:
+    """Anderson acceleration helper for Omnical's per-pixel fixed-point updates.
+
+    History is stored as concatenated vectors over all solved variables for the
+    currently active pixels. Because Omnical solves each active time/frequency
+    pixel independently, when the active set shrinks we can simply down-select
+    the stored history to the surviving active pixels rather than clearing it.
+    """
+
+    def __init__(self, history=0, start=2, damping=0.5, ridge=1e-8, max_weight=10.0):
+        self.history = int(history)
+        self.start = int(start)
+        self.damping = float(damping)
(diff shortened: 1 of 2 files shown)


hera_cal/tests/test_redcal.py
```diff
@@ -2041,3 +2041,122 @@ def test_redcal_argparser(self):
         assert a.ex_ants == [5, 6]
         assert a.gain == 0.4
         assert a.verbose is True
+
+
+class TestOmnicalAnderson(object):
```
steven-murray, line 2046:
```suggestion
class TestOmnicalAnderson:
```

(`object` no longer necessary in python 3)

hera_cal/redcal.py
```diff
@@ -812,6 +812,82 @@ def parse_pol_mode(reds):
         return 'unrecognized_pol_mode'
 
 
+class _OmnicalAndersonAccelerator:
+    """Anderson acceleration helper for Omnical's per-pixel fixed-point updates.
+
+    History is stored as concatenated vectors over all solved variables for the
```
steven-murray, line 818: Would be great to have a reference to a paper or something on which this is based.

steven-murray (commented):
Thanks @AaronParsons. This looks reasonable to me, and I'm happy if tests pass. I made a couple of small comments. Overall, it would be good to have a little more docstring/comments to make it more clear what's going on (in both tests and code)

jsdillon:
It appears this PR is ever-so-slightly failing its test on python 3.11

PR #1046, opened 2026-04-14
Allow selection of polarization in SingleBaselineStacker

jsdillon:
This would allow us to refactor the LST-stacking notebook to load and analyze one polarization at a time, saving substantial amounts of memory.


tyler-a-cox (approved):
This looks good to me, thanks @jsdillon!

PR #1047, opened 2026-04-21
Improve memory usage inside lst_align from 3x to 2x

jsdillon:
In `lst_align()` there's a TODO note about how we're breifly tripling the data in memory. When we've got 131 nights, this is becoming a real problem.  

With @claude's help, this PR brings it down to doubling by modifying arrays in place that we intend to later delete anyways.

### DIFF
--- a/hera_cal/lst_stack/binning.py
+++ b/hera_cal/lst_stack/binning.py
@@ -70,11 +70,22 @@ def lst_align(
     represent the start, end, or centre of each integration -- either choice will
     incur similar errors.
 
+    .. warning::
+        When ``rephase=True``, the input ``data`` array is modified **in place** to
+        save memory. Rows whose LSTs fall inside the LST range are
+        overwritten with their rephased values; rows outside the LST range are left
+        unchanged (their rephase shift is forced to 0). If the caller still needs
+        the original unrephased data afterward, it must pass in a ``data.copy()``.
+        The per-bin arrays returned in the output list are fancy-index copies of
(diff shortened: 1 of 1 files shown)


steven-murray (approved):

steven-murray:
This is good. I think the in place modification is key here. I'm less convinced that the gymnastics with the del statements is actually doing anything other than obfuscating the code. You still have a reference to the arrays in the calling function, it's just now in the dict. The del inside lst_align therefore can't actually release memory. 

It would be good to test that out and if I'm right I would remove this gymnastics

jsdillon:
I think you're right. I'm doing some profiling, though it's slow because it takes like 20 minutes to load libraries on lustre right now...

jsdillon:
Here's the main branch when I try to load 500 LST bins:

```[mem] after SingleBaselineStacker crosses load: current RSS = 4.83 GB, peak RSS = 20.16 GB, elapsed = 23.00 min```

Here's this PR as of last night doing the same thing:

```[mem] after SingleBaselineStacker crosses load: current RSS = 4.84 GB, peak RSS = 15.86 GB, elapsed = 27.18 min```

Gonna try it now w/o those extra dels as you suggested.

jsdillon:
Yeah, the current version w/o those extra dels works just as well.

```[mem] after SingleBaselineStacker crosses load: current RSS = 4.84 GB, peak RSS = 15.86 GB, elapsed = 4.23 min```

(don't worry about the runtime, that's all due to library loading)

PR #1048, opened 2026-04-22
Implement in-place row permutation and add related tests in lst_stack

jsdillon:
This PR follows up from #1047 and should allow us to reduce the memory footprint from by a factor of 2 (for a total of a factor of 3 improvement including that PR) by shuffling data, flags, nsamples, and where_inpainted arrays around in place so that individual LST bins can be accessed as _views_ not copies via fancy indexing. 

This PR also adds a bunch of tests to make sure that shuffling works properly.

### DIFF
--- a/hera_cal/lst_stack/binning.py
+++ b/hera_cal/lst_stack/binning.py
@@ -44,6 +44,72 @@ def adjust_lst_bin_edges(lst_bin_edges: np.ndarray) -> np.ndarray:
         lst_bin_edges -= 2 * np.pi
 
 
+def _permute_rows_inplace(
+    arrs: Sequence[np.ndarray], perm: np.ndarray
+) -> None:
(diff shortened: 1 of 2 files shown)


jsdillon:
Using the same test case I used over in #1047, here's the before:

```
[mem] after SingleBaselineStacker crosses load: current RSS = 4.84 GB, peak RSS = 15.87 GB, elapsed = 4.44 min
```

and here's the after:

```
[mem] after SingleBaselineStacker crosses load: current RSS = 6.04 GB, peak RSS = 12.79 GB
```

Not sure why total went up a bit... it might have to do with lsts that get loaded but not included with the select. Regardless, I still think this is a win since peak memory is what matters for job allocation.

steven-murray (approved):
Thanks @jsdillon this looks very good to me.

PR #1054, opened 2026-08-03
Add `skycal` module

jsdillon:
This is my first PR in what will likely be a series dealing with the per-SNAP, per-X-engine decoherence, which effectively introduces a time dependent gain on inter-SNAP baselines (so not autos). This needs to be measured, and corrected, which I'll do in a future PR.

The calibration routine developed here is designed to calibrate against a redundantly averaged sky model (in practice, redavg, LST-stacked, and filtered data). It does so in four steps:

1) delay only firstcal
2) rough amplitude calibration using autocorrelations (this has a possible Trx bias)
3) log-linearized solver for amplitudes and phases separately, with a wrap-immunue phase measurement
4) properly linearized, full complex iterator using Gauss-Newton

Working with @claude code, I developed this independently of @tyler-a-cox's implementation of `StEFCal` in #892, which never got merged. This provided a nice test of the code, where I've done steps 1 and 2 to get initial gains on real data, then fed that into either StEFCal or steps 3 and 4 above. The results are below:

image

The good news is that they get the same answer. With ~105 good antennas on this file, the Gauss-Newton approach is ~5x faster. The scaling of StEFCal should be better, so if we got up to 500+ antennas, it might start being advantageous.... not a regime we're going to be in with HERA.

So, I think this implementation can replace that one and we can close #892 , though @tyler-a-cox we should talk that through.

### DIFF
--- a/hera_cal/skycal.py
+++ b/hera_cal/skycal.py
@@ -561,17 +561,27 @@ def _solve_phase_updates(phase_mats, chan_pos, ei, ej, round_wgts, resids,
     return delta_phase
 
 
-def _stationarity_residual(vis_ratio, ratio_wgts, full_gains, ant_i_idx,
-                           ant_j_idx):
-    '''Compute the convergence certificate: the per-channel MAXIMUM
-    fixed-point residual over all solved antennas. At the exact weighted
-    least-squares optimum, every gain equals the weighted projection of the
-    data onto the other antennas' gains,
-    g_i = sum_j(w_ij * z_ij * g_j) / sum_j(w_ij * |g_j|^2) = U / D, so
-    max |U/D - g| / |g| measures how far each cell is from stationarity,
-    independently of the solver's own update sizes. Certifying on maxima
+def _relative_chi2_gradient(vis_ratio, ratio_wgts, full_gains, ant_i_idx,
+                            ant_j_idx):
+    '''Compute the convergence criterion: the per-channel MAXIMUM of the
+    relative gradient of chi^2 over all solved antennas. At the exact
+    weighted least-squares optimum the gradient of chi^2 with respect to
+    every gain vanishes — equivalently, every gain equals the weighted
+    projection of the data onto the other antennas' gains,
+    g_i = sum_j(w_ij * z_ij * g_j) / sum_j(w_ij * |g_j|^2) = U / D — so
+    max |U/D - g| / |g|, the chi^2 gradient normalized by the local
+    curvature D and gain scale, measures how far each cell is from the
+    optimum, independently of the solver's own update sizes. Taking maxima
     (never a median or percentile) is deliberate: a median criterion can
     declare victory while an entire contiguous band remains unconverged.
+
+    NOTE: this deliberately differs from redcal's conv_crit, which is the
+    relative step between iterations. A step-size rule shows that the
+    solver stopped moving (true distance ~ step / (1 - contraction rate)),
+    not that the solution is near the optimum; the chi^2 gradient measures
+    distance from the optimum directly and works for gains from any
+    solver. The 'iter'/'conv_crit' meta keys are shared with redcal, but
+    conv_crit's semantics differ as described here.
 
     Returns: (Nfreqs,) ndarray of the max residual over solved antennas in
         each channel; np.nan for channels with no solved cells.'''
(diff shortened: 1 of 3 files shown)


hera_cal/skycal.py
```diff
@@ -0,0 +1,975 @@
...
+                nactive, nsel, chan_pos, ei, ej, round_wgts)
+            # log-amplitudes: log|resid_ratio| = eta_i + eta_j is exactly
+            # linear in the log-amplitudes — no approximation needed
+            amp_resids = np.log(np.abs(resid_ratio))
+            # phases: wrap-immune eigenvector synchronization (see helper)
+            delta_phase = _phase_sync_init(
+                resid_ratio / np.abs(resid_ratio), round_wgts, chan_pos,
+                ei, ej, nactive, nsel, sync_tol=sync_tol,
+                sync_maxiter=sync_maxiter)
+        else:
```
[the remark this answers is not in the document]
jsdillon: This now properly raises an error.

hera_cal/skycal.py
```diff
@@ -0,0 +1,975 @@
...
+               verbose=verbose)
+    dlys, offsets = model_based_firstcal(data_model_ratio, wgts, freqs,
+                                         verbose=verbose)
+    fc_gains = firstcal_gains(dlys, offsets, freqs)
+
+    utils.echo('Calibrating amplitudes from autocorrelations...',
+               verbose=verbose)
+    abs_amp_gains = calibrate_abs_amp_from_autos(autos, auto_flags=auto_flags)
+    g0 = merge_gains([abs_amp_gains, fc_gains])
+
```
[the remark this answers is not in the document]
jsdillon: if there's a mismatch, we now raise an error

hera_cal/skycal.py
```diff
@@ -0,0 +1,975 @@
...
+    Arguments:
+        data_model_ratio: DataContainer from build_data_model_ratio
+        wgts: DataContainer of weights from build_data_model_ratio
+        g0: dict mapping (ant, antpol) to starting gain waterfalls (e.g. the
+            product of firstcal_gains and calibrate_abs_amp_from_autos gains)
+        ant_to_SNAP_dict: optional dict mapping antenna numbers to SNAP IDs
+            (any hashable). If given, EVERY antenna appearing in the solve must
+            be present (raises ValueError otherwise); if None, all cross
+            baselines participate.
```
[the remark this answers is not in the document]
jsdillon: clarified docstring

hera_cal/skycal.py
```diff
@@ -0,0 +1,1004 @@
...
+            ls_wgts[eqn] = wtot
+        else:
+            ls_data[eqn] = vals[bi]
+            ls_wgts[eqn] = wgts[bi]
+    x = np.full(nants, np.nan)
+    if len(ls_data) > 0:
+        # pinv gives the minimum-norm solution of the (singular) difference
+        # system; the explicit mean subtraction then removes the degeneracy
+        # by fixing the mean to 0
+        sol = linsolve.LinearSolver(ls_data, wgts=ls_wgts).solve(mode='pinv')
```
tyler-a-cox: Should the mode be a parameter here?

hera_cal/skycal.py
```diff
@@ -0,0 +1,1004 @@
...
+
+def _stationarity_residual(vis_ratio, ratio_wgts, full_gains, ant_i_idx,
+                           ant_j_idx):
+    '''Compute the convergence certificate: the per-channel MAXIMUM
+    fixed-point residual over all solved antennas. At the exact weighted
+    least-squares optimum, every gain equals the weighted projection of the
+    data onto the other antennas' gains,
+    g_i = sum_j(w_ij * z_ij * g_j) / sum_j(w_ij * |g_j|^2) = U / D, so
+    max |U/D - g| / |g| measures how far each cell is from stationarity,
+    independently of the solver's own update sizes. Certifying on maxima
```
tyler-a-cox: Is this the same behavior as the convergence criteria for redcal? If not, should make them consistent?

tyler-a-cox (approved):
This looks good thanks @jsdillon! I just left a couple of comments for things to think about, but I think this should be good to go

hera_cal/skycal.py
jsdillon: Done.

hera_cal/skycal.py
jsdillon: No, it's different and arguably better. It's more about demonstrating that the gradient in chi^2 is very small, which we expect to be the case when we're near the optimum, rather than about the step size being small. When dealing with degeneracies, the former is better. I've changed the name of function that computes this quantity to clarify that and added a note to the docstring.

It's possible that we ought to do something similar in redcal, but I'm inclined not to touch it.

PR #1055, opened 2026-08-04
Add fitter for per-SNAP, per-X-engine decoherence

jsdillon:
Long story short:

- I tracked down so-called "fringing RFI" seen in round 5 (see https://reionization.org/manual_uploads/HERA134_H6C_IDR_3_1.pdf) to track real objects in the sky (relatively bright radio galaxies) AND, more importantly, to show up at X-engine boundaries... i.e. every 96 channels. But it doesn't show up at every boundary and it doesn't show up in every file or in every night.
- After doing sky-based calibration against a red_avg, LST-stacked, and filtered set of visibilities from H6C IDR3.1, I found evidence for real discontinuities in the gains which could not survive smooth_cal.
- These discontinuities are due to gains that have their bandpasses multiplied by a set of different flat values for different X-engines.
- I do not see these discontinuities in the autos, despite the much higher SNR.
- Interestingly, the amount of size of the effect is constant over all antennas (both pols) in a given SNAP for a given X-engine and time.
- This leads to the following theory: missed packets from an individual SNAP through the switch to an individual X-engine  lead to stale values in the correlator which do not correlate across SNAPs. Thus, inter-SNAP baselines show decoherence and thus effective discontinuous gain structure which does not show up on intra-SNAP baselines (which includes autos, by definition). In the absolute worst case, these got as high as ~10%, though that's probably rare.
- These discontinuities which were not calibrated (due to smooth_cal), causing sky-like (i.e. fringing) structure to survive the delay filters in Round 5. These then showed up as RFI-like "spikes" because the metrics only looked for the effect incoherently. 

This PR follows from #1054 and adds technology for fitting per-SNAP and per-X-engine decoherence.

Here's perhaps the single cleanest example from a single file on 2459935, the night with the strongest "fringing RFI"

image

And here's an example where we don't find the problem:

image

The basic idea behind the algorithm is to fit each antenna's gain amplitude with smooth DPSS modes while simultaneously fitting a per-SNAP and per-X-engine decoherence. There's some additional fancy machinery @claude helped me develop to try to assess whether the measured decoherence is significant (as compared to the noise covariance of the result), and to smoothly go from rejecting low SNR decoherences to accepting higher SNR ones. But at it's core, that simultaneous fit is the key idea here.

I've also added a new function in `appy_cal` whose job is to take these decoherence results and correct visibility data.

One thing to note: any decoherence that's shared across all X-engines in a band is nearly indistinguishable from smooth bandpass. So we effectively assume that the best X-engine has no decoherence. But if we can correct for the relative decoherence, I think we'll be in a much better position, since that relative decoherence is what introduces spectral structure.

Another thing to note, I believe this effect needs to be measured and corrected on a per-time basis, since we have evidence from a related statistic (looking for differential decoherence in the evens or odds using the diffs) that the correlator can spontaneously recover in the middle of the night:

image


hera_cal/skycal.py
```diff
@@ -1016,3 +1031,662 @@ def sky_calibrate(data, model, autos=None, data_flags=None, model_flags=None,
...
+                resid = _project_out_smooth(
+                    np.column_stack([log_amp, block_design]), wgt,
+                    band_slices, dpss_bases)
+                normal_mat += resid[:, 1:].T @ (wgt[:, None] * resid[:, 1:])
+                rhs += resid[:, 1:].T @ (wgt * resid[:, 0])
+                total_wgts += wgt
+            normal_diag = np.diag(normal_mat).copy()
+            ok_blocks = normal_diag > 0
+            if not ok_blocks.any():
+                continue
```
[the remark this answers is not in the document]
jsdillon: fixed

hera_cal/apply_cal.py
```diff
@@ -107,6 +107,124 @@ def calibrate_redundant_solution(data, data_flags, new_gains, new_flags, all_red
                 data[bl] *= avg_gains**exponent
 
 
+def correct_SNAP_decoherence_in_place(data, decoherence, ant_to_SNAP_dict,
+                                      data_flags=None, nchans_per_block=96):
+    '''Correct visibilities in place for measured per-SNAP, per-X-engine-block
+    signal loss ("decoherence"). The correction is baseline-CLASS dependent,
+    which is why it cannot be folded into per-antenna gains:
+
+        * cross-correlations between antennas on DIFFERENT SNAPs are divided
```
tyler-a-cox, line 116: Are p_i and p_j here the decoherence estimated from SNAPs i and j?

hera_cal/skycal.py
```diff
@@ -1016,3 +1032,692 @@ def sky_calibrate(data, model, autos=None, data_flags=None, model_flags=None,
...
+    (Nfreqs, Nblocks) 0/1 matrix with design[c, b] = 1 where
+    chan_to_block[c] == b.'''
+    chan_to_block = np.arange(nfreqs) // nchans_per_block
+    nblocks = int(chan_to_block[-1]) + 1
+    design = np.zeros((nfreqs, nblocks))
+    design[np.arange(nfreqs), chan_to_block] = 1.0
+    return chan_to_block, design
+
+
+def _dpss_bases(freqs, band_slices, gain_smoothing_scale, eigenval_cutoff,
```
tyler-a-cox, line 1054: Do you expect the band slices to have different sizes? It probably won't matter too much, but you could probably reduce the number of calls to dspec.dpss_operator by assuming the band slices are the same for all of the x-engine bands

hera_cal/skycal.py
```diff
@@ -1016,3 +1032,692 @@ def sky_calibrate(data, model, autos=None, data_flags=None, model_flags=None,
...
+                              eigenval_cutoff=1e-12, detection_sigma=2.0,
+                              full_sigma=3.0, band_split_freq=100e6,
+                              min_block_coverage=0.05,
+                              full_block_coverage=0.9, verbose=False):
+    '''Estimate per-SNAP, per-X-engine-block signal loss ("decoherence")
+    from the spectral structure of the refined gains.
+
+    Model, per (SNAP, integration): the SNAP's antenna-pol log-gain spectra
+    share
+        ln|g_ap(nu)| = smooth_ap(nu) - log_suppression_b + n_ap(nu) + r(nu)
```
tyler-a-cox, line 1250: I think I'm probably just missing something here, but why are you fitting the log of the gain? I'm sure there's a good reason, but it seems like it does make the fitting more complicated. Are linearizing the ratio of two factors?

tyler-a-cox (approved):
Thanks for the updates @jsdillon! I think this is ready to go. I've left some comments because the method was a bit more complex than I had originally thought it might be and had questions as I was reading through. Feel free to update or not if you feel like changes are actually necessary here.

hera_cal/skycal.py
tyler-a-cox: I realized I misunderstood this, band_slices accounts for the high/low band, not the x-engines

hera_cal/apply_cal.py
jsdillon: Correct. Clarified the docstring.

hera_cal/skycal.py
jsdillon: It makes the whole thing a linear system of equations, which then lets me use the non-negative linear least squares fitter to enforce that decoherence is between 0 and 1. I've updated the docstring to clarify this.

hera_cal/skycal.py
jsdillon: that's right. This is the same convention I use in several of the notebooks for filtering and inpainting.

PR #1059, opened 2026-08-19
Allow some fraction of channels to fail calibration convergence temporarily

jsdillon:
Bad antennas can lead to non-convergence. We instead want to temporarily tolerate some level of non-convergence, find those bad antennas, and then re-run the calibration without them.

### DIFF
--- a/hera_cal/skycal.py
+++ b/hera_cal/skycal.py
@@ -540,6 +540,24 @@ def _phase_sync_init(phasors, round_wgts, chan_pos, ei, ej, nchan_active,
     return phases
 
 
+def _solve_or_nan(mats, rhs):
+    '''np.linalg.solve over a stack of systems, with np.nan solutions for any singular
+    members instead of a raise. LAPACK judges near-singular matrices platform-dependently
+    (OpenBLAS raises where Accelerate returns garbage), so a diverging channel's collapsed
+    normal matrix must not take down the whole batch: its np.nan updates propagate into
+    non-finite gains that the refinement loop detects and drops next round.'''
+    try:
+        return np.linalg.solve(mats, rhs)
+    except np.linalg.LinAlgError:
+        solutions = np.full(rhs.shape, np.nan)
+        for k in range(mats.shape[0]):
+            try:
+                solutions[k] = np.linalg.solve(mats[k], rhs[k])
+            except np.linalg.LinAlgError:
+                pass
+        return solutions
+
+
 def _solve_logamp_updates(amp_mats, chan_pos, ei, ej, round_wgts, resids,
                           nchan_active, nsel):
     '''Solve the batched linear weighted least-squares systems for the
@@ -555,7 +573,7 @@ def _solve_logamp_updates(amp_mats, chan_pos, ei, ej, round_wgts, resids,
                + np.bincount(base + ej, weights=wgtd_resids,
                              minlength=nchan_active * nsel)
                ).reshape(nchan_active, nsel)
-    return np.linalg.solve(amp_mats, amp_rhs[:, :, None])[:, :, 0]
+    return _solve_or_nan(amp_mats, amp_rhs[:, :, None])[:, :, 0]
 
 
 def _solve_phase_updates(phase_mats, chan_pos, ei, ej, round_wgts, resids,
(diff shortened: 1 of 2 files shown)


hera_cal/skycal.py
```diff
@@ -632,7 +632,8 @@ def _relative_chi2_gradient(vis_ratio, ratio_wgts, full_gains, ant_i_idx,
 def _refine_gains_single_pol_time(vis_ratio, ratio_wgts, ant_i_idx, ant_j_idx,
                                   nants, refine_tol=1e-8,
                                   refine_maxiter=100, sync_tol=1e-3,
-                                  sync_maxiter=200, verbose=False):
+                                  sync_maxiter=200,
+                                  max_divergent_chan_frac=0.0, verbose=False):
```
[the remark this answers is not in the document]
jsdillon: This function is too new to have any downstream consumers. I think it's fine.

hera_cal/skycal.py
```diff
@@ -780,9 +794,20 @@ def _refine_gains_single_pol_time(vis_ratio, ratio_wgts, ant_i_idx, ant_j_idx,
...
-            # inputs were validated above, so this can only be divergence
-            raise RuntimeError('Per-channel gain refinement diverged: '
-                               f'non-finite gains in round {niter}.')
+            # inputs were validated above, so this can only be divergence.
+            # Drop the offending channels and carry on if allowed; they are
+            # returned as np.nan, so nothing partial escapes.
+            newly_failed = active_idx[np.unique(chan_pos[~np.isfinite(resid_ratio)])]
+            failed_chans[newly_failed] = True
+            active_chans[newly_failed] = False
+            if failed_chans.sum() > max_failed:
```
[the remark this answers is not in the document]
jsdillon: now divergence gets a -1 flag

hera_cal/skycal.py
```diff
@@ -758,6 +770,8 @@ def _refine_gains_single_pol_time(vis_ratio, ratio_wgts, ant_i_idx, ant_j_idx,
 
     gains = np.ones((nsel, nchans), dtype=complex)
     active_chans = np.ones(nchans, dtype=bool)   # channels still iterating
+    failed_chans = np.zeros(nchans, dtype=bool)  # channels that gave up (nan)
+    max_failed = int(np.floor(max_divergent_chan_frac * nchans))
     chan_iters = np.zeros(nfreqs, dtype=int)     # rounds each channel used
```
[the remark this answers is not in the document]
jsdillon: done

hera_cal/skycal.py
```diff
@@ -702,26 +703,38 @@ def _refine_gains_single_pol_time(vis_ratio, ratio_wgts, ant_i_idx, ant_j_idx,
         sync_tol: convergence threshold (radians) for the phase
             synchronization initialization (see _phase_sync_init)
         sync_maxiter: iteration cap for the phase synchronization
+        max_divergent_chan_frac: fraction of channels allowed to fail
+            (diverge or exhaust refine_maxiter) and come back as np.nan
+            instead of raising. Default 0.0 requires all to converge; a
```
tyler-a-cox: Is this suppose to be "raising an error" or something similar?

hera_cal/skycal.py
```diff
@@ -779,14 +794,35 @@ def _refine_gains_single_pol_time(vis_ratio, ratio_wgts, ant_i_idx, ant_j_idx,
...
         with np.errstate(invalid='ignore', divide='ignore'):
             resid_ratio = entry_ratio[in_active] / (gi * np.conj(gj))
-        if not np.isfinite(resid_ratio).all():
-            # inputs were validated above, so this can only be divergence
-            raise RuntimeError('Per-channel gain refinement diverged: '
-                               f'non-finite gains in round {niter}.')
+        # weights for this round: the initialization uses the data weights, while
+        # Gauss-Newton rounds propagate them through the division by the current
+        # gains, which multiplies each weight by (|g_i| |g_j|)^2
+        with np.errstate(over='ignore', invalid='ignore'):
```
tyler-a-cox, line 818: Why the gains squared here? Does this assume you're weighting by the inverse noise variance?

tyler-a-cox (approved):
Okay, this mostly looks good to me! I'm happy with this once the tests pass. Thanks @jsdillon!

hera_cal/skycal.py
jsdillon: yes. I'll fix it.

hera_cal/skycal.py
jsdillon: Yes, the weights are inverse variances, so they carry units of visibility^-2, which means two factors of each gain.

jsdillon: Clarified in the docstring

PR #1060, opened 2026-08-20
Graduate functions used in file_sky_calibration, namely `skycal.expand_sky_gains()` and `apply_cal.calibrate_and_red_avg()`

jsdillon:
This PR graduates three functions from the Phase II `file_sky_calibration` notebook into `hera_cal`:

- **`apply_cal.calibrate_and_red_avg`**: calibrates and redundantly averages visibilities one group at a
  time (never materializing a full-size calibrated copy of the data), with inverse-variance noise weights
  from calibrated autocorrelations, optional per-SNAP/per-X-engine-block decoherence handling consistent
  across baseline classes, and redundant-baseline χ² with exact DoF accounting (including a separate tier
  for excluded antennas). By default the returned nsamples are "effective nsamples," which make the
  standard noise predictor $\sigma^2 = \bar{A}^2/(\Delta t\, \Delta\nu\, n)$ exact for inverse-variance
  averages; the required `reds` (which must include co-polarized auto groups) is passed verbatim to the
  returned `RedDataContainer`s, and averaged autos are always included.
- **`io.SNAPDecoherence.correct_gains`**: removes the fitted per-SNAP suppression staircase from a gain
  dictionary using the stored antenna→SNAP mapping.
- **`skycal.expand_sky_gains`**: closed-form solve for the gains of antennas excluded from sky calibration
  against the frozen solved gains and the sky model, updating gains and per-antenna χ² in place so every
  antenna gets comparable statistics for downstream full-day flagging. Unlike its namesake
  `redcal.expand_omni_gains`, it is deliberately single-pass, so excluded-antenna solutions never
  influence anything else.

**We already have ways of doing redundant averaging, so why not `utils.red_average` or `RedSol.update_vis_from_data`?**
- The decoherence correction only applies to inter-SNAP cross-correlations, so it cannot be folded into
  per-antenna gains; it has to be applied to each group's calibrated data before averaging. The
  redundant-baseline χ² and effective nsamples also require per-baseline calibrated data. Doing all of
  this one group at a time avoids ever holding a full-size calibrated copy of the data in memory.
- `utils.red_average` averages already-calibrated data, so using it would require calibrating (and
  decoherence-correcting) the full data set up front.
- `RedSol.update_vis_from_data` does calibrate group-by-group, but it requires gains packaged as a
  `RedSol` (ours come from sky calibration with an external model), returns no flags or nsamples, and
  falls back to uniform averaging over partially-flagged baselines where a group's weights sum to zero,
  whereas we want those cells zeroed and flagged.

Adds 24 tests; all new code fully covered.

### DIFF
--- a/hera_cal/apply_cal.py
+++ b/hera_cal/apply_cal.py
@@ -11,6 +11,8 @@ import warnings
 from . import io
 from . import utils
 from . import redcal
+from . import noise
+from . import datacontainer
 import pyuvdata.utils as uvutils
 from pyuvdata import UVData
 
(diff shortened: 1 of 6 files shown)


hera_cal/apply_cal.py
```diff
@@ -227,6 +229,285 @@ def correct_SNAP_decoherence_in_place(data, decoherence, ant_to_SNAP_dict,
...
+        if auto_bl in data:
+            with np.errstate(divide='ignore', invalid='ignore'):
+                cal_autos[auto_bl] = np.abs(data[auto_bl]) / np.abs(finite_gains[ant])**2
+    cal_autos = datacontainer.DataContainer(cal_autos)
+
+    def _bl_flags(bl):
+        '''Both antennas' gain_flags, plus unmeasured decoherence blocks for inter-SNAP baselines.'''
+        ant_i, ant_j = utils.split_bl(bl)
+        flags = gain_flags[ant_i] | gain_flags[ant_j]
+        if snap_decoherence is not None and _is_inter_SNAP(bl):
```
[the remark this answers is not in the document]
jsdillon: Adding data_flags is doable. Adding nsamples is probably not appropriate—it makes all the optimal weighting math more complicated. If data is already averaged, then it's already calibrated, and this isn't the right function for further averaging.

hera_cal/apply_cal.py
```diff
@@ -227,6 +229,296 @@ def correct_SNAP_decoherence_in_place(data, decoherence, ant_to_SNAP_dict,
...
+    same way if their groups are listed, though they are never used for noise weights.
+    What is averaged is governed entirely by reds.
+
+    By default the returned nsamples are EFFECTIVE nsamples, defined so that the
+    standard predictor sigma^2 = Abar_i * Abar_j / (dt * df * nsamples), evaluated with
+    the averaged autocorrelations returned here, is exact for every averaged product:
+    Abar_i * Abar_j * sum(w) / (dt * df) for cross groups (above the count, since the
+    average leans on quieter-than-typical pairs) and n^2 * Abar_i * Abar_j / (the sum
+    over antennas of their co-polarized auto products) for the autocorrelations.
+    Effective nsamples is spectrally smooth (auto structure common to all antennas
```
tyler-a-cox: The documentation states that the effective samples are spectrally smooth, but isn't that only the case if the averaged autos (or per-antenna autos) have been inpainted? I may be misinterpreting this statement.

tyler-a-cox (approved):
Looks good to me! Just had one question about the documentation. Thanks @jsdillon!

hera_cal/apply_cal.py
jsdillon: Yes, that's right. I can clarify.

PR #1061, opened 2026-09-01
Correct noise prediction for decoherence

jsdillon:
This PR adjusts the weights in redundant averaging with decoherence to more properly account for that decoherence, which adjusts effective nsamples, which then fixes the apparent x-engine based structure in z-scores. It's a small effect, but also a relatively easy fix.

Old:

image

New:
image


tyler-a-cox (approved):
Thanks @jsdillon! This change looks good to me.

PR #1062, opened 2026-09-07
Add inverse option to SNAPDecoherence.correct_gains()

jsdillon:
Also adds a bit more metadata to freq-averaged chi^2 per baseline.


tyler-a-cox (approved):
This all looks good to me! Thanks @jsdillon

PR #1064, opened 2026-09-22
select times on read when LST-stacking a restricted LST range

tyler-a-cox:
## Summary

This PR makes it cheap to LST-stack a restricted range of LSTs. It's needed by HERA-Team/hera_notebook_templates#72, which adds `LST_MIN`/`LST_MAX` options to `single_baseline_lst_stack_and_reinpaint.ipynb` for exploratory stacks. Without this PR, the notebook still reads every night's file in full, and ranges that wrap through 0h fail.

It makes two changes in `hera_cal/lst_stack/binning.py`:

- **Select times on read.** When only some of a file's integrations fall in the requested LST bins, `_read_one_file` now passes those times to `HERAData.read(times=...)`. uvh5 files are then only partly read, instead of being read in full and cut down afterwards. When every integration is needed, as in the usual stack of all LSTs, the whole file is read as before. The time indices must be strictly increasing (they're cast with `np.asarray`, and anything else raises a `ValueError`), and boolean `time_idx` masks passed to `lst_bin_files_for_baselines()` are converted to indices first.
- **LST ranges that wrap through 2π.** For bin edges that cross 2π (e.g. 23h → 1h), `lst_bin_files_for_baselines()` raised `ValueError: lst_bin_edges must be monotonically increasing`. The cause was `lst_bin_edges %= 2π` being applied in place before the edges were passed to `lst_align()`. `time_idx` now comes from `get_lst_bins()`, the same logic that `lst_align()` and `load_single_baseline_lstcal_gains()` already use, so the data and LSTCal time selections are consistent.

## Performance

Timings are for `_read_one_file` on a synthetic single-baseline file (4000 integrations × 1536 channels × 4 pols, ~180 MB), from a cold page cache on a local SSD:

| integrations needed | `main` | this PR |
|---|---|---|
| all | 1.5 s | 1.6 s |
| 10% | 1.4 s | 0.5 s |

Most of the remaining ~0.5 s per file is fixed metadata overhead: the header is parsed twice, each time with an astropy LST check. `where_inpainted` files are still read in full (~0.3 s here), because `UVFlag` has no select-on-read.

## Tests

- **`test_restricted_lst_range_matches_full_range`:** stacks of a restricted range, one non-wrapping and one wrapping through 2π, give bins identical to the matching bins of a full stack. Both cases fail on `main`.
- **`test_read_one_file_time_subset`:** the right times are passed on read (for arrays and lists), and data, flags, nsamples and where_inpainted all match a full read.
- **`test_read_one_file_unsorted_tind_raises`:** unsorted or repeated indices raise a `ValueError`, including unsigned ones.
- **`test_boolean_time_idx`:** a boolean `time_idx` mask gives the same result as the equivalent indices.
- **Existing tests:** all `hera_cal/lst_stack` tests pass (656 passed, 30 xfailed).
- **End to end:** running the notebook over all LSTs on synthetic data gives bit-identical outputs with this branch and with `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CsJ8HoEMhc2NNU8gVULQqD


hera_cal/lst_stack/binning.py
```diff
@@ -463,17 +463,25 @@ def _read_one_file(
     # Read visibility data
     logger.info(f"Reading {meta_path}")
 
-    # TODO: use Fast readers here instead, and select times directly on read.
+    # If only some of the file's integrations fall within an LST bin, select them
+    # on read, which is true partial I/O for uvh5 files. This requires tind to be
+    # sorted, since data are read in file order.
+    file_times = meta.get_transactional("times")
+    select_times_on_read = (ntimes < len(file_times)) and np.all(np.diff(tind) > 0)
```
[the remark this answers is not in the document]
tyler-a-cox: I guess that's technically correct although I think impossible with the existing code to get that failure mode. Fixed it so it shouldn't be a problem in the future.

hera_cal/lst_stack/binning.py
```diff
@@ -463,17 +463,25 @@ def _read_one_file(
     # Read visibility data
     logger.info(f"Reading {meta_path}")
 
-    # TODO: use Fast readers here instead, and select times directly on read.
+    # If only some of the file's integrations fall within an LST bin, select them
+    # on read, which is true partial I/O for uvh5 files. This requires tind to be
+    # sorted, since data are read in file order.
+    file_times = meta.get_transactional("times")
+    select_times_on_read = (ntimes < len(file_times)) and np.all(tind[1:] > tind[:-1])
```
jsdillon: I feel like this should error if tind is not sorted.

jsdillon: Also, we probably want to explicitly cast `tind = np.asarray(tind)` to make sure this comparison works right

jsdillon (approved):
A couple of small suggestions, but otherwise good to go. Thanks @tyler-a-cox !

